PR #181, opened 2024-11-05
chore: Refactor Pagination in all routes

tipogi:
# Pre-submission Checklist


src/routes/v0/queries.rs
```diff
@@ -1,44 +1,45 @@
...
 
 #[derive(Deserialize)]
 pub struct TagsQuery {
     pub limit_tags: Option<usize>,
     pub limit_taggers: Option<usize>,
 }
 
-#[derive(Deserialize)]
+#[derive(Default, Deserialize, Debug, ToSchema)]
 pub struct PaginationQuery {
```
SHAcollision, line 12: We are using `PaginationQuery` for many more things than the axum handler queries. In fact, `Query` . What about we call this struct simply `Pagination` and move it to `src/types.rs`  or `src/models/types.rs` ? 

Maybe there is other small types around our codebase that could be move into `types.rs` . Enum JsonAction or ScoreAction? I don't know :)

src/routes/v0/stream/queries.rs
```diff
@@ -0,0 +1,195 @@
...
+#[derive(Debug, Serialize, Deserialize, ToSchema, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum PostStreamSorting {
+    Timeline,
+    TotalEngagement,
+}
+
+#[derive(Deserialize, ToSchema, Debug, Clone, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum ViewerStreamSource {
```
SHAcollision, line 16: Do we need both `ViewerStreamSource` and `ViewerStreamSourceQuery` ?

I guess, now that we have broken apart `viewer_id` into the two usages: 1. classical `viewer_id` use for relationships and new 2. `observer_id` for post stream curation, we can rename these structs as `StreamSource` and `StreamSourceQuery`

src/routes/v0/stream/queries.rs
```diff
@@ -0,0 +1,268 @@
...
+        observer_id: String,
+    },
+    // Global stream or author related stream
+    All {
+        author_id: Option<String>,
+    },
+}
+
+impl StreamSource {
+    pub fn has_observer(&self) -> Option<&String> {
```
SHAcollision, line 138: This is a "getter" `get_observer` is the right name. `has_observer` should return a boolean. Same for `has_author` -> `get_author` . 

In any case, it feels that if we need these two functions is because we are not using the `StreamSource` right.

src/models/post/stream.rs
```diff
@@ -57,36 +40,37 @@ impl PostStream {
     }
 
     pub async fn get_posts(
-        post_stream_values: PostStreamValues,
-        post_stream_filters: PostStreamFilters,
+        stream_params: PostStreamQuery,
```
SHAcollision, line 43: There is no need to feed PostStreamQuery to `get_posts`

We are now doing something different in this endpoint versus every other one.

By doing this we: can't have optional `source` and default to `All`. let's just pass down the components: pagination, source, sorting, etc.

src/models/post/stream.rs
```diff
@@ -57,36 +40,37 @@ impl PostStream {
...
     ) -> Result<Option<Self>, Box<dyn Error + Send + Sync>> {
         // Decide whether to use index or fallback to graph query
         let use_index = Self::can_use_index(
-            &post_stream_filters.sorting,
-            &post_stream_values.author_id,
-            &post_stream_filters.source,
-            &post_stream_values.tags,
+            stream_params.sorting.as_ref().unwrap(),
```
SHAcollision, line 47: Clunky. Easier to solve if we pass just the right few components to `get_post()`

PR #215, opened 2024-11-21
feat: add graph queries timeout

tipogi:
# Pre-submission Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 docker/.env
 docker/.database/neo4j/*
 !docker/.database/neo4j/plugins
+!docker/.database/neo4j/conf
 docker/.database/redis
 **/.DS_Store
 /static


tipogi:
There are two ways to apply a timeout to queries in Neo4j:

## Global timeout

To set a global timeout for all transactions, modify the `neo4j.conf `configuration file. In that branch, the file path would be `docker/.database/neo4j/conf/neo4j.conf`:
```conf
db.transaction.timeout=1s
```
#### Verifying the Configuration
To verify if the global timeout configuration was successfully applied (after restarting the container), execute the following Cypher command:
```cypher
CALL dbms.listConfig() YIELD name, value
WHERE name = 'db.transaction.timeout'
RETURN name, value;
```
#### Testing the Global Timeout
Run a query that exceeds the specified timeout, such as:
```cypher
UNWIND range(1, 1000000000) AS i
RETURN i;
```
If the timeout is configured correctly, you should receive an error indicating that the query has timed out

global_timeout

## Per query timeout
For more customized query timeouts, the APOC plugin is required. Follow these steps
#### Configuring the APOC Plugin
1. Download the APOC plugin from the [releases page](https://github.com/neo4j/apoc/releases).
2. Move the downloaded `.jar` file to the plugins directory in our Neo4j setup, `docker/.database/neo4j/plugins`
3. Add Apoc configuration in `docker-compose.yml`
```yml
NEO4JLABS_PLUGINS: '["apoc"]'
NEO4J_dbms_security_procedures_unrestricted: 'apoc.*'
NEO4J_dbms_security_procedures_allowlist: 'apoc.*'
```
#### Restarting the Neo4j Container
1. Remove the data folder from the Neo4j directory (`docker/.database/neo4j/data`) to reset the database.
2. Restart the container to apply the changes

#### Using Per Query Timeout
Once the plugin is installed and the container is restarted, you can apply a timeout to specific queries using `apoc.cypher.runTimeboxed`
```cypher
CALL apoc.cypher.runTimeboxed("MATCH (u1:User)-[f:FOLLOWS]-(u2:User)
 RETURN u1, f, u2",
 {}, 20)
```
If any query exceeding this limit will terminate with an error

apoc_timeout

Issue #169, opened 2024-10-17
Chore: save failed events for debugging purposes


amirRamirfatahi:

SHAcollision:
To further complete this task definition.

We basically just need to catch watcher errors and output to console enough info to understand whether they are expected or not. Make a list out of the unexpected ones and start solving those bugs.

tipogi:
## Common Issues Encountered During Homeserver Synchronization
These are the main problems faced during the initial sync of the homeserver. Each issue is outlined with its symptoms, probable causes, and event logs.

#### A) Unindexed User Profile
The watcher processes a `PUT` event for a post before the user's profile (`profile.json`) has been indexed. This leads to an error when attempting to update the user's post count.
```bash
[2024-12-04T15:42:20Z DEBUG pubky_nexus::events] New event: PUT pubky://e4aakyixz9nqz4omn5z7mfedxsxrsoowyyus3ahebg55q3htg1no/pub/pubky.app/posts/0032FF17768KG
[2024-12-04T15:42:20Z DEBUG pubky_nexus::events::processor] Processing event: Event { uri: "pubky://e4aakyixz9nqz4omn5z7mfedxsxrsoowyyus3ahebg55q3htg1no/pub/pubky.app/posts/0032FF17768KG", event_type: Put, resource_type: Post { author_id: PubkyId("e4aakyixz9nqz4omn5z7mfedxsxrsoowyyus3ahebg55q3htg1no"), post_id: "0032FF17768KG" }, pubky_client: PubkyClient { http: Client { cookie_store: true, accepts: Accepts, proxies: [Proxy(System({}), None)], referer: true, default_headers: {"accept": "*/*", "user-agent": "pubky/0.3.0"} }, pkarr: PkarrClientAsync(PkarrClient { address: Some(0.0.0.0:65437), sender: Sender, cache: InMemoryPkarrCache { inner: Mutex { data: LruCache { len: 10, cap: 1000 }, poisoned: false, .. } }, minimum_ttl: 300, maximum_ttl: 86400 }) } }
[2024-12-04T15:42:20Z DEBUG pubky_nexus::events] Handling PUT event for Post { author_id: PubkyId("e4aakyixz9nqz4omn5z7mfedxsxrsoowyyus3ahebg55q3htg1no"), post_id: "0032FF17768KG" }
[...]

PR #218, opened 2024-11-21
Feat(watcher): add event retry queue

amirRamirfatahi:
bin to handle failures without disrupting the watcher's flow of reading events.
  
# Pre-submission Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,6 +56,10 @@ path = "src/service.rs"
 name = "watcher"
 path = "src/watcher.rs"
 
+[[bin]]
+name = "retry"
+path = "src/retry.rs"
+
 [[bench]]
 name = "user"
 harness = false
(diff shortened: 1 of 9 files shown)


SHAcollision:
Hey @amirRamirfatahi, thanks for putting together this PR!

A couple of different things here:

- Error Curation: It could be useful to start curating a list of errors to identify watcher bugs. The approach in this PR has some potential, though dumping detailed errors straight to the terminal and manually collecting them might be the simplest option for now. Still, feel free to experiment and collect some Errors/bugs that we can act on. 

This relates directly to https://github.com/pubky/pubky-nexus/issues/169#issuecomment-2483902058 and I would say this is pretty high priority. Quoting from the linked comment:

> This is a janitorial task: compile an Issue with unexpected errors, their logs and if possible, how to reproduce them.

- Retry Queue: A retry mechanism could be handy in the future, but it feels premature at the moment. We’d need to carefully define conditions for retries (e.g., "homeserver not found" vs. validation errors). I think we can only be careful enough on defining "what to retry" after we make an exploration of the Errors we get in the watcher. That said, when we scale to multiple homeservers, the retry service concept will definetely be worth revisiting. Low priority at the moment.

I think we can let this PR in paused until the complexity is actually merited, as going in the weeds to review in depth, solve tests issues and merge this code subtracts work-power from other tasks.

PR #221, opened 2024-11-22
feat: graph based profile tags

tipogi:
# Pre-submission Checklist

## PR description
**Added two new query parameters**: 

`viewer_id` and `depth` to the user-related endpoints. The depth parameter specifies the maximum traversal level of FOLLOWING relationships, with a maximum value of 3. This allows traversing up to 3 levels of relationships starting from the viewer_id.

**Retrieving WoT tags in the user profile**:

Use the following endpoints with the new parameters:
- `/v0/user/{user_id}/tags?viewer_id={viewer_id}&depth=2`
- `/v0/user/{user_id}?viewer_id={viewer_id}&depth=2`

These queries will populate the cache with a Time-to-Live (_TTL_) of up to three hours (we could change).

**Precomputing user stream results**:

The user stream endpoint, which also accepts the new parameters, is structured as:
- `/v0/stream/users?user_id={user_id}&source={following|followers|friends}&depth=2&viewer_id={viewer_id}`.

This will write the results to the cache for each _viewer:user_ pair. When visiting the user profile later, the tags will already be precomputed and available in the cache.

**Fetching taggers for a specific WoT tag**:

Use the endpoint: `/v0/user/{user_id}/taggers/{label}?viewer_id={viewer_id}&depth=2`

If the user profile has not been queried previously with the same `viewer_id`, the taggers field will be empty because the necessary data is not cached. To avoid this, the above endpoints should be queried first.

**Handling cache expiration and depth changes**:

If a different depth value is requested within the cache TTL, the results will remain tied to the initial request. New results will only appear once the cache has expired.


src/db/kv/index/sorted_sets.rs
```diff
@@ -56,16 +56,29 @@ pub async fn check_member(
 /// # Errors
 ///
 /// Returns an error if the operation fails.
-pub async fn put(prefix: &str, key: &str, items: &[(f64, &str)]) -> Result<(), DynError> {
```
SHAcollision, line 59: `expiration` is missing from docs

src/db/kv/index/sorted_sets.rs
```diff
@@ -56,16 +56,29 @@ pub async fn check_member(
...
+pub async fn put(
+    prefix: &str,
+    key: &str,
+    items: &[(f64, &str)],
+    expiration: Option<i64>,
```
SHAcollision, line 63: `expiration` is provided in  `minutes` and not documented. However in `index::json` and in `index::sets` we provide expiration in `seconds` . Maybe best to always provide expiration in seconds to avoid confusion.

src/models/tag/traits/collection.rs
```diff
@@ -13,6 +13,11 @@ use crate::{
 
 use crate::models::tag::TagDetails;
 
+const CACHE_SORTED_SET_PREFIX: &str = "Cache:Sorted";
+pub const CACHE_SET_PREFIX: &str = "Cache";
+// Minutes
+const CACHE_TTL: i64 = 180;
```
SHAcollision, line 19: Best if seconds.

We are not creating `CACHE_EXPIRATION` constants in the other `CACHE` implementations ( that is, `recommeneded` user streams). Maybe we should unify patterns: either create constants for all of them or hardcode the value within the function. Creating constants is better.

src/db/kv/index/sets.rs
```diff
@@ -270,14 +270,18 @@ pub async fn put_multiple_sets(
...
     let mut redis_conn = get_redis_conn().await?;
     let mut pipe = redis::pipe();
 
     for (i, key) in index.iter().enumerate() {
         let full_index = format!("{}:{}:{}", &prefix, common_key.join(":"), key);
         if !collections[i].is_empty() {
-            pipe.sadd(full_index, collections[i]);
+            pipe.sadd(&full_index, collections[i]); // Add expiration to the pipeline if specified
+            if let Some(ttl) = expiration {
+                pipe.expire(&full_index, ttl * 60);
```
SHAcollision, line 283: lets always do either seconds (as so far) or minutes. 

Document the new `expiration` param in the docstring

src/db/graph/queries/get.rs
```diff
@@ -236,6 +236,56 @@ pub fn user_tags(user_id: &str) -> neo4rs::Query {
...
+        call_statements.push(format!(
+            "WITH viewer MATCH {} RETURN tagger AS potentialTagger",
+            depth_match
+        ));
+    }
+
+    // Combine the CALL block with UNION
+    let call_clause = format!("CALL {{\n{}\n}}\n", call_statements.join("\nUNION\n"));
+
+    cypher.push_str(&call_clause);
```
SHAcollision, line 268: Looks overly complex. Cypher is always the same regardless of depth level. We can use the `[:FOLLOWS*1..depth]` to match taggers at any depth with simple cypher syntax, right? Could we maybe just replace these 20 lines for this for something like this? Or is the behaviour different?
```rust 
query = format!("(viewer)-[:FOLLOWS*1..{}]->(tagger:User)", depth);
```

SHAcollision (changes requested):
Left a few comments, it's looking good!

PR #222, opened 2024-11-23
Fix tags endpoints with reach

amirRamirfatahi:
# Pre-submission Checklist

### DIFF
--- a/src/db/graph/queries/get.rs
+++ b/src/db/graph/queries/get.rs
@@ -362,28 +362,71 @@ pub fn get_global_hot_tags_taggers(tag_list: &[&str]) -> Query {
     .param("labels", tag_list)
 }
 
-// Analyzes tag usage for a specific list of user IDs. Groups tags by name,
-// showing for each: label, post count and list of user IDs
-// Orders by post_count (descending).
-// Note: Only considers users from the provided users_id list.
-pub fn get_tags_by_user_ids(users_id: &[&str]) -> Query {
+pub fn get_tag_taggers_by_reach(
+    label: &str,
+    user_id: &str,
+    reach_subquery: String,
+    skip: Option<usize>,
+    limit: Option<usize>,
+) -> Query {
+    let skip = skip.unwrap_or(0);
+    let limit: usize = limit.unwrap_or(10);
     query(
-        "
-        UNWIND $ids AS id
-        MATCH (u:User)-[tag:TAGGED]->(p:Post)
-        WHERE u.id = id
-        WITH tag.label AS label, COLLECT(DISTINCT u.id) AS taggers, COUNT(DISTINCT p) AS uniquePosts
-        WITH {
+        format!(
+            "
+            {}
+            MATCH (reach)-[tag:TAGGED]->()
+            WHERE user.id = $user_id AND tag.label = $label
+            WITH reach, MAX(tag.indexed_at) AS tag_time
+            RETURN reach.id AS id
+            ORDER BY tag_time DESC
+            SKIP $skip LIMIT $limit
+            ",
+            reach_subquery
+        )
+        .as_str(),
+    )
+    .param("label", label)
+    .param("user_id", user_id)
+    .param("skip", skip as i64)
+    .param("limit", limit as i64)
+}
+
+pub fn get_hot_tags_by_reach(
+    user_id: &str,
+    reach_subquery: String,
+    skip: usize,
+    limit: usize,
+    max_taggers: usize,
+) -> Query {
+    query(
+        format!(
+            "
+        {}
+        MATCH (reach)-[tag:TAGGED]->(tagged)
+        WHERE user.id = $user_id
+        WITH 
+            tag.label AS label,
+            COLLECT(DISTINCT reach.id)[..{}] AS taggers,
+            COUNT(DISTINCT tagged) AS uniqueTaggedCount,
+            COUNT(DISTINCT reach.id) AS taggers_count
+        WITH {{
             label: label,
             taggers_id: taggers,
-            post_count: uniquePosts,
-            taggers_count: SIZE(taggers)
-        } AS hot_tag
-        ORDER BY hot_tag.post_count DESC
-        RETURN COLLECT(hot_tag) AS hot_tags
+            tagged_count: uniqueTaggedCount,
+            taggers_count: taggers_count
+        }} AS hot_tag
+        ORDER BY hot_tag.tagged_count DESC, hot_tag.label ASC
+        SKIP $skip LIMIT $limit
+        RETURN COLLECT(hot_tag) as hot_tags
     ",
+            reach_subquery, max_taggers
+        )
+        .as_str(),
     )
-    .param("ids", users_id)
+    .param("user_id", user_id)
+    .param("skip", skip as i64)
+    .param("limit", limit as i64)
 }
 
 pub fn get_thread(
--- a/src/models/tag/global.rs
+++ b/src/models/tag/global.rs
@@ -1,21 +1,57 @@
 use super::stream::{TagStreamReach, Taggers};
 use crate::types::DynError;
-use crate::RedisOps;
+use crate::{get_neo4j_graph, queries, RedisOps};
 
 pub struct TagGlobal {}
 
 impl TagGlobal {
     pub async fn get_tag_taggers(
         label: String,
+        user_id: Option<String>,
         reach: Option<TagStreamReach>,
+        skip: Option<usize>,
+        limit: Option<usize>,
     ) -> Result<Option<Vec<String>>, DynError> {
-        match reach {
-            None => read_from_set(&label).await,
-            _ => Ok(None),
+        match user_id {
+            None => read_from_set(&label, skip, limit).await,
+            Some(id) => get_tag_taggers_by_reach(&label, &id, reach.unwrap(), skip, limit).await,
         }
     }
 }
 
-pub async fn read_from_set(label: &str) -> Result<Option<Vec<String>>, DynError> {
-    Taggers::try_from_index_set(&[label], None, None).await
+pub async fn read_from_set(
+    label: &str,
+    skip: Option<usize>,
+    limit: Option<usize>,
+) -> Result<Option<Vec<String>>, DynError> {
+    Taggers::try_from_index_set(&[label], skip, limit).await
+}
+
+pub async fn get_tag_taggers_by_reach(
+    label: &str,
+    user_id: &str,
+    reach: TagStreamReach,
+    skip: Option<usize>,
+    limit: Option<usize>,
+) -> Result<Option<Vec<String>>, DynError> {
+    let graph = get_neo4j_graph()?;
+    let graph = graph.lock().await;
+
+    let query = queries::get::get_tag_taggers_by_reach(
+        label,
+        user_id,
+        reach.to_graph_subquery(),
+        skip,
+        limit,
+    );
+    let mut result = graph.execute(query).await?;
+
+    let mut tagger_ids: Vec<String> = vec![];
+    while let Some(row) = result.next().await? {
+        if let Some(id) = row.get("id")? {
+            tagger_ids.push(id)
+        };
+    }
+
+    Ok(Some(tagger_ids))
 }
--- a/src/models/tag/stream.rs
+++ b/src/models/tag/stream.rs
@@ -7,7 +7,6 @@ use std::ops::Deref;
 use utoipa::ToSchema;
 
 use crate::db::kv::index::sorted_sets::SortOrder;
-use crate::models::follow::{Followers, Following, Friends, UserFollows};
 use crate::{db::connectors::neo4j::get_neo4j_graph, queries};
 use crate::{RedisOps, ScoreAction};
 
@@ -22,6 +21,23 @@ pub enum TagStreamReach {
     Followers,
     Following,
     Friends,
+    MostFollowed,
+    Influencers,
+}
+
+impl TagStreamReach {
+    pub fn to_graph_subquery(&self) -> String {
+        let query = match self {
+            TagStreamReach::Followers => "MATCH (user:User)<-[:FOLLOWS]-(reach:User)",
+            TagStreamReach::Following => "MATCH (user:User)-[:FOLLOWS]->(reach:User)",
+            TagStreamReach::Friends => {
+                "MATCH (user:User)-[:FOLLOWS]->(reach:User), (user)<-[:FOLLOWS]-(reach)"
+            }
+            TagStreamReach::MostFollowed => "",
+            TagStreamReach::Influencers => "",
+        };
+        String::from(query)
+    }
 }
 
 #[async_trait]
@@ -62,13 +78,13 @@ impl Taggers {
 pub struct HotTag {
     label: String,
     taggers_id: Taggers,
-    post_count: u64,
+    tagged_count: u64,
     taggers_count: usize,
 }
 
 // Define a newtype wrapper
 #[derive(Serialize, Deserialize, Debug, ToSchema, Default)]
-pub struct HotTags(Vec<HotTag>);
+pub struct HotTags(pub Vec<HotTag>);
 
 impl RedisOps for HotTags {}
 
@@ -152,7 +168,7 @@ impl HotTags {
                     Some(HotTag {
                         label,
                         taggers_id,
-                        post_count: score as u64,
+                        tagged_count: score as u64,
                         taggers_count,
                     })
                 }
@@ -162,40 +178,22 @@ impl HotTags {
         Ok(Some(hot_tags_stream))
     }
 
-    pub async fn get_stream_tags_by_reach(
+    pub async fn get_hot_tags_by_reach(
         user_id: String,
         reach: TagStreamReach,
+        skip: usize,
+        limit: usize,
+        max_taggers: usize,
     ) -> Result<Option<HotTags>, DynError> {
-        // We cannot use here limit and skip because we want to get all the users reach
-        let users = match reach {
-            TagStreamReach::Followers => {
-                Followers::get_by_id(&user_id, None, Some(isize::MAX as usize))
-                    .await?
-                    .map(|u| u.0)
-            }
-            TagStreamReach::Following => {
-                Following::get_by_id(&user_id, None, Some(isize::MAX as usize))
-                    .await?
-                    .map(|u| u.0)
-            }
-            TagStreamReach::Friends => {
-                Friends::get_by_id(&user_id, None, Some(isize::MAX as usize))
-                    .await?
-                    .map(|u| u.0)
-            }
-        };
-
-        match users {
-            Some(users) => get_users_tags_by_reach(&users).await,
-            None => Ok(None),
-        }
+        let query = queries::get::get_hot_tags_by_reach(
+            &user_id,
+            reach.to_graph_subquery(),
+            skip,
+            limit,
+            max_taggers,
+        );
+        retrieve_from_graph::<HotTags>(query, "hot_tags").await
     }
-}
-
-async fn get_users_tags_by_reach(users: &[String]) -> Result<Option<HotTags>, DynError> {
-    let user_slice = users.iter().map(AsRef::as_ref).collect::<Vec<&str>>();
-    let query = queries::get::get_tags_by_user_ids(user_slice.as_slice());
-    retrieve_from_graph::<HotTags>(query, "hot_tags").await
 }
 
 // Generic function to retrieve data from Neo4J
--- a/src/routes/v0/endpoints.rs
+++ b/src/routes/v0/endpoints.rs
@@ -54,10 +54,10 @@ pub const SEARCH_TAGS_ROUTE: &str = concatcp!(SEARCH_PREFIX, "/tags/{label}");
 
 // TAG endpoints
 // Axum routes
-const TAG_PREFIX: &str = concatcp!(VERSION_ROUTE, "/tag");
-pub const TAG_HOT_ROUTE: &str = concatcp!(TAG_PREFIX, "/hot");
-pub const TAG_REACH_ROUTE: &str = concatcp!(TAG_PREFIX, "/reached/{user_id}/{reach}");
-pub const TAG_TAGGERS_ROUTE: &str = concatcp!(TAG_PREFIX, "/taggers/{label}");
+const TAG_PREFIX: &str = concatcp!(VERSION_ROUTE, "/tags");
+pub const HOT_TAGS_ROUTE: &str = concatcp!(TAG_PREFIX, "/hot");
+pub const HOT_TAGS_BY_REACH_ROUTE: &str = concatcp!(HOT_TAGS_ROUTE, "/{user_id}/{reach}");
+pub const TAG_TAGGERS_ROUTE: &str = concatcp!(TAG_PREFIX, "/tag/{label}/taggers");
 
 // FILE endpoints
 // Axum routes
--- a/src/routes/v0/tag/global.rs
+++ b/src/routes/v0/tag/global.rs
@@ -1,6 +1,6 @@
 use crate::models::tag::global::TagGlobal;
 use crate::models::tag::stream::{HotTag, HotTags, TagStreamReach, Taggers};
-use crate::routes::v0::endpoints::{TAG_HOT_ROUTE, TAG_REACH_ROUTE, TAG_TAGGERS_ROUTE};
+use crate::routes::v0::endpoints::{HOT_TAGS_BY_REACH_ROUTE, HOT_TAGS_ROUTE, TAG_TAGGERS_ROUTE};
 use crate::types::Pagination;
 use crate::{Error, Result};
 use axum::extract::{Path, Query};
@@ -18,7 +18,7 @@ pub struct HotTagsQuery {
 
 #[utoipa::path(
     get,
-    path = TAG_HOT_ROUTE,
+    path = HOT_TAGS_ROUTE,
     params(
         ("max_taggers" = Option<usize>, Query, description = "Retrieve N user_id for each tag"),
         ("skip" = Option<usize>, Query, description = "Skip N tags"),
@@ -34,7 +34,7 @@ pub struct HotTagsQuery {
 )]
 pub async fn hot_tags_handler(Query(query): Query<HotTagsQuery>) -> Result<Json<HotTags>> {
     info!(
-        "GET {TAG_HOT_ROUTE} skip:{:?}, limit:{:?}, max_tagger: {:?}",
+        "GET {HOT_TAGS_ROUTE} skip:{:?}, limit:{:?}, max_tagger: {:?}",
         query.pagination.skip, query.pagination.limit, query.max_taggers
     );
 
@@ -51,18 +51,21 @@ pub async fn hot_tags_handler(Query(query): Query<HotTagsQuery>) -> Result<Json<
     }
 }
 
-#[derive(Deserialize)]
+#[derive(Deserialize, Debug)]
 pub struct TagTaggersQuery {
+    pagination: Pagination,
+    user_id: Option<String>,
     reach: Option<TagStreamReach>,
 }
 
 #[utoipa::path(
     get,
     path = TAG_TAGGERS_ROUTE,
-    tag = "Global tag Taggers",
+    tag = "Tag Taggers",
     params(
         ("label" = String, Path, description = "Tag name"),
-        ("reach" = TagStreamReach, Path, description = "Reach type: Follower | Following | Friends")
+        ("reach" = TagStreamReach, Path, description = "Reach type: Follower | Following | Friends"),
+        ("user_id" = Option<String>, Query, description = "User ID to base reach on"),
     ),
     responses(
         (status = 200, description = "Taggers", body = Vec<String>),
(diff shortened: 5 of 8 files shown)


src/routes/v0/tag/global.rs
```diff
@@ -51,18 +51,21 @@ pub async fn hot_tags_handler(Query(query): Query<HotTagsQuery>) -> Result<Json<
...
 
 #[utoipa::path(
     get,
     path = TAG_TAGGERS_ROUTE,
-    tag = "Global tag Taggers",
+    tag = "Tag Taggers",
     params(
         ("label" = String, Path, description = "Tag name"),
-        ("reach" = TagStreamReach, Path, description = "Reach type: Follower | Following | Friends")
+        ("reach" = TagStreamReach, Path, description = "Reach type: Follower | Following | Friends"),
```
tipogi, line 67: that param should be optional. It is but in swagger does not reflect. Apart of that we should do a `reach` and `user_id` error control before call to the model to fetch the info.

```rust
// XOR operation maybe?
if query.reach.is_some() ^ query.user_id.is_some() {
    // Bad request error
}

File: src/models/tag/global.rs [hunk 2]
```diff
@@ -1,21 +1,57 @@
...
+        limit,
+    );
+    let mut result = graph.execute(query).await?;
+
+    let mut tagger_ids: Vec<String> = vec![];
+    while let Some(row) = result.next().await? {
+        if let Some(id) = row.get("id")? {
+            tagger_ids.push(id)
+        };
+    }
```
tipogi, line 54: Maybe return from the graph query an array and skip the loop?

```rust
if let Some(row) = result.next().await? {
      let users: Vec<String> = row.get("ids").unwrap_or([]);
                ```

File: src/models/tag/global.rs [hunk 3]
```diff
@@ -1,21 +1,57 @@
...
+}
+
+pub async fn get_tag_taggers_by_reach(
+    label: &str,
+    user_id: &str,
+    reach: TagStreamReach,
+    skip: Option<usize>,
+    limit: Option<usize>,
+) -> Result<Option<Vec<String>>, DynError> {
+    let graph = get_neo4j_graph()?;
```
tipogi, line 37: Usually the graph oparations, we wrap inside an scope. 
```
let mut result;
{
   let graph = ...
   result = graph.execute(query).await?;
   ...

File: src/routes/v0/tag/global.rs [hunk 4]
```diff
@@ -75,53 +78,74 @@ pub async fn tag_taggers_handler(
...
-    tag = "Global Tags by reach",
+    path = HOT_TAGS_BY_REACH_ROUTE,
+    tag = "Hot Tags by reach",
     params(
         ("user_id" = String, Path, description = "User Pubky ID"),
-        ("reach" = TagStreamReach, Path, description = "Reach type: Follower | Following | Friends")
+        ("reach" = TagStreamReach, Path, description = "Reach type: Follower | Following | Friends"),
+        ("skip" = Option<usize>, Query, description = "Skip N tags"),
+        ("limit" = Option<usize>, Query, description = "Retrieve N tag"),
+        ("max_taggers" = Option<usize>, Query, description = "Retrieve N user_id for each tag")
```
tipogi, line 122: following the convention, would be `limit_taggers`

src/routes/v0/tag/global.rs
```diff
@@ -75,53 +78,74 @@ pub async fn tag_taggers_handler(
...
+        path.user_id, path.reach, query
     );
 
     let reach = path.reach.unwrap_or(TagStreamReach::Following);
     let user_id = path.user_id;
 
-    match HotTags::get_stream_tags_by_reach(user_id, reach.clone()).await {
+    let skip = query.skip.unwrap_or(0);
+    let limit = query.limit.unwrap_or(20);
+    let max_taggers = query.max_taggers.unwrap_or(20);
```
tipogi, line 144: add default maximun value

src/routes/v0/tag/global.rs
amirRamirfatahi: It shouldn't be optional. These are both in the Path and not the query. so the check is also not needed.

src/routes/v0/tag/global.rs
amirRamirfatahi: it's there?

src/db/graph/queries/get.rs
```diff
@@ -362,28 +362,71 @@ pub fn get_global_hot_tags_taggers(tag_list: &[&str]) -> Query {
...
+        WITH 
+            tag.label AS label,
+            COLLECT(DISTINCT reach.id)[..{}] AS taggers,
+            COUNT(DISTINCT tagged) AS uniqueTaggedCount,
+            COUNT(DISTINCT reach.id) AS taggers_count
+        WITH {{
             label: label,
             taggers_id: taggers,
-            post_count: uniquePosts,
-            taggers_count: SIZE(taggers)
```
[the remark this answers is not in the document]
amirRamirfatahi: fixed

src/routes/v0/tag/global.rs
```diff
@@ -75,53 +78,74 @@ pub async fn tag_taggers_handler(
...
 #[utoipa::path(
     get,
-    path = TAG_REACH_ROUTE,
-    tag = "Global Tags by reach",
+    path = HOT_TAGS_BY_REACH_ROUTE,
+    tag = "Hot Tags by reach",
     params(
         ("user_id" = String, Path, description = "User Pubky ID"),
-        ("reach" = TagStreamReach, Path, description = "Reach type: Follower | Following | Friends")
+        ("reach" = TagStreamReach, Path, description = "Reach type: Follower | Following | Friends"),
```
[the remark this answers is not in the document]
amirRamirfatahi: it's not. it's part of the path. the handler that doesn't have the reach is the global handler, which is a separate handler altogether.

src/models/tag/global.rs
amirRamirfatahi: d74f29c91c83fb0700b0b83b31dd9d6bfbeb76b2

src/models/tag/global.rs
amirRamirfatahi: d74f29c91c83fb0700b0b83b31dd9d6bfbeb76b2

src/routes/v0/tag/global.rs
```diff
@@ -75,53 +78,74 @@ pub async fn tag_taggers_handler(
...
 
-    let reach = query.reach;
-
-    match TagGlobal::get_tag_taggers(label.clone(), reach).await {
+    match TagGlobal::get_tag_taggers(
+        label.clone(),
+        query.user_id,
+        query.reach,
+        query.pagination.skip,
+        query.pagination.limit,
```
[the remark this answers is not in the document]
amirRamirfatahi: 2b334dd0aefdd83880da7fc4654ba7cda1c2889a

src/routes/v0/tag/global.rs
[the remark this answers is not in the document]
tipogi: oh! I see now 👍 What about if we make one endpoint joining the `tags_by_reach_handler` and `hot_tags_handler` similar posts stream. If it has _reach_ param, make query to the graph and if it has not, fetch from cache the global tags. Does it make sense?

src/db/graph/exec.rs
```diff
@@ -25,3 +26,24 @@ pub async fn exec_boolean_row(query: Query) -> Result<bool, DynError> {
     }
     Ok(boolean)
 }
+
+// Generic function to retrieve data from Neo4J
+pub async fn retrieve_from_graph<T>(query: Query, key: &str) -> Result<Option<T>, DynError>
```
tipogi, line 31: better in that file! like this we could use it with others that has the same behavior. What about if we rename to be more precise, for example `retrieve_row_by_key` or up to you if you have better candidate

src/routes/v0/tag/global.rs
```diff
@@ -34,15 +43,27 @@ pub struct HotTagsQuery {
...
 pub async fn hot_tags_handler(Query(query): Query<HotTagsQuery>) -> Result<Json<HotTags>> {
     info!(
-        "GET {TAG_HOT_ROUTE} skip:{:?}, limit:{:?}, max_tagger: {:?}",
-        query.pagination.skip, query.pagination.limit, query.max_taggers
+        "GET {HOT_TAGS_ROUTE} skip:{:?}, limit:{:?}, taggers_limit: {:?}, tagged_type: {:?}, from: {:?}, to: {:?}",
+        query.pagination.skip, query.pagination.limit, query.taggers_limit, query.tagged_type, query.from, query.to
     );
 
     let skip = query.pagination.skip.unwrap_or(0);
     let limit = query.pagination.limit.unwrap_or(40);
```
tipogi, line 51: I think, in all the params that there is an option to fetch unlimited result, we should set a maximun number
```rust
    let limit = query.pagination.limit.unwrap_or(40).min(50);
```

src/routes/v0/tag/global.rs
tipogi: no. It is possible if the user sets `max_taggers` and `limit` params with 100, there is not limit if the user provides the value
```rust
 let limit = query.pagination.limit.unwrap_or(20).min(50);
```

src/routes/v0/tag/global.rs
```diff
@@ -75,53 +99,77 @@ pub async fn tag_taggers_handler(
...
     );
 
     let reach = path.reach.unwrap_or(TagStreamReach::Following);
     let user_id = path.user_id;
 
-    match HotTags::get_stream_tags_by_reach(user_id, reach.clone()).await {
+    let skip = query.pagination.skip.unwrap_or(0);
+    let limit = query.pagination.limit.unwrap_or(40);
+    let taggers_limit = query.taggers_limit.unwrap_or(20);
+    let from = query.from.unwrap_or(0);
```
SHAcollision, line 157: Timeframe pagination with `from` and `to` should not be exposed to the end users. There should be 3 pre-defined options. All, 24h and 1 month.

For reach all -> we precompute the sorted set every 24h. We can do that with a cronjob or by having a TTL of 24 hours and a fallback (on-cache-miss) to graph (re-populates the cache).

src/routes/v0/tag/global.rs
```diff
@@ -75,53 +99,77 @@ pub async fn tag_taggers_handler(
...
+        ("tagged_type" = Option<TaggedType>, Query, description = "Retrieve hot tags by the type of entities tagged with it"),
     ),
     responses(
         (status = 200, description = "Retrieve tags by reach cluster", body = Vec<HotTag>),
         (status = 404, description = "Hot tags not found"),
         (status = 500, description = "Internal server error")
     )
 )]
-pub async fn tags_by_reach_handler(Path(path): Path<TagsByReachQuery>) -> Result<Json<HotTags>> {
+pub async fn tags_by_reach_handler(
```
SHAcollision, line 142: Same as the frontend requested from `/stream` endpoints, we need to merge all of our `hot tags` handlers into one that uses graph or indexes as available for each combination and supports `all|followers|following|friends` and `all|24h|last_month` joint filters.

src/routes/v0/endpoints.rs
```diff
@@ -54,10 +54,10 @@ pub const SEARCH_TAGS_ROUTE: &str = concatcp!(SEARCH_PREFIX, "/tags/{label}");
 
 // TAG endpoints
 // Axum routes
-const TAG_PREFIX: &str = concatcp!(VERSION_ROUTE, "/tag");
-pub const TAG_HOT_ROUTE: &str = concatcp!(TAG_PREFIX, "/hot");
-pub const TAG_REACH_ROUTE: &str = concatcp!(TAG_PREFIX, "/reached/{user_id}/{reach}");
-pub const TAG_TAGGERS_ROUTE: &str = concatcp!(TAG_PREFIX, "/taggers/{label}");
+const TAG_PREFIX: &str = concatcp!(VERSION_ROUTE, "/tags");
+pub const HOT_TAGS_ROUTE: &str = concatcp!(TAG_PREFIX, "/hot");
+pub const HOT_TAGS_BY_REACH_ROUTE: &str = concatcp!(HOT_TAGS_ROUTE, "/{user_id}/{reach}");
```
SHAcollision, line 59: 1. Only `HOT_TAGS_ROUTE` is needed, the reach and timeframe options should all be available in the same endpoint.
2. Therefore `user_id` and `reach` should become a query params.
3. Feels like starting with `HOT` instead of `TAG` breaks our endpoint `const` naming convention.

src/db/graph/queries/get.rs
```diff
@@ -362,28 +363,108 @@ pub fn get_global_hot_tags_taggers(tag_list: &[&str]) -> Query {
...
+    let input_tagged_type = match &tags_query.tagged_type {
+        Some(tagged_type) => tagged_type.to_string(),
+        None => String::from("Post|User"),
+    };
+    query(
+        format!(
+            "
+        {}
+        MATCH (reach)-[tag:TAGGED]->(tagged:{})
+        WHERE user.id = $user_id AND tag.indexed_at >= $from AND tag.indexed_at < $to
```
SHAcollision, line 408: As far as we've learned this is computationally expensive... but, whatever (for now) :sweat_smile:

SHAcollision (changes requested):
Hey @amirRamirfatahi just left a few comments, keep it up ! :rocket:

src/models/tag/stream.rs
```diff
@@ -22,6 +20,47 @@ pub enum TagStreamReach {
...
+        };
+        String::from(query)
+    }
+}
+
+#[derive(Deserialize, Debug, ToSchema, Clone)]
+pub enum TaggedType {
+    Post,
+    User,
+}
```
SHAcollision, line 46: It's not inmediately obvious to me whether we need really need this or whether this is the place for it. Maybe we do, maybe we don't? Certainly on `Post` is needed in current design.

src/models/tag/stream.rs
```diff
@@ -22,6 +20,47 @@ pub enum TagStreamReach {
...
+
+pub struct HotTagsInput {
+    pub from: i64,
+    pub to: i64,
+    pub skip: usize,
+    pub limit: usize,
+    pub taggers_limit: usize,
+    pub tagged_type: Option<TaggedType>,
+}
```
SHAcollision, line 55: A mix of existing types (see `/src/types/pagination`) with custom fields needed for HotTags. If the idea behind this struct is to keep the function number of input params under control, let's aim to reuse the existing Pagination struct.

src/db/graph/queries/get.rs
amirRamirfatahi: if it becomes heavy, we can add special indices in the graph for it.

src/models/tag/stream.rs
amirRamirfatahi: It's providing what the frontend team needs right now, and will also support future use-cases

src/db/graph/queries/get.rs
SHAcollision: Let's add them or add a `//TODO: ` so we are mindful we are doing this here.

src/models/tag/stream.rs
SHAcollision: The idea of `Pagination` is to be universally used across the `service` models and routes all the way down to the RedisOps trait and avoid as much "prop drilling" as possible. I think if we use it as well in routes that only have `skip` and `limit` like `user_following_handler()` for example.

src/models/tag/stream.rs
```diff
@@ -24,6 +23,78 @@ pub enum TagStreamReach {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            TaggedType::Post => write!(f, "Post"),
+            TaggedType::User => write!(f, "User"),
+        }
+    }
+}
+
+#[derive(Deserialize, Debug, ToSchema, Clone)]
+pub enum Timeframe {
```
SHAcollision, line 42:
```suggestion
#[derive(Deserialize, Debug, ToSchema, Clone)]
#[serde(rename_all = "snake_case")]
pub enum Timeframe {
```

src/models/tag/stream.rs
```diff
@@ -24,6 +23,78 @@ pub enum TagStreamReach {
...
+
+impl Display for Timeframe {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Timeframe::Today => write!(f, "Today"),
+            Timeframe::ThisMonth => write!(f, "ThisMonth"),
+            Timeframe::AllTime => write!(f, "AllTime"),
+        }
+    }
+}
```
SHAcollision, line 56: We have never needed these `Display` for any other Enum, let's use `serde`

src/models/tag/stream.rs
```diff
@@ -24,6 +23,78 @@ pub enum TagStreamReach {
...
+            Timeframe::AllTime => 0,
+        };
+        (start, now.timestamp_millis())
+    }
+
+    pub fn to_cache_period(&self) -> i64 {
+        match self {
+            Timeframe::Today => 60 * 60,
+            Timeframe::ThisMonth => 60 * 60 * 24,
+            Timeframe::AllTime => 60 * 60 * 24,
```
SHAcollision, line 85: Looks like we are re-computing and caching `?timeframe=all_time&reach=all` every 24 hours. While these are already always available and always up-to-date directly from Index, is this correct @tipogi ? We might need a bit of triaging `can_use_index()` as in `PostStream` to always retrieve from the most efficient DB.

src/models/tag/stream.rs
```diff
@@ -117,104 +184,66 @@ impl HotTags {
...
-
-        if hot_tags.is_empty() {
-            return Ok(None);
+    fn get_cache_key_parts(tags_query: &HotTagsInput) -> Vec<String> {
+        match &tags_query.tagged_type {
+            Some(tagged) => vec![
+                tags_query.timeframe.to_string(),
+                tagged.to_string(),
+                tags_query.limit.to_string(),
+                tags_query.skip.to_string(),
```
SHAcollision, line 193: It seems unreasonable to have different caches for each `skip` and `limit` . In addition, this will make pagination buggy, as queries and expirations across "pages" might have a computed hot-tags at different times.

src/models/tag/stream.rs
```diff
@@ -90,133 +166,157 @@ impl FromIterator<HotTag> for HotTags {
...
+            key_parts_vector.as_slice(),
+            result
+                .iter()
+                .map(|tag| (tag.tagged_count as f64, tag.label.as_str()))
+                .collect::<Vec<(f64, &str)>>()
+                .as_slice(),
+            Some(GLOBAL_HOT_TAGS_PREFIX),
+            Some(tags_query.timeframe.to_cache_period()),
+        )
+        .await?;
```
tipogi, line 274: Could we do in one loop the creation of the `SORTED_SET` (tag/score) and also, the JSON object (hashMap). Maybe refactor the `HotTags` like that?
```rust
pub struct HotTag {
    taggers_id: Taggers,
    post_count: u64
}
```  
- `label` already is the key of the HashMap, so we do not need that info
- `tagged_count`, we get from the `SORTED_SET` score
- `taggers_count`, we get from the length of the array of `taggers_id`

Once we have the two lists, save in the indexes

tipogi:
- Correct swagger documentation for the endpoint, [here](https://github.com/pubky/pubky-nexus/pull/222/files#diff-3a6a8276dbe26576eec79347daa08c96be304df88891d3da71ed1f2c62936b8dR74-R81): Add `timeframe` query param instead of `from` and `to`. `reach` query param options all in lower case. Not need it `tagged_type` query param now.
- [`HotTagsInput`](https://github.com/pubky/pubky-nexus/pull/222/files#diff-3a6a8276dbe26576eec79347daa08c96be304df88891d3da71ed1f2c62936b8dR105-R111) maybe not need for function signature. What about if we pass one by one
- Add more integration test, with all the query params and some that throws error

src/models/tag/stream.rs
```diff
@@ -1,19 +1,22 @@
...
-use crate::db::kv::index::sorted_sets::SortOrder;
-use crate::models::follow::{Followers, Following, Friends, UserFollows};
-use crate::{db::connectors::neo4j::get_neo4j_graph, queries};
+use crate::queries;
 use crate::{RedisOps, ScoreAction};
 
 pub const TAG_GLOBAL_HOT: [&str; 3] = ["Tags", "Global", "Hot"];
 
-#[derive(Serialize, Deserialize, Debug, ToSchema)]
+const GLOBAL_HOT_TAGS_PREFIX: &str = "Hot_Tags";
```
tipogi, line 17: Lets follow the convention that we do, all the indexes has to be constant and names separated by `:`. An option or you can do something similar:
```rust
// SORTED SET, global. Use TAG_GLOBAL_HOT
// INDEX: Sorted:Tags:Global:Hot
pub const TAG_GLOBAL_HOT: [&str; 3] = ["Tags", "Global", "Hot"];
// These ones, the prefix has to be `CACHE` because they have an expiration time
// SORTED SETs, timeline. 
// INDEX: Cache:Sorted:Tags:Today:Post
pub const TAG_HOT_TODAY: [&str; 3] = ["Tags", "Today", "Post"];
// INDEX: Cache:Sorted:Tags:ThisMonth:Post
pub const TAG_HOT_TODAY: [&str; 3] = ["Tags", "ThisMonth", "Post"];
// Taggers JSON
// INDEX: Cache:Tags:Today:Post:Taggers
pub const TAGGERS_HOT_TODAY: [&str; 3] = ["Tags", "Today", "Post", "Taggers"];
// INDEX: Cache:Tags:ThisMonth:Post:Taggers
pub const TAGGERS_HOT_THIS_MONTH: [&str; 3] = ["Tags", "ThisMonth", "Post", "Taggers"];
```

tipogi:
We are missing `timeframe`-related integration tests. The current mocks are not valid for `Today` and `ThisMonth` scenarios. To address this, we can dynamically create timestamps when importing the mock data, `hot_tags.cypher`. This way, when we run the tests, we ensure that they always pass because the timestamps will consistently fall within the correct temporal context and won't be outdated. Continue working on the [mock](https://github.com/pubky/pubky-nexus/blob/0b189da6d428c41012cae25a46c353b2d655184f/docker/db-graph/mocks/hot_tags.cypher) and add timestamps to the tags associated with posts.

src/routes/v0/tag/global.rs
```diff
@@ -77,54 +50,65 @@ pub async fn tag_taggers_handler(
...
     tag = "Tags",
     params(
-        ("user_id" = String, Path, description = "User Pubky ID"),
-        ("reach" = TagStreamReach, Path, description = "Reach type: follower | following | friends")
+        ("user_id" = Option<String>, Query, description = "User Pubky ID"),
+        ("reach" = Option<TagStreamReach>, Query, description = "Reach type: follower | following | friends | wot"),
+        ("taggers_limit" = Option<usize>, Query, description = "Retrieve N user_id for each tag"),
+        ("skip" = Option<usize>, Query, description = "Skip N tags"),
+        ("limit" = Option<usize>, Query, description = "Retrieve N tag"),
+        ("timeframe" = Option<Timeframe>, Query, description = "Retrieve hot tags for this specific timeframe. Defaults to all_time"),
```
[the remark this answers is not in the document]
amirRamirfatahi: Yeah it's coming in the next PR. I mistakenly added it here but will remove and add it to the influencers one

tipogi:
All the endpoint tests are covered 👍, but the Redis indexes with expiring times are missing the `Cache` prefix, [here](https://github.com/pubky/pubky-nexus/pull/222#discussion_r1868077735) example. Skipping that missing point, it's ready to merge 🚀

src/models/tag/stream.rs
```diff
@@ -90,133 +116,157 @@ impl FromIterator<HotTag> for HotTags {
...
-                    Some(HotTag {
-                        label,
-                        taggers_id,
-                        post_count: score as u64,
-                        taggers_count,
-                    })
-                }
-                None => None,
-            })
+        let mapping = data.unwrap();
```
SHAcollision, line 166: :smiling_face_with_tear: Prompt: _I cry when `unwrap` is used. I don't want to cry. Please fix so I don' cry._

src/models/tag/stream.rs
```diff
@@ -90,133 +116,157 @@ impl FromIterator<HotTag> for HotTags {
...
-        limit: Option<usize>,
-        taggers_limit: Option<usize>,
-    ) -> Result<Option<Self>, DynError> {
-        let hot_tags = Self::try_from_index_sorted_set(
-            &TAG_GLOBAL_HOT,
+    async fn get_from_global_cache(tags_query: &HotTagsInput) -> Result<Option<HotTags>, DynError> {
+        let key_parts = HotTags::get_cache_key_parts(tags_query);
+        let key_parts_vector: Vec<&str> =
+            key_parts.iter().map(|s| s.as_str()).collect::<Vec<&str>>();
+        let data = HotTagsData::try_from_index_json(key_parts_vector.clone().as_slice()).await?;
```
SHAcollision, line 150: But why? We loading the whole json when we had in place everything ALREADY to bring only the wanted values from Redis. Let's stick to the patterns we were using: there was a reason for them.

src/db/graph/queries/get.rs
```diff
@@ -411,28 +414,125 @@ pub fn get_global_hot_tags_taggers(tag_list: &[&str]) -> Query {
...
-        MATCH (u:User)-[tag:TAGGED]->(p:Post)
-        WHERE u.id = id
-        WITH tag.label AS label, COLLECT(DISTINCT u.id) AS taggers, COUNT(DISTINCT p) AS uniquePosts
-        WITH {
+        format!(
+            "
+            {}
+            MATCH (reach)-[tag:TAGGED]->()
+            WHERE user.id = $user_id AND tag.label = $label
+            WITH reach, MAX(tag.indexed_at) AS tag_time
```
SHAcollision, line 441: Please benchmark!
Do we need a new index for `indexed_at` on tag relationships?
```
"CREATE INDEX taggedIndexedAtIndex IF NOT EXISTS FOR ()-[r:TAGGED]-() ON (r.indexed_at)",
```    

Compare performance with and without index. Benchmarks will be useful.

src/models/tag/global.rs
```diff
@@ -1,21 +1,40 @@
...
         reach: Option<TagStreamReach>,
+        skip: usize,
+        limit: usize,
     ) -> Result<Option<Vec<String>>, DynError> {
-        match reach {
-            None => read_from_set(&label).await,
-            _ => Ok(None),
+        match user_id {
+            None => read_from_set(&label, Some(skip), Some(limit)).await,
+            Some(id) => get_tag_taggers_by_reach(&label, &id, reach.unwrap(), skip, limit).await,
```
SHAcollision, line 18: `unwrap()`

src/models/tag/stream.rs
```diff
@@ -90,133 +116,157 @@ impl FromIterator<HotTag> for HotTags {
...
 
-    pub async fn get_stream_tags_by_reach(
+    pub async fn get_hot_tags(
+        user_id: Option<String>,
+        reach: Option<TagStreamReach>,
+        tags_query: &HotTagsInput,
+    ) -> Result<Option<HotTags>, DynError> {
+        match user_id {
+            Some(user_id) => {
+                HotTags::get_hot_tags_by_reach(user_id, reach.unwrap(), tags_query).await
```
SHAcollision, line 235: `unwrap()` all of them must go.

src/models/tag/stream.rs
```diff
@@ -24,6 +26,29 @@ pub enum TagStreamReach {
...
+    }
+}
+
+pub struct HotTagsInput {
+    pub timeframe: Timeframe,
+    pub skip: usize,
+    pub limit: usize,
+    pub taggers_limit: usize,
+    pub tagged_type: Option<TaggedType>,
+}
```
SHAcollision, line 50: I'm not against creating wrappers for inputs. But we need a comprehensive approach. We can't simply just sometimes make them for some models/endpoints. This is not the current pattern we are following, maybe best stick to what we do and them either do or not the same change for every model/endpoint.

amirRamirfatahi: Huge refactors like that are too big to happen fast enough.  
The need for this arose because for the first time a function was going over 7 or something args and we got a lint error because of the number of input args to a function. I think having both approaches at the same time is perfectly fine. It won't hurt readability imo.

src/models/tag/stream.rs
```diff
@@ -58,20 +79,25 @@ impl Taggers {
...
 // Define a newtype wrapper
-#[derive(Serialize, Deserialize, Debug, ToSchema, Default)]
-pub struct HotTags(Vec<HotTag>);
+#[derive(Serialize, Deserialize, Debug, ToSchema, Default, Clone)]
+pub struct HotTags(pub Vec<HotTag>);
 
 impl RedisOps for HotTags {}
 
+#[derive(Serialize, Deserialize, Debug, ToSchema, Default)]
+pub struct HotTagsData(pub HashMap<String, HotTag>);
```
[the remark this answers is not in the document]
amirRamirfatahi: Why put the name of a type inside the name of struct in a statically typed language?  
About the approach, I don't think we need to do everything according to a rule book, when the needs of product are different in various areas.  
Now that we're going to the DB to make the query when it's not cached, why not use the same response for future uses too?

src/models/tag/stream.rs
amirRamirfatahi: https://github.com/pubky/pubky-nexus/pull/252

src/models/tag/global.rs
amirRamirfatahi: https://github.com/pubky/pubky-nexus/pull/252

src/models/tag/stream.rs
SHAcollision: > I think having both approaches at the same time is perfectly fine. It won't hurt readability imo.

Please, it's a matter of mental overhead and entropy, let's KISS.

src/models/tag/stream.rs
```diff
@@ -13,7 +13,7 @@ use crate::{RedisOps, ScoreAction};
 
 pub const TAG_GLOBAL_HOT: [&str; 3] = ["Tags", "Global", "Hot"];
 
-const GLOBAL_HOT_TAGS_PREFIX: &str = "Hot:Tags";
+const GLOBAL_HOT_TAGS_PREFIX: &str = "Cached:Hot:Tags";
```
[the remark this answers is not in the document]
SHAcollision: @amirRamirfatahi please, these still need fixing!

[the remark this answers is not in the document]
tipogi: @amirRamirfatahi @SHAcollision, I am on that in the #292

PR #230, opened 2024-11-26
Feat: eject pubky_app specs

SHAcollision:
For best reusability and specs versioning the Pubky App models have been ejected to its own project:
https://github.com/pubky/pubky-app-specs

Next, we should publish it as crate and import as crate (instead of git).

# Pre-submission Checklist


tipogi (commented):
looks good! nothing to comment 🚀

PR #232, opened 2024-11-27
bug: Update total engagement score just for root posts

tipogi:
# Pre-submission Checklist

### DIFF
--- a/src/events/handlers/post.rs
+++ b/src/events/handlers/post.rs
@@ -90,12 +90,16 @@ pub async fn sync_put(
         PostCounts::update_index_field(parent_post_key_parts, action, JsonAction::Increment(1))
             .await?;
 
-        PostStream::put_score_index_sorted_set(
-            &POST_TOTAL_ENGAGEMENT_KEY_PARTS,
-            parent_post_key_parts,
-            ScoreAction::Increment(1.0),
-        )
-        .await?;
+        // Post replies cannot be included in the total engagement index after they are reposted or receive a reply
+        // Only root posts should be included. Ensure that the parent post is the root post
+        if PostRelationships::is_root(&parent_author_id, &parent_post_id).await? {
+            PostStream::put_score_index_sorted_set(
+                &POST_TOTAL_ENGAGEMENT_KEY_PARTS,
+                parent_post_key_parts,
+                ScoreAction::Increment(1.0),
+            )
+            .await?;
+        }
 
         if action == "replies" {
             // Populate the reply parent keys to after index the reply
--- a/src/models/post/relationships.rs
+++ b/src/models/post/relationships.rs
@@ -45,6 +45,25 @@ impl PostRelationships {
         Ok(None)
     }
 
+    /// Determines whether a post is a root post.
+    /// # Arguments
+    /// - `author_id` - A string slice representing the ID of the author of the post.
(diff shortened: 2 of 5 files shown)


src/models/post/relationships.rs
```diff
@@ -45,6 +45,24 @@ impl PostRelationships {
...
+    /// Determines whether a post is a root post.
+    /// # Arguments
+    /// - `author_id` - A string slice representing the ID of the author of the post.
+    /// - `post_id` - A string slice representing the unique identifier of the post being checked.
+    /// # Returns
+    /// - `Ok(true)` - If the post is a root post (i.e., it has not been replied to or reposted).
+    /// - `Ok(false)` - If the post is not a root post (e.g., it has replies, reposts, or does not exist).
+    /// - `Err(DynError)` - If an error occurs while retrieving or processing the post data.
+    pub async fn is_root(author_id: &str, post_id: &str) -> Result<bool, DynError> {
+        match Self::try_from_index_json(&[author_id, post_id]).await? {
```
SHAcollision, line 57: `Self::get_by_id( )` ?

src/models/post/relationships.rs
```diff
@@ -45,6 +45,24 @@ impl PostRelationships {
...
+            Some(relationship) => {
+                Ok(relationship.replied.is_none() && relationship.reposted.is_none())
```
SHAcollision, line 59: Reposts can be root of a thread and are indexed into `Timeline` and `TotalEngagement` sorted sets.

SHAcollision, line 59:
```suggestion
                Ok(relationship.replied.is_none())
```

src/events/handlers/post.rs
```diff
@@ -373,12 +382,17 @@ pub async fn sync_del(author_id: PubkyId, post_id: String) -> Result<(), DynErro
...
+            // Post replies cannot be included in the total engagement index after the reply is deleted
+            // Only root posts should be included. Ensure that the parent post is the root post
+            if PostRelationships::is_root(&parent_user_id, &parent_post_id).await? {
+                PostStream::put_score_index_sorted_set(
+                    &POST_TOTAL_ENGAGEMENT_KEY_PARTS,
+                    &parent_post_key_parts,
+                    ScoreAction::Decrement(1.0),
+                )
+                .await?;
+            }
```
SHAcollision, line 395: Maybe this is time to do a bit of boy scouting here. We are doing the same thing in two different ways in the same file. We have `is_reply()` function that is not part of the `impl PostRelationship` , instead it is a util hanging out here by itself at the bottom of `handlers/post.rs` , but it is functionally the same (just inverse) to `is_root()`. In specific:
```
    let relationships = PostRelationships::get_by_id(&author_id, &post_id).await?;
    let is_reply = is_reply(&relationships);
```
and
```
PostRelationships::is_root(&parent_user_id, &parent_post_id).await?
```
are just 2 different ways we are now doing the exact same thing.

Ideally we want to keep only one: `is_root()` or `is_reply()`. It should be part of the `PostRelationship` impl and use `self` `is_root(&self)` instead of doing a look up into redis (because sometimes, we will already have the `post_relationship` object and we can avoid the lookup in redis).

src/models/post/relationships.rs
tipogi: yes! that makes more sense, good catch!

PR #234, opened 2024-11-29
Feature/influencers

amirRamirfatahi:
# Pre-submission Checklist

### DIFF
--- a/src/models/user/influencers.rs
+++ b/src/models/user/influencers.rs
@@ -1,0 +1,195 @@
+use crate::db::graph::exec::retrieve_from_graph;
+use crate::db::kv::index::sorted_sets::SortOrder;
+use crate::types::DynError;
+use crate::types::StreamReach;
+use crate::types::Timeframe;
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::ops::Deref;
+use utoipa::ToSchema;
+
+use crate::queries;
+use crate::RedisOps;
+
+const GLOBAL_INFLUENCERS_PREFIX: &str = "Influencers";
+
+#[derive(Deserialize, Serialize, ToSchema, Debug, Clone)]
+pub struct Influencer {
+    id: String,
+    score: f64,
+}
+
+// Define a newtype wrapper
+#[derive(Serialize, Deserialize, Debug, ToSchema, Default, Clone)]
+pub struct Influencers(pub Vec<Influencer>);
+
+impl RedisOps for Influencers {}
+
+#[derive(Serialize, Deserialize, Debug, ToSchema, Default)]
+pub struct InfluencersData(pub HashMap<String, Influencer>);
+
+impl RedisOps for InfluencersData {}
+
+// Create a Influencers instance directly from an iterator of Influencer items
+// Need it in collect()
+impl FromIterator<Influencer> for Influencers {
+    fn from_iter<I: IntoIterator<Item = Influencer>>(iter: I) -> Self {
+        Influencers(iter.into_iter().collect())
+    }
+}
+
+// Implement Deref so Influencers can be used like Vec<String>
+impl Deref for Influencers {
+    type Target = Vec<Influencer>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
+impl Influencers {
+    pub async fn reindex() -> Result<(), DynError> {
+        Influencers::get_global_influencers(0, 100, Timeframe::AllTime).await?;
+        Influencers::get_global_influencers(0, 100, Timeframe::ThisMonth).await?;
+        Ok(())
+    }
+
+    fn get_cache_key_parts(timeframe: Timeframe) -> Vec<String> {
+        vec![timeframe.to_string()]
+    }
+
+    async fn get_from_global_cache(
+        skip: usize,
+        limit: usize,
+        timeframe: Timeframe,
+    ) -> Result<Option<Influencers>, DynError> {
+        let key_parts = Influencers::get_cache_key_parts(timeframe);
+        let key_parts_vector: Vec<&str> =
+            key_parts.iter().map(|s| s.as_str()).collect::<Vec<&str>>();
+        let data =
+            InfluencersData::try_from_index_json(key_parts_vector.clone().as_slice()).await?;
+        let ranking = Influencers::try_from_index_sorted_set(
+            key_parts_vector.as_slice(),
+            None,
+            None,
+            Some(skip),
+            Some(limit),
+            SortOrder::Descending,
+            Some(GLOBAL_INFLUENCERS_PREFIX),
+        )
+        .await?;
+
+        if data.is_none() || ranking.is_none() {
+            return Ok(None);
+        }
+
+        let mapping = data.unwrap();
+
+        // for each value in ranking, look up the value in data
+        let mut influencers = Vec::new();
+        for (id, _) in ranking.unwrap() {
+            if let Some(influencer) = mapping.0.get(&id) {
+                influencers.push(Influencer {
+                    id,
+                    score: influencer.score,
+                });
+            }
+        }
+        Ok(Some(Influencers(influencers)))
+    }
+
+    async fn set_to_global_cache(
+        result: Influencers,
+        timeframe: Timeframe,
+    ) -> Result<(), DynError> {
+        let key_parts = Influencers::get_cache_key_parts(timeframe.clone());
+        let key_parts_vector: Vec<&str> =
+            key_parts.iter().map(|s| s.as_str()).collect::<Vec<&str>>();
+
+        // turn result which is a vector of Influencer, into a mapping from label to Influencer
+        let mapping = result.clone();
+        let influencers_data: HashMap<String, Influencer> = mapping
+            .iter()
+            .map(|item| item.clone())
+            .into_iter()
+            .map(|influencer| (influencer.id.clone(), influencer))
+            .collect();
+
+        // store the data as json in cache
+        InfluencersData::put_index_json(
+            &InfluencersData(influencers_data),
+            key_parts_vector.as_slice(),
+            Some(timeframe.to_cache_period()),
+        )
+        .await?;
+
+        // store the ranking as sorted set in cache
+        Influencers::put_index_sorted_set(
+            key_parts_vector.as_slice(),
+            result
+                .iter()
+                .map(|influencer| (influencer.score as f64, influencer.id.as_str()))
+                .collect::<Vec<(f64, &str)>>()
+                .as_slice(),
+            Some(GLOBAL_INFLUENCERS_PREFIX),
+            Some(timeframe.to_cache_period()),
+        )
+        .await?;
+        Ok(())
+    }
+
+    pub async fn get_influencers(
+        user_id: Option<String>,
+        reach: Option<StreamReach>,
+        skip: usize,
+        limit: usize,
+        timeframe: Timeframe,
+    ) -> Result<Option<Influencers>, DynError> {
+        match user_id {
+            Some(user_id) => {
+                Influencers::get_influencers_by_reach(
+                    user_id,
+                    reach.unwrap(),
+                    skip,
+                    limit,
+                    timeframe,
+                )
+                .await
+            }
+            None => Influencers::get_global_influencers(skip, limit, timeframe).await,
+        }
+    }
+
+    async fn get_influencers_by_reach(
+        user_id: String,
+        reach: StreamReach,
+        skip: usize,
+        limit: usize,
+        timeframe: Timeframe,
+    ) -> Result<Option<Influencers>, DynError> {
+        let query =
+            queries::get::get_influencers_by_reach(user_id.as_str(), reach, skip, limit, timeframe);
+        retrieve_from_graph::<Influencers>(query, "influencers").await
+    }
+
+    async fn get_global_influencers(
+        skip: usize,
+        limit: usize,
+        timeframe: Timeframe,
+    ) -> Result<Option<Influencers>, DynError> {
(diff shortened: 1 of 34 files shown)


src/models/user/influencers.rs
tipogi, line 1: The `Influencers` struct can remain minimal, as its primary purpose is limited to this module. Unnecessary data structures like the `HashMap` can be removed, simplifying the model and reducing overhead.
- The `Influencers` struct is only used within this module, there are no external dependencies on its structure. The stream module (`src/user/stream.rs`), only needs `pubky`s to after create `UserView`s so we can return the vector of tuples (`Vec<(String, u64)>`)
- All required functionality, including filtering and ranking influencers by *timeframe*, can be achieved directly with Redis Sorted Sets.
- The cypher query for influencers, filtered by timeframe, can be optimized to return an array of arrays (e.g., [[pubky, score], [...]]). This eliminates the need for `Influencers` struct conversions.
```rust
let result = retrieve_from_graph::<Vec<(String, u64)>>(query, "influencers").await?;
```
- All Redis indexes that have an expiration time are prefixed with `CACHE` keyword for clarity and consistency throughout the project. This prefix denotes their cache nature and helps distinguish them from other keys in Redis

src/routes/v0/stream/users.rs
```diff
@@ -31,7 +34,10 @@ pub struct UserStreamQuery {
         ("skip" = Option<usize>, Query, description = "Skip N followers"),
         ("limit" = Option<usize>, Query, description = "Retrieve N followers"),
         ("source" = Option<UserStreamSource>, Query, description = "Source of users for the stream."),
-        ("depth" = Option<usize>, Query, description = "User trusted network depth, user following users distance. Numbers bigger than 4, will be ignored")
+        ("source_reach" = Option<StreamReach>, Query, description = "The target reach of the source. Supported in 'influencers' source."),
```
tipogi, line 37: To continue with the naming conventions, it should be `reach`

src/routes/v0/stream/users.rs
```diff
@@ -92,7 +99,10 @@ pub async fn stream_users_handler(
         Some(skip),
         Some(limit),
         source.clone(),
+        query.source_reach,
         query.depth,
+        Some(timeframe),
+        query.preview,
     )
     .await
```
tipogi, line 107: If it does not have any influencers in the selected `timeframe`, it throws a _500_ error which it should be _404_ type

tipogi (commented):
For future reference, it would be helpful to avoid merging one PR into another, as it can make things a bit confusing to review

src/routes/v0/stream/users.rs
amirRamirfatahi: this `reach` is related to the `source` so it's source's reach, hence the naming.

src/routes/v0/stream/users.rs
amirRamirfatahi: It's using the same function we already have? I don't think so as this function is returning a None if the inner user id list is  empty. Am I missing something?

amirRamirfatahi: ok fixed it.

src/models/user/influencers.rs
```diff
@@ -0,0 +1,170 @@
...
+use crate::types::Timeframe;
+use chrono::Utc;
+use serde::{Deserialize, Serialize};
+use std::ops::Deref;
+use utoipa::ToSchema;
+
+use crate::queries;
+use crate::RedisOps;
+
+const GLOBAL_INFLUENCERS_PREFIX: &str = "Cached:Influencers";
```
tipogi, line 14: Wrong cache prefix, it has to be `Cache`

src/models/user/influencers.rs
```diff
@@ -0,0 +1,170 @@
...
+
+// Implement Deref so Influencers can be used like Vec<String>
+impl Deref for Influencers {
+    type Target = Vec<Influencer>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
+
```
tipogi, line 44: The use of _Influencer_ struct feels unnecessary, as a tuple `(score, pubky)` seems to capture the same logic effectively. A vector of such tuples, `Vec<(score, pubky)>`, _Influencers_ struct now, would suffice, reducing complexity without losing clarity

src/routes/v0/stream/users.rs
tipogi: The `reach` parameter name is concise, aligns with existing conventions, and avoids unnecessary redundancy. The swagger description already clarifies its relationship to the source. Another point, `source_reach` cannot be combined with the other `UserStreamSource` so better candidate, if you want to add a prefix, would be `influencer_reach` but I do not see neither...

src/models/user/stream.rs
```diff
@@ -237,17 +257,21 @@ impl UserStream {
...
+                preview.unwrap_or(false),
             )
             .await?
-            .map(|set| set.into_iter().map(|(user_id, _score)| user_id).collect()),
+            .map(|result| {
+                result
+                    .iter()
+                    .map(|influencer| influencer.id.clone())
+                    .collect()
+            }),
```
tipogi, line 274: I think the previous implementation was correct. We do not need to `clone()` the user pubky plus do not the create new structs

src/routes/v0/stream/users.rs
```diff
@@ -82,18 +89,37 @@ pub async fn stream_users_handler(
...
+        user_id: query.user_id.clone(),
+        viewer_id: query.viewer_id,
+        skip: Some(skip),
+        limit: Some(limit),
+        source: source.clone(),
+        source_reach: query.source_reach,
+        depth: query.depth,
+        timeframe: Some(timeframe),
+        preview: query.preview,
+    })
```
tipogi, line 122: This function exceeds the recommended maximum of 7 parameters, which may cause a `clippy` warning

src/models/user/influencers.rs
amirRamirfatahi: I wouldn't call it unnecessary.  
Having a struct for it means in the future we can easily add more properties to it if needed, without having to refactor a lot of code.  
This isn't hurting readability, performance or consistency, so I don't think it hurts to keep it here.  
What do you think?

src/models/user/influencers.rs
```diff
@@ -0,0 +1,170 @@
...
+            let skip = Utc::now().timestamp_subsec_micros() % 98;
+            (skip as usize, 3)
+        } else {
+            (skip, limit)
+        };
+        match user_id {
+            Some(user_id) => {
+                Influencers::get_influencers_by_reach(
+                    user_id,
+                    reach.unwrap(),
```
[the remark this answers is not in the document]
amirRamirfatahi: reach couldn't be none because of previous checks.  
Nonetheless, you are correct. will fix it.

src/routes/v0/stream/users.rs
amirRamirfatahi: That's why it's a struct and the function is just getting one argument.

src/models/user/stream.rs
amirRamirfatahi: https://github.com/pubky/pubky-nexus/pull/234#discussion_r1884585292

src/routes/v0/stream/users.rs
tipogi: ups! you are right, I miss that struct haahh

src/models/user/influencers.rs
tipogi: You know what I think from the beginning 😁. That data about users will always go to get `UserViews` so now and in the future, I do not think that we will need that struct. @SHAcollision any opinion about that?

src/models/user/stream.rs
tipogi: yes, that depends in the `Influencers` struct 👍

src/models/user/stream.rs
```diff
@@ -1,47 +1,63 @@
...
+use crate::types::{DynError, StreamReach, Timeframe};
 use crate::{db::kv::index::sorted_sets::SortOrder, RedisOps};
 use crate::{get_neo4j_graph, queries};
 use serde::{Deserialize, Serialize};
 use tokio::task::spawn;
 use utoipa::ToSchema;
 
 pub const USER_MOSTFOLLOWED_KEY_PARTS: [&str; 2] = ["Users", "MostFollowed"];
-pub const USER_PIONEERS_KEY_PARTS: [&str; 2] = ["Users", "Pioneers"];
+pub const USER_INFLUENCERS_KEY_PARTS: [&str; 2] = ["Users", "Influencers"];
```
SHAcollision, line 11: This is a breaking change of Index schemas and we do not have a "migration". What is the plant? Is this rename needed just yet? If we update or staging sever the feature will stop working.

amirRamirfatahi: this was fixed like this because the graph is still too small for a reindex to be costly

src/types/mod.rs
```diff
@@ -19,3 +19,12 @@ pub enum StreamSorting {
     Timeline,
     TotalEngagement,
 }
+
+#[derive(Deserialize, Debug, ToSchema, Clone)]
+#[serde(rename_all = "snake_case")]
+pub enum StreamReach {
```
tipogi, line 25: maybe `UserReach`? it describes how far a user can reach within their network

src/routes/v0/stream/users.rs
```diff
@@ -31,7 +34,10 @@ pub struct UserStreamQuery {
         ("skip" = Option<usize>, Query, description = "Skip N followers"),
         ("limit" = Option<usize>, Query, description = "Retrieve N followers"),
         ("source" = Option<UserStreamSource>, Query, description = "Source of users for the stream."),
-        ("depth" = Option<usize>, Query, description = "User trusted network depth, user following users distance. Numbers bigger than 4, will be ignored")
+        ("reach" = Option<StreamReach>, Query, description = "The target reach of the source. Supported in 'influencers' source."),
```
tipogi, line 37: Regarding _line 32_, in the source list are missing `most_followed` and `influencers

src/models/user/influencers.rs
tipogi: @amirRamirfatahi @SHAcollision, I still think the [same](https://github.com/pubky/pubky-nexus/pull/234/files#r1884041899) —we are adding unnecessary code without a clear need

src/models/user/influencers.rs
```diff
@@ -0,0 +1,173 @@
...
+        reach: StreamReach,
+        skip: usize,
+        limit: usize,
+        timeframe: &Timeframe,
+    ) -> Result<Option<Influencers>, DynError> {
+        let query = queries::get::get_influencers_by_reach(user_id, reach, skip, limit, timeframe);
+        retrieve_from_graph::<Influencers>(query, "influencers").await
+    }
+
+    async fn get_global_influencers(
```
tipogi, line 150: possibility to refactor

src/routes/v0/stream/users.rs
```diff
@@ -17,7 +17,10 @@ pub struct UserStreamQuery {
     skip: Option<usize>,
     limit: Option<usize>,
     source: Option<UserStreamSource>,
+    reach: Option<StreamReach>,
     depth: Option<u8>,
+    timeframe: Option<Timeframe>,
+    preview: Option<bool>,
```
tipogi, line 23: 🔥

tipogi (commented):
few suggested changes, are we are there 🚀

src/types/mod.rs
amirRamirfatahi: I named it StreamReach because it's the same for tags, users, posts, etc.  so it's any stream really.

src/models/user/influencers.rs
amirRamirfatahi: how's this comment helpful? Is there actually something messy or you just don't like the style? It's less than 10 lines of code so I don't even know what you mean here.

tipogi: sorry to not be helpful... I wanted to point out the comment that I let you above, that `Influencers` struct is not need it actually

src/models/user/influencers.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub async fn get_influencers(
+        user_id: Option<&str>,
+        reach: Option<StreamReach>,
+        skip: usize,
+        limit: usize,
+        timeframe: &Timeframe,
+        preview: bool,
+    ) -> Result<Option<Influencers>, DynError> {
+        let (skip, limit) = if preview {
+            let skip = Utc::now().timestamp_subsec_micros() % 98;
```
SHAcollision, line 119: Please add a comment here to explain this is a pseudo-random between 0 and 97. It's not straightforward obvious what we are doing.

src/models/user/influencers.rs
```diff
@@ -11,31 +11,25 @@ use utoipa::ToSchema;
...
-pub struct Influencer {
-    pub id: String,
-    pub score: f64,
-}
+const GLOBAL_INFLUENCERS_PREFIX: &str = "Cache";
 
 // Define a newtype wrapper
 #[derive(Serialize, Deserialize, Debug, ToSchema, Default, Clone)]
-pub struct Influencers(pub Vec<Influencer>);
+pub struct Influencers(pub Vec<(String, f64)>); // (user_id, score)
```
tipogi, line 18: what about if we create a new type for that (if you have better candidate for that, go for it)
```rust
pub type InfluencerScore = (String, f64)
```

tipogi (approved):
hop hop!! 🚀

src/models/user/influencers.rs
amirRamirfatahi: Utopia kinda sucks for type aliases. This would require us to add `utopia-config` crate and use something like this:
```
use utoipa_config::Config;

fn main() {
    Config::new()
        .alias_for("Influencer", "(String, f64)")
        .write_to_file();
}
```

PR #235, opened 2024-12-02
Fix user graph query handling of None values

amirRamirfatahi:
Fixes the put query for user using string "null" value instead of null.  

# Pre-submission Checklist


tipogi (approved):
good catch 🚀 merge it!

PR #239, opened 2024-12-05
feat: add post_replies source in user stream

### LINKED ISSUE #228: Feat (service): user stream of reply participants
Add a new `source=post_replies` to `/stream/users` with a new param `...&post_id={string}&author_id={string}` (mimicking current stream of posts of `PostReplies`).

Can be implemented with currently existing indexes. We just need to retrieve the Sorted set of replies of that `post_key` , subset the `user_id:` part of the found members and return unique `user_ids` to the `UserStream` model to craft a stream of user profiles for the existing. 

Alternatively, we could also simply return the `user_ids` and assume the client has already cached the profiles.

amirRamirfatahi:
This PR fixes #228 
# Pre-submission Checklist


src/models/user/stream.rs
```diff
@@ -257,6 +267,36 @@ impl UserStream {
...
+                    .map(|replies| {
+                        replies
+                            .into_iter()
+                            .map(|reply| reply.0.split(":").next().unwrap().to_string())
+                            .collect::<Vec<String>>()
+                    })
+                    .into_iter()
+                    .flatten()
+                    .collect();
+                Some(unique_user_ids.into_iter().collect())
```
SHAcollision, line 298: Avoid `unwrap()`

Untested, maybe something like.
```
   let unique_user_ids: HashSet<String> = replies
        .map(|replies_batch| {
            replies_batch
                .into_iter()
                .filter_map(|(raw_id, _some_other_data)| {
                    // Split on ":" and take the first piece; avoid `unwrap()`
                    raw_id.split(':').next().map(ToString::to_string)
                })
                .collect::<Vec<String>>()
        })
        .flatten()
        .collect();

File: src/models/user/stream.rs [hunk 2]
```diff
@@ -257,6 +267,36 @@ impl UserStream {
...
+            }
```
SHAcollision, line 299: Too much logic inside of this arm of the match statement. 

We should do the same as for the `UserStreamSource::Recommended` arm. Move the logic to the `get_post_replies_ids(post_id: &str, author_id: &str)` function.

SHAcollision (changes requested):
The PR looks good. There is a few things we can improve.

This PR still needs at least 1 more test and it would be very nice if we also add a benchmark for this new stream of users.

SHAcollision:
> This PR solves #228

Please use the keyword `Fixes` as in `Fixes #228` to automatically link this PR to the issue.

src/models/user/stream.rs
```diff
@@ -187,10 +213,55 @@ impl UserStream {
...
+        )
+        .await?;
+        let unique_user_ids: HashSet<String> = replies
+            .map(|replies| {
+                replies
+                    .into_iter()
+                    .map(|reply| {
+                        reply
+                            .0
+                            .split(":")
```
SHAcollision, line 246: I'm not 100% sure on this one. I think `reply.0.split(":").map(ToString::to_string).collect::<Vec<String>>()[0].clone()` could potentially panic if the reply string does not contain a colon (I know, it should). But instead, we could consider using iterator methods like `.next()` with proper error handling.

src/models/user/stream.rs
```diff
@@ -187,10 +213,55 @@ impl UserStream {
...
+                            .map(ToString::to_string)
+                            .collect::<Vec<String>>()[0]
+                            .clone()
+                    })
+                    .collect::<Vec<String>>()
+            })
+            .into_iter()
+            .flatten()
+            .collect();
+        Ok(Some(unique_user_ids.into_iter().collect()))
```
SHAcollision, line 256: NIT: There is a lot going on in the 2 last statements of this function. Maybe we can divide them into more lines for clarity.

NIT: It might be nice to document (small comment right here `//`) that we are using HashSet to enforce uniqueness, but the order of the repliers is not guaranteed. We could also sort the vector before returning (not worth it IMO, no requirement for ordering AFAIK).

SHAcollision, line 256: Maybe something like this is clearer and panic safe.
```suggestion
        // If there are replies, extract unique user IDs using a HashSet.
        let unique_user_ids: HashSet<String> = if let Some(replies) = replies {
            replies
                .into_iter()
                .filter_map(|reply| reply.0.split(':').next().map(|s| s.to_string()))
                .collect()
        } else {
            // If no replies are found, return None.
            return Ok(None);
        };

        // Convert the HashSet to a Vec. (Note: the ordering will be arbitrary.)
        Ok(Some(unique_user_ids.into_iter().collect()))
```

SHAcollision (commented):
This PR is basically ready to merge I only left a few minor NIT comments that I think are worth adressing.

src/models/user/stream.rs
```diff
@@ -23,6 +26,18 @@ pub enum UserStreamSource {
     MostFollowed,
     Pioneers,
     Recommended,
+    PostReplies,
+}
+
+pub struct UserStreamInput {
```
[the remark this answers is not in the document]
amirRamirfatahi: I don't know about "not improving clarity". The `clippy` rule is clearly there because when the number of args increases, it becomes hard to track them, while the verbosity + unsorted nature of struct helps with dealing these cases. Let's keep the struct as it is. The `clippy` rule is strict as is with 7 args, meaning we probably won't have many of these cases.

PR #247, opened 2024-12-10
feat: retry manager

tipogi:
## Description
This PR introduces a foundational implementation for the `RetryManager`. It includes:

Initialization of a message-passing system with a channel (`SenderChannel` and `ReceiverChannel`).
A `RetryEvent` structure to represent retryable events, including fields like`error_type` and `retry_count`.
##### Basic commands for:
- Adding failed events to a retry queue (`ProcessEvent`)
##### Motivation
The `RetryManager` is a core component to handle failed events during homeserver interaction. This initial implementation sets up the required structure and logic to:
- Track and store retryable events.

## Pre-submission Checklist


amirRamirfatahi:
2 initial comments to think about:
1. Can we use Redis as the messaging queue?  
2. being able to identify ghost(?) users.

src/events/mod.rs
```diff
@@ -86,15 +86,16 @@ impl Event {
             "PUT" => EventType::Put,
             "DEL" => EventType::Del,
             _ => {
-                return Err(EventProcessorError::InvalidEvent {
-                    message: format!("Unknown event type: {}", parts[0]),
+                return Err(EventProcessorError::InvalidEventLine {
+                    message: format!("Unknown event type, {}", line),
```
SHAcollision, line 90: This Error is thrown when the method is neither `"PUT"` or `"DEL"` , should we leave the description with `parts[0]` (the unexpected method) instead of the full `line` ?

src/events/mod.rs
```diff
@@ -104,34 +105,58 @@ impl Event {
             },
             _ if uri.contains("/posts/") => ResourceType::Post {
                 author_id: parsed_uri.user_id,
-                post_id: parsed_uri.post_id.ok_or("Missing post_id")?,
+                post_id: parsed_uri
+                    .post_id
+                    .ok_or(EventProcessorError::InvalidEventLine {
+                        message: format!("Missing post_id, {} {}", parts[0], uri),
```
SHAcollision, line 111: Can we do this instead for all of the messages, after all `parts[0] + uri == line` 
```
format!("Missing post_id, {}", line),
```

src/events/mod.rs
```diff
@@ -104,34 +105,58 @@ impl Event {
...
-                file_id: parsed_uri.file_id.ok_or("Missing file_id")?,
+                file_id: parsed_uri
+                    .file_id
+                    .ok_or(EventProcessorError::InvalidEventLine {
+                        message: format!("Missing file_id, {} {}", parts[0], uri),
+                    })?,
             },
             _ if uri.contains("/blobs") => return Ok(None),
             _ if uri.contains("/last_read") => return Ok(None),
             _ if uri.contains("/settings") => return Ok(None),
```
SHAcollision, line 156: We also have 
```
            _ if uri.contains("/feeds") => return Ok(None),
```
It's a known type in `pubky-app-specs` but we are ignoring in Nexus.

src/events/mod.rs
tipogi: good catch!

src/watcher.rs
```diff
@@ -19,5 +43,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Sync + Send>> {
         }
         // Wait for X milliseconds before fetching events again
         sleep(Duration::from_millis(config.watcher_sleep)).await;
+        retry_failed_events += 1;
+
+        if RETRY_THRESHOLD == retry_failed_events {
+            let sender = event_processor.sender.lock().await;
+            let _ = sender
+                .send(SenderMessage::RetryEvent(
```
amirRamirfatahi, line 51: why the `let _`?  
This will fail silently if the `.send` fails and essentially breaks the retry manager.  
It's better to have this throw the exception if necessary.

src/events/handlers/user.rs
```diff
@@ -24,7 +25,15 @@ pub async fn sync_put(user: PubkyAppUser, user_id: PubkyId) -> Result<(), DynErr
     // Create UserDetails object
     let user_details = UserDetails::from_homeserver(user, &user_id).await?;
     // SAVE TO GRAPH
-    user_details.put_to_graph().await?;
+    match user_details.put_to_graph().await {
+        Ok(_) => (),
+        Err(e) => {
```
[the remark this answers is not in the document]
tipogi: might be better, `map_err` because `is_err()` discards the error details

src/events/processor.rs
```diff
@@ -96,13 +115,13 @@ impl EventProcessor {
                 let event = match Event::parse_event(line) {
                     Ok(event) => event,
                     Err(e) => {
-                        error!("Error while creating event line from line: {}", e);
+                        error!("{}", e);
```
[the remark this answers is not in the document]
tipogi: because the error itself has a error type and might be too repetitive? for that error type `InvalidEventLine: {message}` the error output would be:
```
InvalidEventLine: Malformed event line, PUT pubky://xxx....
```

src/events/processor.rs
```diff
@@ -111,35 +130,55 @@ impl EventProcessor {
...
+                        error!("Unhandled error type for URI: {}, {:?}", event.uri, e);
                         return Ok(());
                     }
+                };
+
+                // Generate a compress index to save in the cache
+                let index = match RetryEvent::generate_index_key(&event.uri) {
+                    Some(retry_index) => retry_index,
+                    None => {
+                        // Unlikely to be reached, as it would typically fail during the validation process
```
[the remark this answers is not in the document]
tipogi: I would say that in general you hate commenting 🤣

src/events/retry/event.rs
```diff
@@ -0,0 +1,145 @@
...
+        let parts: Vec<&str> = event_uri.split('/').collect();
+        // Ensure the URI structure matches the expected format
+        if parts.first() != Some(&HOMESERVER_PROTOCOL)
+            || parts.get(3) != Some(&HOMESERVER_PUBLIC_REPOSITORY)
+            || parts.get(4) != Some(&HOMESERVER_APP_REPOSITORY)
+        {
+            return None;
+        }
+
+        match parts.as_slice() {
```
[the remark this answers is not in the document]
tipogi: for example how would you do in a better way to get the `pubky`, `domain` of the event and the `event_id`?

src/events/retry/event.rs
```diff
@@ -0,0 +1,145 @@
...
+            _ => None,
+        }
+    }
+
+    /// Generates the opposite event line by replacing the event type, effectively
+    /// converting delete operations into put operations
+    ///
+    /// # Arguments
+    /// * `event_line` - A string slice representing the event line.
+    pub fn generate_opposite_event_line(event_line: &str) -> String {
```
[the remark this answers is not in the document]
tipogi: deleted as it is not in use

src/events/retry/event.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Stores an event in both a sorted set and a JSON index in Redis.
+    /// It adds an event line to a Redis sorted set with a timestamp-based score
+    /// and also stores the event details in a separate JSON index for retrieval.
+    /// # Arguments
+    /// * `event_line` - A `String` representing the event line to be indexed.
+    pub async fn put_to_index(&self, event_line: String) -> Result<(), DynError> {
+        Self::put_index_sorted_set(
+            &RETRY_MANAGER_EVENTS_INDEX,
+            // NOTE: Don't know if we should use now timestamp or the event timestamp
```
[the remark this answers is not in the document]
tipogi: as we are now just storing the fail events in the queue, that time value is not that important, in the future would be. When we will start retrying all the events, we will need to come back and think what it would be the best solution. From now, I think it is ok

src/events/retry/event.rs
```diff
@@ -0,0 +1,145 @@
...
+        let index = &[RETRY_MANAGER_STATE_INDEX, [&event_line]].concat();
+        self.put_index_json(index, None).await?;
+
+        Ok(())
+    }
+
+    /// Checks if a specific event exists in the Redis sorted set
+    /// # Arguments
+    /// * `event_index` - A `&str` representing the event index to check
+    pub async fn check_uri(event_index: &str) -> Result<Option<isize>, DynError> {
```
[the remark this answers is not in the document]
tipogi: yes, thats true! it does not make sense that if 🤣 the reason of that it is to avoid repetition in the watcher tests

src/events/retry/event.rs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            return Ok(Some(post_details));
+        }
+        Ok(None)
+    }
+
+    /// Retrieves an event from the JSON index in Redis based on its index
+    /// # Arguments
+    /// * `event_index` - A `&str` representing the event index to retrieve
+    pub async fn get_from_index(event_index: &str) -> Result<Option<Self>, DynError> {
```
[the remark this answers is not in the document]
tipogi: true! same reason as above and when we will start retrying events, these both function are going to be useful

src/events/retry/event.rs
```diff
@@ -0,0 +1,145 @@
...
+        if let Some(fail_event) = Self::try_from_index_json(index).await? {
+            found_event = Some(fail_event);
+        }
+        Ok(found_event)
+    }
+
+    /// Deletes event-related indices from the retry queue in Redis
+    /// # Arguments
+    /// - `event_line` - A `String` representing the event line that needs to be unindexed from the retry queue
+    pub async fn delete(&self, event_line: String) -> Result<(), DynError> {
```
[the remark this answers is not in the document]
tipogi: deleted!

src/events/handlers/post.rs
```diff
@@ -40,10 +42,24 @@ pub async fn sync_put(
...
         OperationOutcome::Updated => true,
-        // TODO: Should return an error that should be processed by RetryManager
-        // WIP: Create a custom error type to pass enough info to the RetryManager
-        OperationOutcome::Pending => {
-            return Err("WATCHER: Missing some dependency to index the model".into())
+        OperationOutcome::MissingDependency => {
+            let mut dependency = Vec::new();
+            if let Some(replied_uri) = &post_relationships.replied {
+                let reply_dependency = RetryEvent::generate_index_key(replied_uri)
+                    // This block is unlikely to be reached, as it would typically fail during the validation process
```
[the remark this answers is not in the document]
tipogi: When an unexpected error occurs, we do not generate the correct index; instead, we simply clone the URI as is using `unwrap_or_else(|| reposted_uri.clone())`. Secondly, we do not want to trigger an error just because a post dependency URL is not formatted as expected. While we acknowledge that this may not be the ideal solution, it is effective

src/events/retry/event.rs
```diff
@@ -0,0 +1,145 @@
...
+pub const RETRY_MANAGER_EVENTS_INDEX: [&str; 1] = ["events"];
+pub const RETRY_MANAGER_STATE_INDEX: [&str; 1] = ["state"];
+pub const HOMESERVER_PROTOCOL: &str = "pubky:";
+pub const HOMESERVER_PUBLIC_REPOSITORY: &str = "pub";
+pub const HOMESERVER_APP_REPOSITORY: &str = "pubky.app";
+
+/// Represents an event in the retry queue and it is used to manage events that have failed
+/// to process and need to be retried
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct RetryEvent {
```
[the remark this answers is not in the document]
tipogi: from now the retry time, it is not need it as we are not retrying and the creation time we get from the sorted set.

src/events/retry/manager.rs
```diff
@@ -0,0 +1,94 @@
...
+
+    /// Executes the main event loop to process messages from the channel
+    /// This function listens for incoming messages on the receiver channel and handles them
+    /// based on their type:
+    /// - **`SenderMessage::Retry`**: Triggers the retry process
+    /// - **`SenderMessage::Add`**: Queues a failed event in the retry cache for future processing
+    ///
+    /// The loop runs continuously until the channel is closed, ensuring that all messages
+    /// are processed appropriately.
+    pub async fn exec(&self) -> Result<(), DynError> {
```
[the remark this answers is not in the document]
tipogi: yep, this is a great idea, makes more sense 👍

src/events/retry/manager.rs
```diff
@@ -0,0 +1,94 @@
...
+        // WIP: Retrieve the homeserver events from the SORTED SET
+        // RetryManager:events
+    }
+
+    /// Stores the event line in the Redis cache, adding it to the retry queue for
+    /// future processing
+    /// # Arguments
+    /// - `index_key`: A `String` representing the compacted key for the event to be stored in Redis
+    /// - `retry_event`: A `RetryEvent` instance containing the details of the failed event
+    async fn queue_failed_event(
```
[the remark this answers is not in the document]
tipogi: Since we are following the "do not speculate about the future" pattern, deleted! However, I think, this function may gain additional functionalities in the future

src/events/uri.rs
```diff
@@ -21,54 +23,72 @@ impl TryFrom<&str> for ParsedUri {
...
+                message: format!("The Event URI is malformed,  {}", uri),
+            }
+            .into());
         }
 
         // Extract post_id if present
-        if let Some(post_id) = extract_segment(uri, "/posts/", "/") {
-            parsed_uri.post_id = Some(post_id.to_string());
-        }
+        parsed_uri.post_id = extract_segment(uri, "/posts/", "/")
```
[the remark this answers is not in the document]
tipogi: The previous snippets were incorrect when the id was empty. As a result, no error was thrown, and it attempted to fetch the event, which was incorrect. For example, `pubky://xxxxx/pub/pubky.app/posts` was populating `parsed_uri.post_id` as `Some("")` and should have triggered an InvalidEventLine error. Now, it does.

src/watcher.rs
```diff
@@ -19,5 +43,15 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Sync + Send>> {
         }
         // Wait for X milliseconds before fetching events again
         sleep(Duration::from_millis(config.watcher_sleep)).await;
+        retry_failed_events += 1;
```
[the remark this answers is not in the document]
tipogi: In this implementation, it is not needed, I forgot to delete it 👍. My initial intention was to retry all failed events after fetching some events

src/watcher.rs
tipogi: yep, might be better approach but now it does not exist (deleted already)

tests/watcher/follows/retry_follow.rs
```diff
@@ -0,0 +1,93 @@
...
+    match event_state.error_type {
+        EventProcessorError::MissingDependency { dependency } => {
+            assert_eq!(dependency.len(), 1);
+            assert_eq!(dependency[0], dependency_uri)
+        }
+        _ => assert!(false, "The error type has to be MissingDependency type"),
+    };
+
+    test.del(&follow_url).await?;
+    tokio::time::sleep(Duration::from_millis(500)).await;
```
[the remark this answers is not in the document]
tipogi: When we write failed events to the cache (through the retry manager channel), it is possible that other operations (like read the cache) execute faster than the write. That's why we temporarily block the thread to ensure consistency. Effective and works

tests/watcher/bookmarks/retry_bookmark.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    assert_eq!(event_state.retry_count, 0);
+
+    let dependency_uri = format!("{fake_user_id}:posts:{fake_post_id}");
+    match event_state.error_type {
+        EventProcessorError::MissingDependency { dependency } => {
+            assert_eq!(dependency.len(), 1);
+            assert_eq!(dependency[0], dependency_uri);
+        }
+        _ => assert!(false, "The error type has to be MissingDependency type"),
```
SHAcollision, line 68: Instead of `assert!(false, ...)`, we could do `panic!(...)` . `assert!(false, ...)` works, but `panic!` might read a bit clearer.

tests/watcher/follows/retry_follow.rs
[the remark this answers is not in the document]
SHAcollision: Two ideas to avoid the large sleeps in tests:
1. Add a `.drain()` method to the retry manager:
We could expose a method like retry_manager.drain(), which blocks until all queued events are processed. This way, tests can explicitly wait for operations to finish.

2. Use a synchronous test entry point:
For tests, we could call the manager’s logic directly via a synchronous helper method (e.g., `put_event_synchronously`). This bypasses concurrency for test purposes, so we get deterministic results while testing the core logic.

Alternatively, much more dirty, we could do a loop of `sleep(10)` with some check for the new state, with a timeout after 1 second, failing the test.

SHAcollision (approved):
Hey @tipogi !This looks good. I didn't super dive in depth on the logic, but trusting @amirRamirfatahi here to have done the homework better than me.

tests/watcher/utils/watcher.rs
```diff
@@ -268,3 +270,32 @@ pub async fn retrieve_and_handle_event_line(event_line: &str) -> Result<(), DynE
 
     Ok(())
 }
+
+/// Attempts to read an event index with retries before timing out
+/// # Arguments
+/// * `event_index` - A string slice representing the index to check
+pub async fn try_until_write(event_index: &str) {
```
SHAcollision, line 277: Maybe more clear naming is `assert_eventually_exists` or `wait_for_event_index` .

tests/watcher/utils/watcher.rs
```diff
@@ -268,3 +270,32 @@ pub async fn retrieve_and_handle_event_line(event_line: &str) -> Result<(), DynE
...
+    const MAX_RETRIES: usize = 50;
+
+    for attempt in 0..MAX_RETRIES {
+        info!(
+            "RetryEvent: Trying to read index {:?}, attempt {}/{} ({}ms)",
+            event_index,
+            attempt + 1,
+            MAX_RETRIES,
+            SLEEP_MS * attempt as u64
+        );
```
SHAcollision, line 288: Probably too spammy. Let's keep it `debug!` or only log it with `info!()` once the condition is fulfilled

tests/watcher/utils/watcher.rs
```diff
@@ -268,3 +270,32 @@ pub async fn retrieve_and_handle_event_line(event_line: &str) -> Result<(), DynE
 
     Ok(())
 }
+
+/// Attempts to read an event index with retries before timing out
+/// # Arguments
+/// * `event_index` - A string slice representing the index to check
+pub async fn try_until_write(event_index: &str) {
+    const SLEEP_MS: u64 = 20;
```
SHAcollision, line 278: We get full write/response from Redis in 1ms. Not sure what's the extra time for the messaging and processing in this case. My guess is that this will always succeed on the first loop if it's as loose as 20ms (at least locally, on GH hosted runners a bit slowlier). I would set this anywhere 1-5 ms if we want it to be tight/useful.

Edit: in fact, it seems even in GH hosted runner it succeeds on a single loop:
```
[2025-01-29T16:32:24Z INFO  r#mod::watcher::utils::watcher] RetryEvent: Trying to read index "PUT:h3mggwij9mgx4afzyg8wh8urttqhtie75138dd1mtkpcci4qw5oo:bookmarks:VHNQGPN6QSKXXY72S8BS7AD6GW", attempt 1/50 (0ms)
```

SHAcollision (approved):
To me it looks good to merge, wdyt @amirRamirfatahi ? Is there anything we can further refine?

tests/watcher/utils/watcher.rs
tipogi: set to `3ms` the sleep time

tipogi: ok the last test failed! I will add tomorrow `assert_eventually_exists` function

src/events/processor.rs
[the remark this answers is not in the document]
amirRamirfatahi: I'm saying if I run into this `InvalidEventLine: Malformed event line, PUT pubky://xxx....`, I know what error has been thrown, but I won't know which line. But, if I run into `Error while creating event line from line: InvalidEventLine: Malformed event line, PUT pubky://xxx....` I know I can search for `Error while creating event line from line` and find the exact line.

src/events/processor.rs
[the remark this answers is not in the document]
amirRamirfatahi: fair enough.  
There's a reason for it though.

src/events/retry/event.rs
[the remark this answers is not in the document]
amirRamirfatahi: what do you mean by needed?  Used as a `.created_at`? or a `.last_retried_at`? sure.  
But needed in monitoring? needed to know what happens if something goes wrong? I think it's much better to foresee these things than not have it later.

src/events/retry/event.rs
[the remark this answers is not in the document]
amirRamirfatahi: you're not extracting these vars.  You're just doing a check to return a value here.  
Why not `match parts.last()` for example?

src/events/handlers/post.rs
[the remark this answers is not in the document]
amirRamirfatahi: I know about that. I can see it because I read the code and that's obvious.  
I'm asking why the comment?! How's this helpful when reading the code? Or, in what scenario would this comment be useful to have here?

src/events/retry/manager.rs
```diff
@@ -0,0 +1,73 @@
...
+/// - Ensures failed events are reprocessed in an orderly manner
+pub struct RetryManager {}
+
+/// Initializes a new `RetryManager` with a message-passing channel
+impl RetryManager {
+    pub fn init_channels() -> (ReceiverChannel, SenderChannel) {
+        let (tx, rx) = mpsc::channel(CHANNEL_BUFFER);
+        (
+            Arc::new(Mutex::new(rx)), // Receiver channel
+            Arc::new(Mutex::new(tx)), // Sender channel
```
amirRamirfatahi, line 46: why are you wrapping these in an Arc<Mutex<>>?  
There's almost no reason to do this. Sender channel is already ready to be cloned and receiver doesn't need it because it's a single consumer.  
Please remove them.

src/db/connectors/pubky.rs
```diff
@@ -2,76 +2,60 @@ use crate::Config;
...
+                            .map_err(|e| PubkyConnectorError::ClientError(e.to_string()))?,
+                        false => Client::new()
+                            .map_err(|e| PubkyConnectorError::ClientError(e.to_string()))?,
+                    },
+                };
+
+                Ok(Arc::new(pubky_client))
+            })
+            .await
+            .map(|_| ())
```
amirRamirfatahi, line 51: This isn't really readable.  
I know it looks smart but it's also unnecessary.

amirRamirfatahi: Actually I think you should remove this whole function. Move the 
```
.get_or_try_init(|| async {
                let pubky_client = match testnet {
                    Some(testnet) => Client::builder()
                        .testnet(testnet)
                        .build()
                        .map_err(|e| PubkyConnectorError::ClientError(e.to_string()))?,
                    None => match config.testnet {
                        true => Client::testnet()
                            .map_err(|e| PubkyConnectorError::ClientError(e.to_string()))?,
                        false => Client::new()
                            .map_err(|e| PubkyConnectorError::ClientError(e.to_string()))?,
                    },
                };

                Ok(Arc::new(pubky_client))
            })
            .await
```

into `get_pubky_client` and just return that.

src/events/processor.rs
```diff
@@ -111,35 +133,55 @@ impl EventProcessor {
...
-                                "Error while handling event: {}. Retrying attempt {}/{}",
-                                e, attempts, self.max_retries
-                            );
-                            // Optionally, add a delay between retries
-                            tokio::time::sleep(Duration::from_millis(100)).await;
-                        }
-                    } else {
-                        error!("PROCESSOR: Sending the event to RetryManager... Missing node(s) and/or relationship(s) to execute PUT or DEL operation(s)");
+    async fn handle_event(&self, event: Event) -> Result<(), DynError> {
+        match event.clone().handle().await {
```
amirRamirfatahi, line 137: could be a .map_err instead of match?

src/events/processor.rs
```diff
@@ -111,35 +133,55 @@ impl EventProcessor {
...
+                        // Unlikely to be reached, as it would typically fail during the validation process
+                        return Ok(());
+                    }
+                };
+                let index_key = format!("{}:{}", event.event_type, index);
+
+                // Send event to the retry manager
+                let sender_channel = self.retry_manager_sender_channel.lock().await;
+                match sender_channel
+                    .send(RetryQueueMessage::ProcessEvent(index_key, retry_event))
```
amirRamirfatahi, line 166: Right now, if the retry manager thread panics, the receiver is basically dropped, meaning no new message will be consumed, and once it hits the buffer size, this line will block the main thread forever, meaning we stop indexing new events!  
One option is to use `try_send` here, which will raise an error, in the above case, and then you can rerun the retry manager thread, making sure that both indexing and retrying are continued without any problem.

src/events/retry/manager.rs
```diff
@@ -0,0 +1,73 @@
...
+    // Command to retry all pending events that are waiting to be indexed
+    RetryEvent(String),
+    /// Command to process a specific event in the retry queue
+    /// This action can either add an event to the queue or use the event as context to remove it from the queue
+    /// - `String`: The index or identifier of the event being processed
+    /// - `RetryEvent`: The event to be added to or processed for removal from the retry queue
+    ProcessEvent(String, RetryEvent),
+}
+
+pub type SenderChannel = Arc<Mutex<Sender<RetryQueueMessage>>>;
```
amirRamirfatahi, line 25: as this is public and you're importing it elsewhere, please rename to something like RetryManagerSenderChannel, because the type includes `RetryQueueMessage`.  
Alternatively, removes this because `Arc<Mutex<>>` is not needed as stated in another comment, and `Sender<RetryQueueMessage>` is then short enough not to need a separate type.

src/events/retry/manager.rs
```diff
@@ -0,0 +1,73 @@
...
+///
+/// ## Responsibilities:
+/// - Receives and processes messages related to event retries
+/// - Maintains a queue of events that need to be retried
+/// - Ensures failed events are reprocessed in an orderly manner
+pub struct RetryManager {}
+
+/// Initializes a new `RetryManager` with a message-passing channel
+impl RetryManager {
+    pub fn init_channels() -> (ReceiverChannel, SenderChannel) {
```
amirRamirfatahi, line 42: Thanks for refactoring the previous version, but this is still not the best approach. It's causing unnecessary cloning.
Let's refactor to this:
```
pub struct RetryManager {
    sender: Sender<RetryQueueMessage>
}
pub fn init() -> RetryManager {}
pub async fn process_messages() {}
```
And then use it like this:
```
let retry_manager = RetryManager::init();
let x = X::new(retry_manager.sender.clone());
retry_manager.start();
```

src/db/connectors/pubky.rs
tipogi: I was thinking to do the same but after, each time that we call `get_pubky_client` we will need to pass the config and testnet argument

tipogi: openned an [issue, #306](https://github.com/pubky/pubky-nexus/issues/306). I think it is out of scope that topic here and maybe we can find better solution if we focus just in the connector

src/events/mod.rs
```diff
@@ -1,201 +1,172 @@
...
+            response = match pubky_client.get(&self.uri).send().await {
+                Ok(response) => response,
+                Err(e) => {
+                    return Err(EventProcessorError::PubkyClientError {
+                        message: format!("{}", e),
+                    }
+                    .into())
+                }
+            };
+        } // drop the pubky_client lock
```
amirRamirfatahi, line 107: does pubky_client have a lock? isn't it Arc<Client>?

src/events/mod.rs
```diff
@@ -1,201 +1,172 @@
...
-            Ok(response) => response,
-            Err(e) => {
-                error!("WATCHER: Failed to fetch content at {}: {}", self.uri, e);
-                return Err(e.into());
-            }
-        };
         let blob = response.bytes().await?;
+        let resource = self.parsed_uri.resource;
+
+        // Use the new importer from pubky-app-specs
```
amirRamirfatahi, line 112: is there a reason for this comment? At what point does the importer stop being new? :))

amirRamirfatahi (commented):
added some minor changes needed. Still some older comments are unanswered.  
Apart from this, I thought this PR is going to handle retrying the events too? Did we just remove the older retry logic in favor of just storing + no retrying? why? If we were only gonna store them, then why all the channel complexity? We have a receiver that only is going to `X::put_to_index()`? Couldn't we do this in the main thread that's already processing the event?

PR #249, opened 2024-12-12
tests: refactor service tests as standalone

### LINKED ISSUE #172: Tests(service): self-contained tests
Currently, all tests for the `service` need `cargo run` running on another terminal.

We could turn our axum server into a mod we can import and run on the tests themselves within a once cell. This is of course not so easy because tokyo kills it when the first test finishes, but there is probably a solution.

This would both:
- make it easier to run the test suite locally as only `cargo test` is needed
- make it faster to run the test suite in the GH workflows as we skip the `cargo run` build.

amirRamirfatahi:
Fixes #172
# Pre-submission Checklist


tests/service/stream/post/author_replies.rs
```diff
@@ -2,7 +2,7 @@ use super::ROOT_PATH;
 use crate::service::utils::make_request;
 use anyhow::Result;
 
-#[tokio::test]
+#[tokio_shared_rt::test(shared)]
```
SHAcollision, line 5: What are the benefits for `service` tests of using the shared tokio runtime?

tests/utils.rs
```diff
@@ -0,0 +1,102 @@
...
+            .exec_async(&mut redis_conn)
+            .await
+            .expect("Failed to flush Redis");
+
+        // Reindex
+        let should_reindex = redis_is_empty().await.unwrap_or(false);
+        if should_reindex {
+            info!("Starting reindexing process.");
+            reindex().await;
+        }
```
SHAcollision, line 75: This logic is not needed. We can do `reindex().await`

tests/utils.rs
```diff
@@ -0,0 +1,102 @@
...
+        graph
+            .lock()
+            .await
+            .run(drop_all_query)
+            .await
+            .expect("Could not drop graph nodes.");
+
+        // Run the run-queries.sh script on the Docker host using docker exec
+        tokio::process::Command::new("docker")
+            .args(&["exec", "neo4j", "bash", "/db-graph/run-queries.sh"])
```
SHAcollision, line 51: There is a few trade offs with this approach as:
- We don't know where neo4j is (maybe not docker, maybe different containter name (as in `pubky-docker`)
- Very long waits when you simply want to run a single test case often during development

Is there a way we can run only `sync_graph()` optionally? I am thinking maybe we can have a more granular control envvar than `SYNC_DB`, maybe an explicit `SYNC_GRAPH` variable ? We already know whether to reindex or not with the `REINDEX` envvar, this way we control both independently.

tests/utils.rs
```diff
@@ -0,0 +1,102 @@
...
+            reindex().await;
+        }
+    }
+
+    async fn start_server() -> Result<()> {
+        let config = Config::from_env();
+        setup(&config).await;
+
+        // make sure DBs are in sync with mock data
+        let sync_db_env = std::env::var("SYNC_DB").unwrap_or("false".to_string());
```
SHAcollision, line 83: Reminder to add the new envvars to `.env-sample` with documentation of that they do

SHAcollision (commented):
It's looking good! Now that you are at it, could you also improve `.github/workflows/test.yml` to eliminate the steps that are not needed anymore? (building and running the service in the background, syncing the neo4j data, waiting for it to be available, etc).

tests/utils.rs
amirRamirfatahi: Will see what I can do.  
About your first point, that's very true. It was just a ton easier to do it this way for now. After all, it's for local tests and pipelines which won't be changing often.

tests/utils.rs
amirRamirfatahi: I'll fix the docs. This doesn't necessarily need to be in .env-sample after it's mentioned in the how to run the tests parts of the README.

SHAcollision:
@amirRamirfatahi let me know when this is ready for re-review

tests/utils.rs
SHAcollision: > About your first point, that's very true. It was just a ton easier to do it this way for now. After all, it's for local tests and pipelines which won't be changing often.

@amirRamirfatahi Has this been handled? so we can move this PR. It breaks the way we develop against `pubky-docker` and this is the recommended way and the way I use.

src/mock_db.rs
tipogi, line 1: What about if we move that to `src/db/mock_db.rs` as it is related with db operations?

tests/utils.rs
amirRamirfatahi: yes new commits also handle this.

src/mock_db.rs
amirRamirfatahi: since it's a bin I'd rather really have it there.  
How much do you insist on it? :D

PR #253, opened 2024-12-16
feat: add graph return control for all ingestion events

tipogi:
# Pre-submission Checklist

### PR description

The initial implementation was risky due to the lack of clear feedback when an event failed to index in the graph. This made troubleshooting difficult. We’ve improved observability and can now identify specific failure scenarios, such as missing relationships in the graph. Knowing this, we can avoid ineffective retries and instead take alternative actions for such cases.

One proposed solution is to implement a `RetryManager`([#247](https://github.com/pubky/pubky-nexus/pull/247)), which would manage retries intelligently by analyzing failures and determining the best course of action—whether retrying after resolving dependencies or flagging for manual intervention. This approach enhances both the reliability and efficiency of the indexing process


src/db/graph/queries/put.rs
```diff
@@ -55,42 +53,98 @@ pub fn create_post(post: &PostDetails) -> Result<Query, DynError> {
...
+    cypher.push_str("
+        // Check if all the dependencies are consistent in the graph
+        MATCH (parent_author:User {id: $parent_author_id})-[:AUTHORED]->(parent_post:Post {id: $parent_post_id})
+        MATCH (author:User {id: $author_id})
+        // Check if the post already exists
+        OPTIONAL MATCH (existing_post:Post {id: $post_id}) 
+    ");
+
+    // Create the generic part of the query depending the post type
+    if action == "replies" {
```
SHAcollision, line 84: Let's use a Match and instead of matching a string "replies" and "repost" let's make "action" and Enum.

src/db/graph/queries/put.rs
```diff
@@ -26,28 +26,24 @@ pub fn create_user(user: &UserDetails) -> Result<Query, DynError> {
...
-                new_relationships.push("MERGE (new_post)-[:REPLIED]->(reply_parent_post)");
-            },
-            PostInteraction::Reposts(_) => {
-                cypher.push_str("
-                    MATCH (repost_parent_author:User {id: $repost_parent_author_id})-[:AUTHORED]->(repost_parent_post:Post {id: $repost_parent_post_id})
-                ");
-                new_relationships.push("MERGE (new_post)-[:REPOSTED]->(repost_parent_post)");
-            }
-        }
+    if let Some(_) = &post_relationships.replied {
```
SHAcollision, line 36:
```
    if post_relationships.replied.is_some() {
```

src/db/graph/queries/put.rs
```diff
@@ -26,28 +26,24 @@ pub fn create_user(user: &UserDetails) -> Result<Query, DynError> {
...
+        cypher.push_str("
+            MATCH (reply_parent_author:User {id: $reply_parent_author_id})-[:AUTHORED]->(reply_parent_post:Post {id: $reply_parent_post_id})
+        ");
+        new_relationships.push("MERGE (new_post)-[:REPLIED]->(reply_parent_post)");
+    };
+    if let Some(_) = &post_relationships.reposted {
```
SHAcollision, line 42:
```
    if post_relationships.reposted.is_some() {
```

src/models/post/relationships.rs
```diff
@@ -88,6 +89,23 @@ impl PostRelationships {
         }
     }
 
+    /// Constructs a `Self` instance by extracting relationships from a `PubkyAppPost` object
+    pub fn get_from_homeserver(post: &PubkyAppPost) -> Self {
```
SHAcollision, line 93: `from_homeserver()` if we want to match the convention we are using in other models

src/models/post/relationships.rs
```diff
@@ -88,6 +89,23 @@ impl PostRelationships {
...
+
+        let mut relationship = Self::default();
+
+        if let Some(parent_uri) = &post.parent {
+            relationship.replied = Some(parent_uri.to_string());
+        }
```
SHAcollision, line 99: This `if` can be removed and simplified as:
```
        relationship.replied = post.parent.clone();
```

src/events/handlers/post.rs
```diff
@@ -104,44 +99,55 @@ pub async fn sync_put(
             )
             .await?;
         }
+         // Populate the reply parent keys to after index the reply
```
SHAcollision, line 102: > // **Populate** the... to **after**....

Suggest to improve this comment to:
**Define** the... to index the reply **later**...

Applies to other comments that contain "to after".

SHAcollision (commented):
It's looking cleaner!

src/db/graph/exec.rs
```diff
@@ -13,18 +13,19 @@ pub async fn exec_single_row(query: Query) -> Result<(), DynError> {
 }
 
 // Exec a graph query that has a single "boolean" return
-pub async fn exec_boolean_row(query: Query) -> Result<bool, DynError> {
+pub async fn exec_boolean_row(query: Query) -> Result<Option<bool>, DynError> {
```
SHAcollision, line 16: Have we documented this well? For example, it can be document the implicit usage we are making of the returned type of this function to control for non existing objects.

> 
> This function is usually used for queries that will return a row where the response is:
> - None: Some dependency is missing (e.g., a reply's parent).
> - Some(true): The node/relationship already existed, and from this, we deduce that it is an EDIT.
> - Some(false): The node/relationship did not exist, so we have created a new node or relationship.

SHAcollision (commented):
We are now implicitly using the `Option<bool>` returned by most graph queries to have a stricter control of what did really happend when we made a graph query. However `Option<bool>` is not very expressive even though we are making a very sistematic usage of it. This will be hard to understand by anyone reading the code, and even hard for us when we add new queries or revisit existing queries in the future.

Do you think there's a way to make this even clearer and more obvious in the  1. Queries themselves, 2. the way we execute them, and 3. the returned type (something more expressive than Option<bool>) ? 

In addition, for instance, `exec_boolean_row` could be renamed to something even clearer for its new very clear purpose. As long as we are consistent in the way we use this function to execute queries.

src/db/graph/exec.rs
tipogi: We could use the following enum

```rust
pub enum QueryResult {
    Pending,      // None: Some dependency is pending
    Edited,       // Some(true): The node/relationship existed (an edit was performed)
    Created,      // Some(false): The node/relationship did not exist (a new one was created)
}
```

src/db/graph/exec.rs
```diff
@@ -12,20 +12,43 @@ pub async fn exec_single_row(query: Query) -> Result<(), DynError> {
...
+
+    if let Some(row) = result.next().await? {
+        // The "flag" field indicates a specific condition in the query
+        match row.get("flag")? {
+            true => Ok(OperationOutcome::Updated),
+            false => Ok(OperationOutcome::Created),
+        }
+    } else {
+        Ok(OperationOutcome::Pending)
     }
```
SHAcollision, line 51:
```
match result.next().await? {
    Some(row) => match row.get("flag")? {
        true => Ok(OperationOutcome::Updated),
        false => Ok(OperationOutcome::Created),
    },
    None => Ok(OperationOutcome::Pending),
}
```

tests/watcher/utils/watcher.rs
```diff
@@ -211,3 +223,20 @@ impl WatcherTest {
         Ok(mute_url)
     }
 }
+
+/// Retrieves an event from the homeserver and handles it asynchronously.
+/// # Arguments
+/// * `event_line` - A string slice that represents the URI of the event to be retrieved
+///   from the homeserver. It contains the event type and the homeserver uri
+pub async fn retrieve_event_from_homeserver(event_line: &str) -> Result<(), DynError> {
```
SHAcollision, line 231:
```suggestion
pub async fn retrieve_and_handle_event_line(event_line: &str) -> Result<(), DynError> {
```

tests/watcher/tags/fail_index.rs
```diff
@@ -0,0 +1,116 @@
...
+        image: None,
+        links: None,
+        name: "Watcher:CannotTag:Tagged:Sync".to_string(),
+        status: None,
+    };
+    let tagged_user_id = test.create_user(&tagged_keypair, &tagged_user).await?;
+
+    // Switch OFF the event processor to simulate the pending events to index
+    // In that case, shadow user
+    test = test.remove_event_processing().await;
```
SHAcollision, line 24: smart

tests/watcher/tags/fail_index.rs
```diff
@@ -0,0 +1,116 @@
...
+        "pubky://{}/pub/pubky.app/tags/{}",
+        shadow_user_id,
+        tag.create_id()
+    );
+
+    // PUT user tag
+    test.put(tag_url.as_str(), tag_blob).await?;
+
+    // Create raw event line to retrieve the content from the homeserver. Event processor is deactivated
+    // Like this, we can trigger the error in that test
```
SHAcollision, line 57: `// Like this, we can trigger the error in that test`
clarify, like what? and what test?

tests/watcher/tags/fail_index.rs
```diff
@@ -0,0 +1,116 @@
...
+    let sync_fail = retrieve_event_from_homeserver(&tag_event)
+        .await
+        .map_err(|e| {
+            error!("SYNC ERROR: {:?}", e);
+        })
+        .is_err();
+
+    assert!(
+        sync_fail,
+        "Cannot exist the tag because it is not in sync the graph with events"
```
SHAcollision, line 112: Clarify

SHAcollision (approved):
Woop, woop!! This is true quality upgrade. Hoping to see much fewer partial synchronization issues from now on :partying_face:

Issue #143, opened 2024-10-08
Discussion: add a reindex (or migrations) binary to run on breaking index changes


SHAcollision:
For example when adding new fields to `UserCounts` existing indexes will be invalid and cause a 500 error. We can run a `reindex` bin script that will update all indexes and sync them.

amirRamirfatahi:
1. Two-Phase Write
The two-phase write approach involves writing data to both the old and new schemas simultaneously during the migration process, ensuring both systems stay in sync. This means every write operation is duplicated—whether it's creating, updating, or deleting data—so both the old and new schemas are consistently updated. The downside is the complexity and overhead involved in managing dual writes, especially when Neo4j and Redis don’t support cross-system transactions. This increases the risk of data inconsistencies if one system fails while the other succeeds, and it can also create performance bottlenecks due to the extra workload, though it's temporary as is the whole migration process. **This is the golden standard and what's most widely used.**

2. Chunked Migration
Chunked migration breaks the data into smaller, manageable chunks and migrates each chunk independently. After migrating each chunk, you validate it to ensure everything is correct before moving on to the next one. This method allows for more control over the migration, making it easier to isolate and fix issues. However, it can be time-consuming, as each chunk needs to be processed and validated separately. Coordinating changes during the migration can also be tricky, especially when there are dependencies between chunks, but it reduces the risk of large-scale failures by migrating in smaller, isolated steps.

3. Snapshot Migration
Snapshot migration takes a full snapshot of the data, transforms it to fit the new schema, and re-imports it into the system. This approach allows for a clean break, as you can validate the transformed data before switching over to the new schema. However, it usually requires downtime or a freeze on writes while the snapshot is taken and processed, which can be a significant drawback for systems requiring high availability. Additionally, if something goes wrong during the snapshot or transformation process, recovery can be complex, making this method less ideal for systems that need to stay online during the migration.

4. Event-Sourced-Based Migration
In event-sourced-based migration, you replay all events from an event log (**Homeserver events**) to rebuild the data in the new schema. The advantage is strong consistency and the ability to rebuild the system state from scratch if needed. However, replaying a large number of events can be slow. If the event log is incomplete or has issues, it can lead to inconsistencies in the migrated data.

amirRamirfatahi:
1. Neo4j Two-Phase Write Approaches
a) Migration Manager (Application Layer)
[...]

PR #255, opened 2024-12-17
Add migration manager

amirRamirfatahi:
#143 
# Pre-submission Checklist


SHAcollision:
Is this inteded only for Redis keys/cache migration?

If so, can we first create an MVP "migration" with the minimal amount of complexity and code keeping it extremely simple?

What's the overall flow and usage? I'd say we can do something straight forward. E.g:
I'd say we need to version the Redis schemas. The current version will always be availble as a Redis value. On service/watcher setup (start-up), we check whether current version is the latest. If it's not latest, apply migrations.

README.md
```diff
@@ -87,6 +88,41 @@ To contribute to Nexus, follow these steps:
...
+
+#### Adding a new migration
+
+To create a new migration, use the migrations binary by running:
+
+```
+cargo run --bin migrations new MigrationNAME
+```
+
+This will generate a new migration file in the `migrations/` directory.
```
tipogi, line 117: maybe be more specific of the migration path? `src/db/migrations/migrations_list`

README.md
```diff
@@ -87,6 +88,41 @@ To contribute to Nexus, follow these steps:
...
+
+#### Understanding Migration Phases
+
+The Migration Manager uses a phased approach to handle data migrations safely and systematically. Each phase serves a distinct purpose in transitioning data from the old source to the new source, ensuring consistency and minimal disruption. Here's an overview of the phases:
+
+- **Dual Write**: During this phase, all writes to the old source are mirrored to the new source. This ensures that both sources remain synchronized during normal application operations. Developers invoke MigrationManager::dual_write in the application logic for this purpose. Once dual writes are stable and verified, the migration can progress to the next phase.
```
tipogi, line 99: > application logic
application data layer logic?

README.md
```diff
@@ -87,6 +88,41 @@ To contribute to Nexus, follow these steps:
...
+  **Note**: Mark a migration as ready for backfill phase using the `MIGRATIONS_BACKFILL_READY` env var.
```
tipogi, line 100: The migration has to be referenced by migration `id`

README.md
```diff
@@ -87,6 +88,41 @@ To contribute to Nexus, follow these steps:
...
+#### Adding a new migration
+
+To create a new migration, use the migrations binary by running:
+
+```
+cargo run --bin migrations new MigrationNAME
+```
+
+This will generate a new migration file in the `migrations/` directory.  
+Next, register your migration in the `get_migration_manager` function, which ensures it is included in the migration lifecycle. Once registered, implement the required phases (dual_write, backfill, cutover, and cleanup) in the generated file. Each phase serves a specific purpose in safely transitioning data between the old and new sources. After implementing your migration, run the migrations binary to execute pending migrations:
```
tipogi, line 118: > `get_migration_manager`

`get_migration_manager` (`src/db/migrations/mod.rs` file)

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+    fn to_string(&self) -> &str {
+        match self {
+            MigrationPhase::DualWrite => "dual_write",
+            MigrationPhase::Backfill => "backfill",
+            MigrationPhase::Cutover => "cutover",
+            MigrationPhase::Cleanup => "cleanup",
+            MigrationPhase::Done => "done",
+        }
+    }
+}
```
tipogi, line 39: Do we really need that implementation having the `Serialize` trait from Serde??
```rust
serde_json::to_string(&MigrationPhase::Backfill).unwrap()
```

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+            migrations: Vec::new(),
+        }
+    }
+
+    pub async fn dual_write<T: Migration>(data: Box<dyn Any + Send>) -> Result<(), DynError> {
+        T::dual_write(data).await
+    }
+
+    pub async fn new_migration(name: &str) -> Result<(), DynError> {
+        let migration_name = format!("{}{}", name, Utc::now().timestamp());
```
tipogi, line 102: we could calculate once and add to the names the timestamp, instead of generate twice

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+        let migration_file_name = format!("{}_{}", to_snake_case(name), Utc::now().timestamp());
+        let migration_template = get_migration_template(migration_name.as_str());
```
tipogi, line 104: we could use the reference `&` instead of `as_str()`

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+        let file_path = format!(
+            "src/db/migrations/migrations_list/{}.rs",
+            migration_file_name.as_str()
+        );
+        tokio::fs::write(file_path.clone(), migration_template).await?;
```
tipogi, line 109: Maybe throw more customised error:

```rust 
.map_err(|e| format!("Failed to write migration file at {}: {}", file_path, e))?;
```

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+
+        // append to migrations_list/mod.rs
+        let mod_file_path = "src/db/migrations/migrations_list/mod.rs";
+        let mod_file_content = format!("pub mod {};\n", migration_file_name);
+        let mut mod_file = tokio::fs::OpenOptions::new()
+            .append(true)
+            .open(mod_file_path)
+            .await?;
+        tokio::io::AsyncWriteExt::write_all(&mut mod_file, mod_file_content.as_bytes()).await?;
```
tipogi, line 118: Another option is to do in that way and also throw some errors meanwhile:

```rust
OpenOptions::new()
        .append(true)
        .open(mod_file_path)
        .await
        .map_err(|e| format!("Failed to open mod.rs at {}: {}", mod_file_path, e))?
        .write_all(mod_file_content.as_bytes())
        .await
        .map_err(|e| format!("Failed to write to mod.rs at {}: {}", mod_file_path, e))?;
```

Here if the `File` (which is the return type of `open`) implements the `tokio::io::AsyncWriteExt` trait, we could chain it all the operation in one

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+#[derive(Deserialize, Serialize, Clone)]
+pub struct MigrationNode {
+    id: String,
+    phase: MigrationPhase,
+    created_at: i64,
+    updated_at: i64,
+}
+
+pub struct MigrationManager {
+    graph: Arc<Mutex<Graph>>,
```
tipogi, line 85: is it some reason to not use already the helpers functions instead of have a graph instance: 
- `exec_single_row`
- `retrieve_from_graph`
Like this we will not need to create more boilerplate code. I think with that two functions would be enough. Also with default trait would be enough for `MigrationManager`

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+                .migrations_backfill_ready
+                .contains(&stored_migration.id)
+            {
+                self.update_migration_phase(&stored_migration.id, &MigrationPhase::Backfill)
+                    .await?;
+            }
+        }
+
+        // get all migrations from the database
+        let stored_migrations = self.get_migrations().await?;
```
tipogi, line 142: could not be better to edit the migration instance instead of calling again to the database to get the migration?

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+                    let new_migration_node = MigrationNode {
+                        id: migration_id.to_string(),
+                        phase: MigrationPhase::Backfill,
+                        created_at: Utc::now().timestamp_millis(),
+                        updated_at: Utc::now().timestamp_millis(),
+                    };
+                    stored_migration = Some(new_migration_node.clone());
+                }
+            }
+            let stored_migration = stored_migration.unwrap();
```
tipogi, line 167: ⚠️ unsafe

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+            if stored_migration.phase == MigrationPhase::Done {
+                println!("Migration {} is already done", migration_id);
+                continue;
+            }
+            println!(
```
tipogi, line 172: maybe instead of `println!` use `info!`?

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+            "MERGE (m:Migration {id: $id}) SET m.phase = $phase, m.updated_at = timestamp()"
+                .to_string(),
+        )
+        .param("id", id)
+        .param("phase", phase.to_string());
+
+        self.graph.lock().await.run(query).await?;
+        Ok(())
+    }
+}
```
tipogi, line 255: using already commented functions, these above three functions could refactor at least in two functions

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+        assert_eq!(super::to_snake_case("PascalCase"), "pascal_case");
+        assert_eq!(super::to_snake_case("snake_case"), "snake_case");
+        assert_eq!(super::to_snake_case("kebab-case"), "kebab-case");
+        assert_eq!(super::to_snake_case("UPPERCASE"), "uppercase");
+        assert_eq!(super::to_snake_case("lowercase"), "lowercase");
+        assert_eq!(super::to_snake_case("12345"), "12345");
+        assert_eq!(super::to_snake_case("snake_case_123"), "snake_case_123");
+        assert_eq!(super::to_snake_case("UserNewField"), "user_new_field");
+    }
+}
```
tipogi, line 336: I would create another file (`utils.rs`, `generetor.rs`, ...) for that kind of utility functions

src/db/migrations/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+use crate::MigrationManager;
+
+pub mod manager;
+mod migrations_list;
+
+pub fn get_migration_manager(graph: Arc<Mutex<Graph>>) -> MigrationManager {
+    // let migration_manager = MigrationManager::new(graph);
+    // migration_manager.register(Box::new(MigrationX));
+    MigrationManager::new(graph)
+}
```
tipogi, line 15: Document a bit that one

tipogi (changes requested):
Some clean up need it but It looks good that migration skeleton

src/db/migrations/manager.rs
amirRamirfatahi: isn't this much heavier than just implementing Display?

src/db/migrations/manager.rs
amirRamirfatahi: I don't think there's any difference at all. I took the convention to use .as_str() as it's much cleaner IMO.

src/db/migrations/manager.rs
```diff
@@ -0,0 +1,336 @@
...
+        T::dual_write(data).await
+    }
+
+    pub async fn new_migration(name: &str) -> Result<(), DynError> {
+        let migration_name = format!("{}{}", name, Utc::now().timestamp());
+        let migration_file_name = format!("{}_{}", to_snake_case(name), Utc::now().timestamp());
+        let migration_template = get_migration_template(migration_name.as_str());
+        let file_path = format!(
+            "src/db/migrations/migrations_list/{}.rs",
+            migration_file_name.as_str()
```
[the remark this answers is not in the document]
amirRamirfatahi: I want to use the same string again. So instead of cloning it, I burrowed it here.

src/db/migrations/manager.rs
amirRamirfatahi: I don't think this improves the readability at all either.

src/db/migrations/manager.rs
amirRamirfatahi: We don't know what's gonna be the required functions to run inside a migration. Most frequently, you're gonna want to start transactions and commit them when you're done so migrations should have total freedom.

src/db/migrations/manager.rs
amirRamirfatahi: These two steps are separate from each other. There's no harm in updating the DB and getting from it again.

src/db/migrations/manager.rs
amirRamirfatahi: This is not in the server. So we do actually want to panic if this ever fails. So it's not unsafe.

src/db/migrations/manager.rs
amirRamirfatahi: some of them need to stay but I changed the rest.

src/db/migrations/manager.rs
amirRamirfatahi: I don't know what you mean? Are you talking about `exec_single_row`?  
I'd rather not mix these two parts together. Think of it as more like a library you'd use.

src/db/migrations/manager.rs
amirRamirfatahi: As I said I want to keep the migration separate from server code and this function then only would be used here so it's fine to keep it local to the file.

src/db/migrations/mod.rs
amirRamirfatahi: Added a line but it's already documented in the DB and there's also an example for you.

src/db/migrations/manager.rs
tipogi: I do not see the sense to keep repeating code. We will see if it is useful to do it in that way in the future...

src/db/migrations/manager.rs
tipogi: Do we really want to `panic()` and stop also other ongoing migrations? I do not think so...

src/db/migrations/manager.rs
tipogi: unnessesary

src/db/migrations/manager.rs
tipogi: I am referring to the content of the function. From _240-247_ and _255-263_ are the same code. It could be part of the `utils.rs` file as I mention above as `update_migration_phase`

src/db/migrations/manager.rs
tipogi: it would be under `db/migrations/utils.rs`

src/db/migrations/manager.rs
amirRamirfatahi: yes we want to panic as the migrations are run sequentially and if one fails the process should stop

tipogi (approved):
Once the errors in the watcher integration tests are fixed, this PR will be ready for merging 🚀

PR #260, opened 2024-12-20
Fix: user search duplicate bug

### LINKED ISSUE #259: Bug (watcher): changing username duplicates the user on the search by username response
Every time we receive an updated profile.json that changes the username we are still leaving behind the old `username:pubky_id` key on the `Sorted:Users:Name` sorted set used to search lexicographically for user pubky id's by their username. 

It leads to duplicated items in the stream of users by searched username
image

amirRamirfatahi:
Fixes #259 

# Pre-submission Checklist


src/models/user/search.rs
```diff
@@ -87,4 +111,92 @@ impl UserSearch {
...
     }
+
+    async fn delete_existing_records(user_ids: &[&str]) -> Result<(), DynError> {
+        if user_ids.is_empty() {
+            return Ok(());
+        }
+        let mut records_to_delete: Vec<String> = Vec::with_capacity(user_ids.len());
+        for user_id in user_ids {
+            let existing_record =
+                Self::try_from_index_hashmap(&USER_NAME_HASHMAP_KEY_PARTS, user_id).await?;
```
SHAcollision, line 122: Why not simply do `UserDetails::get_bt_id(user_id)` to figure what this user's current username (if any) is?

SHAcollision (changes requested):
The proposed solution to this bug works. 

However, the solution is overly complex. Adding a new Hashmap to find Usernames by pubky_id does not seem necessary. This lookup table is not needed as we already have methods to retrieve a username of any user by its pubky_id. We already have `User:Details:{user_id}` keys that we can access with `UserDetails::get_by_id(pubky_id).username` . This is enough to figure what `username:pubky_id` (if any) we should remove from the sorted set.

src/models/user/search.rs
```diff
@@ -57,8 +58,19 @@ impl UserSearch {
...
     pub async fn put_to_index(details_list: &[&UserDetails]) -> Result<(), DynError> {
+        // ensure existing records are deleted
+        Self::delete_existing_records(
+            details_list
+                .iter()
+                .map(|details| details.id.0.as_str())
+                .collect::<Vec<&str>>()
+                .as_slice(),
+        )
+        .await?;
```
SHAcollision, line 69: Nice.

src/models/user/search.rs
```diff
@@ -72,9 +84,21 @@ impl UserSearch {
             // The value in the sorted set will be `username:user_id`
             let member = format!("{}:{}", username, user_id.0);
 
+            hashmap_items.push((user_id.0.clone(), username.clone()));
```
SHAcollision, line 87: Not really needed, unless we can proof some sort of performance gain by maintaining this hashmap vs retrieving the usernames from the existing Redis JSON per user

src/models/user/search.rs
amirRamirfatahi: `UserDetails::get_bt_id(user_id)` will try to set the UserSearch index if it faces an index miss, leading to an infinite loop situation.  This will be a separate index that UserSearch itself is holding and maintaining, with no outer exposure, and it's precisely because we're performing the search like this.  
We can move the whole thing to a full text search index on the graph, which is my suggested way of doing search anyway. + it gives you a more sane searching experience by having things like fuzzy search.
For now, the extra hashmap seems to be worth it.

amirRamirfatahi: `let user_details_list = UserDetails::get_from_index(keys).await?;` 
at this point the index and graph have both been updated. This requires its own storage to hold the previous username. We could do without anything extra if zrangebylex supported suffix searches, but it doesn't. We need to know which records are already there for the user id.  We can either use the hashmap, or use another in reverse for this query like `{id}:{username}`.  
I love the sentiment about being lean but this one hashmap (because of how we're doing search) is not going to haunt us.  
Again, the correct thing, is to have these searches on the graph db with a full_text search index on it and cache the most searched queries.

SHAcollision: We are always doing `UserDetails::put_to_index()` after `UserSearch::put_to_index()`.

If you think we should do full text search, please open an issue for it, we discuss it.

src/models/user/details.rs
```diff
@@ -71,7 +71,10 @@ where
                 .map_err(serde::de::Error::custom)?;
             Ok(Some(urls))
         }
-        _ => Err(serde::de::Error::custom("Expected a string or an array")),
+        serde_json::Value::Null => Ok(None),
+        _ => Err(serde::de::Error::custom(
+            "Expected either a string, an array or null",
+        )),
```
SHAcollision, line 77: I guess you figured something here. What was going on?

src/db/kv/index/json.rs
```diff
@@ -358,22 +358,24 @@ pub async fn get_multiple<T: DeserializeOwned + Send + Sync>(
...
-                serde_json::from_str::<Vec<T>>(&value_str)
-                    .ok()
-                    .and_then(|vec| vec.into_iter().next())
-            })
+        .map(|value_str| match value_str {
+            Some(value) => {
+                let value: Vec<T> = serde_json::from_str(&value)?;
+                Ok(value.into_iter().next())
+            }
+            None => Ok(None),
```
SHAcollision, line 378: We might have been falling back to graph all this time?

src/db/kv/index/hashmap.rs
```diff
@@ -0,0 +1,69 @@
...
+///
+/// # Errors
+///
+/// Returns an error if the operation fails.
+pub async fn del_index_hashmap(prefix: &str, key: &str, field: &str) -> Result<(), DynError> {
+    let mut conn = get_redis_conn().await?;
+    let index_key = format!("{}:{}", prefix, key);
+    let _: () = conn.hdel(index_key, field).await?;
+    Ok(())
+}
```
SHAcollision, line 69: We can remove `src/db/kv/index/hashmap.rs` and the 2 new `traits.rs` methods from this PR, but add them to a new draft PR in case we need to use Redis HashMaps in the future.

SHAcollision (commented):
Any idea why the Files tests are failing?

src/models/user/details.rs
amirRamirfatahi: links is an option, but if it was null in the DB, the deserialization was failing.

src/db/kv/index/json.rs
amirRamirfatahi: Only those that had a failed deserialization, as we were failing silently.  
I don't think there were that many.

src/db/kv/index/hashmap.rs
amirRamirfatahi: why though? I understand because it's not being used anywhere XD but still I don't think there's any harm in having it merged rn.

SHAcollision: We would need to review / test / mantain it. Better don't go through that until there is a need.

SHAcollision (approved):
Looks really good! Let's merge!

Thanks a lot @amirRamirfatahi not only for fixing the duplicated user search obvious bug, but most importantly for discovering a serialization bug and fixing it making the code safer (no silent failures on JSON errors), more consistent with standard Serde patterns, and fix an important bug in the user indexing logic :rocket:

PR #271, opened 2024-12-27
Feat: logging open telemetry

### LINKED ISSUE #270: Feat: improve logging and observability of Nexus
Right now our debugging ability is really hampered due to our limited visibility into the server.  
We need standard, robust logging and tracing capabilities to ensure we can effectively debug and/or measure metrics for our backend.

amirRamirfatahi:
Fixes #270 
# Pre-submission Checklist


src/routes/middlewares/tracing.rs
```diff
@@ -0,0 +1,38 @@
...
+use opentelemetry::{
+    global::{self, ObjectSafeSpan},
+    trace::{FutureExt, Status, TraceContextExt, Tracer},
+    Context, KeyValue,
+};
+
+// middleware for tracing
+pub async fn tracing_middleware(request: Request, next: Next) -> Response {
+    let tracer = global::tracer("nexus.service");
+    let mut span = tracer.start("REST API");
```
SHAcollision, line 12: `"REST API"`

All operations are now called `REST API` . Does it make sense if, for easier sorting, we use the endpoint called here `request.uri().path().to_string()` (only top levels, avoiding path params) instead of looking for them by filtering by tags like `"http.route="/v0/user/` ?

For example, internal jaeger tracing is easily subsetable based on "operation" (`tracer.start("operation name")`) 
image
However, nexus service only has one operation for all endpoints.
image

SHAcollision (commented):
Hey @amirRamirfatahi I have not dived into the codebase changes but I have been testing the functionality with [Jaeger all-in-one](https://hub.docker.com/r/jaegertracing/all-in-one/tags). I've read this might be a good relatively lightweight metrics backend to use in local development open telemetry I pushed a commit to add it. Feel free to replace by anything else you think it's better.

To make it work, I had to change the port from `4318` to `4317` (gRPC). I think the less dependencies we use the better, so if there is any dependency we can avoid by using HTTP instead of gRPC, that's be great.

SHAcollision:
Please can you share results with/without of our benchmarks running un-dockerized Redis with and without this feature?

src/routes/middlewares/tracing.rs
amirRamirfatahi: @SHAcollision I used https://signoz.io/ and it was working out of the box.  
Also keeping the operation name the same gives us some nice averages and numbers. if you look a couple lines below it, we have 
```
span.set_attribute(KeyValue::new(
        "http.route",
        request.uri().path().to_string(),
    ));
```

so we still have the filtering, etc capabilities.

tipogi: I've run the docker container of `signoz` and run that command to run the service binary but no any info in the dashboard, should it be like that?

```bash
# Tried with 4318 port and also with service.name=nexus
OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317 OTEL_RESOURCE_ATTRIBUTES=service.name=nexus.service cargo run
```

[signoz docker setup](https://signoz.io/docs/install/docker/)

src/events/processor.rs
```diff
@@ -49,11 +51,19 @@ impl EventProcessor {
...
-        let Some(lines) = self.poll_events().await? else {
-            return Ok(());
+        let lines = {
+            let tracer = global::tracer("nexus.watcher");
+            let span = tracer.start("Polling Events");
+            let cx = Context::new().with_span(span);
+            self.poll_events()
+                .with_context(cx)
+                .await
+                .unwrap_or_default()
```
tipogi, line 61: We cannot continue here with the execution if we do not get any events. Also doing `unwrap_or_default()`, we are loosing the error the we might throw in PubkyClient. Usually are that kind of events

```
ERROR watcher] Uncaught error occurred while processing events: PubkyClientError { message: "Some(hyper_util::client::legacy::Error(Connect, ConnectError(\"dns error\", FailedToResolveEndpoint)))" }
```
In that case, it would be because the homeserver is not resolved

PR #294, opened 2025-01-17
chore: refactor hot tags endpoint

tipogi:
# Pre-submission Checklist


src/routes/v0/tag/global.rs
```diff
@@ -72,10 +119,10 @@ pub async fn tag_taggers_handler(
...
         ("user_id" = Option<String>, Query, description = "User Pubky ID"),
         ("reach" = Option<TagStreamReach>, Query, description = "Reach type: follower | following | friends"),
-        ("taggers_limit" = Option<usize>, Query, description = "Retrieve N user_id for each tag"),
-        ("skip" = Option<usize>, Query, description = "Skip N tags"),
-        ("limit" = Option<usize>, Query, description = "Retrieve N tag"),
-        ("timeframe" = Option<Timeframe>, Query, description = "Retrieve hot tags for this specific timeframe. Defaults to all_time"),
+        ("taggers_limit" = Option<usize>, Query, description = "Retrieve N user_id for each tag. Defaults to `20`"),
+        ("skip" = Option<usize>, Query, description = "Skip N tags. Defaults to `0`"),
+        ("limit" = Option<usize>, Query, description = "Retrieve N tag. Defaults to `40`"),
+        ("timeframe" = Option<Timeframe>, Query, description = "Retrieve hot tags for this specific timeframe (not applied for reach). Defaults to `all_time`"),
```
amirRamirfatahi, line 125: what do you mean not applied for reach?

src/db/graph/queries/get.rs
```diff
@@ -432,16 +407,25 @@ pub fn get_tag_taggers_by_reach(
...
-            WITH reach, MAX(tag.indexed_at) AS tag_time
-            ORDER BY tag_time DESC
-            SKIP $skip LIMIT $limit
-            RETURN COLLECT(reach.id) as tagger_ids
+
+            // Get the latest tagged timestamp per `reach` user
+            WITH DISTINCT reach, MAX(tag.indexed_at) AS latest_tag_time
+            ORDER BY latest_tag_time DESC
+
+            // Use slice notation instead of SKIP and LIMIT
```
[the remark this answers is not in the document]
tipogi: Yes, if we use `SKIP` and `LIMIT` together, the `SKIP` clause does not work as it has to. That's why I went with slice notation, it is not the optimal but it works

src/routes/v0/tag/global.rs
tipogi: We cannot apply `reach` and `timeframe` together because the query does not allow. Maybe better approach is to fix that adding in the `WHERE` clause that conditions (https://github.com/pubky/pubky-nexus/blob/c9dc5a9cc076e87e3a998298dddc1908f48eca4c/src/db/graph/queries/get.rs#L418). The hottags endpoint description for timeframe is wrong (C&P seems, L70)

src/db/graph/queries/get.rs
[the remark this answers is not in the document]
amirRamirfatahi: yeah the original question meant to ask why would skip and limit work?

src/routes/v0/tag/global.rs
amirRamirfatahi: this is the hot tags handler?  The line you're mentioning is for tag_taggers.  
The hot tags query can very well handle both together.

tipogi: yes, this is the hot tag taggers endpoint swagger description but in the hot tags the description for the timeframe is wrong because it can handle as you said, both.

tipogi: Yes, I think I mixed up the endpoint. I thought you were referring to the taggers. In this case, the line is for hot tags, so I'll correct it 👍

src/db/kv/traits.rs
```diff
@@ -46,24 +46,21 @@ pub trait RedisOps: Serialize + DeserializeOwned + Send + Sync {
...
+    /// * `prefix` - An optional string representing the prefix for the Redis keys. If `Some(String)`, the prefix will be used
+    /// * `key_parts` - A slice of string slices that represent the parts used to form the key under which the value is stored
+    /// * `expiration` - An optional `i64` specifying the TTL (in seconds) for the set. If `None`, no TTL will be set.
     ///
     /// # Errors
     ///
     /// Returns an error if the operation fails, such as if the Redis connection is unavailable.
     async fn put_index_json(
         &self,
+        prefix: Option<String>,
```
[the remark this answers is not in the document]
tipogi: corrected here, https://github.com/pubky/pubky-nexus/pull/294/commits/dc4cdfcfe3326a3c0238c24f24da6f51e4daf52d

src/routes/v0/tag/global.rs
tipogi: Corrected the Swagger endpoint parameter description and also covered the missing integration test for the hot tags endpoint, https://github.com/pubky/pubky-nexus/pull/294/commits/42af710d11c61d073c039c1173ccc461b8ec1265

tests/service/tags/hot.rs
```diff
@@ -598,16 +672,18 @@ async fn test_hot_tags_by_friends_reach_and_all_timeframe() -> Result<()> {
 
 const PUBKY_TAG: &str = "pubky";
 
+// KEEP ON EYE in the list. Sometimes it changes the order. No reason
```
amirRamirfatahi, line 675: this isn't really good. If the order changes we have two options:
1. Fix the database queries to make sure the order is set so it's not random.
2. In the test, evaluate the list in a way that the order won't matter.

tests/service/tags/wot.rs
```diff
@@ -196,3 +200,24 @@ fn verify_user_taggers(mock_taggers: Vec<&str>, tag_details: Value, tag: String)
         );
     }
 }
+
+async fn clear_wot_tags_cache() -> Result<(), DynError> {
```
amirRamirfatahi, line 242: I'd rather have you delete everything that starts with `Cache:User:Taggers`

tests/service/tags/hot.rs
amirRamirfatahi: how's that going to solve it?  Also why does it do 2 requests? First one seems to not pass params but the second one is passing a skip, limit?  
I don't understand why.

Issue #209, opened 2024-11-19
Feat (dex): multi-event processor


SHAcollision:
The Nexus watcher should process events for every Homeserver object we create concurrently.

PR #314, opened 2025-02-04
Feat: Add watcher support indexing multiple homeservers

### LINKED ISSUE #200: Feat (dex): permanently store homeservers as graph nodes
Super speculative and low priority task:

We currently have a [Homserver model ](https://github.com/pubky/pubky-nexus/blob/main/src/models/homeserver.rs) that we use to store `pubky` `url` and `cursor`  for the homeserver in Redis. Storing them into the graph as well might be useful but there are some trade-offs. Note that `cursor` is expected to not be permanent: this way if we lose Redis data, we can sync again from the last cursor stored in Redis. 

Maybe there is a use to know which user is hosted in what server when it comes to moderating spamming homeservers, etc. So optionally, maybe we can add a "HOSTED" relationship between User and Homeserver nodes. But: 1. we do not have a use for it 2. makes the graph larger / harder to traverse 3. will be hard to keep up to date as users migrate homeserver.

amirRamirfatahi:
Fixes #200 and #209 
# Pre-submission Checklist

### DIFF
--- a/src/db/graph/queries/put.rs
+++ b/src/db/graph/queries/put.rs
@@ -1,3 +1,4 @@
+use crate::models::homeserver::Homeserver;
 use crate::models::post::PostRelationships;
 use crate::models::{file::FileDetails, post::PostDetails, user::UserDetails};
 use crate::types::DynError;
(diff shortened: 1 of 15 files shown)


src/db/graph/queries/put.rs
```diff
@@ -309,3 +310,20 @@ pub fn create_file(file: &FileDetails) -> Result<Query, DynError> {
 
     Ok(query)
 }
+
+// Create a Homeserver node
+pub fn create_homeserver(homeserver: &Homeserver) -> Result<Query, DynError> {
```
SHAcollision, line 315: I thought about storing Homeserver into Neo4J also when I implemented it. I think the main issue with of storing `Homeserver` in the graph is a possible dataloss of indexes on crash. 

In case of restart/crash, we might restore the `cursor` from graph instead of Redis. Redis last snapshot might not be the latest cursor on graph, therefore, we will resume watching homeservers and skip some events.

amirRamirfatahi: Hmmm...I don't know if I understood that correctly or not, but I'm also updating the cursor in graph?

SHAcollision: Yes exactly the problem: the cursor in graph might be newer than the latest synced events in Redis in case of a restart/crash. 

In other words, our assumption has always been that using Redis without persistence is OK, if and only if when we lose Redis data we also lose the cursor: this way, in the next startup the lost events will be re-acquired. However, saving the cursor to graph breaks this guarantee.

PR #165, opened 2024-10-15
Feat: files processing

amirRamirfatahi:
# Pre-submission Checklist

> For tests to work you need a working neo4j and redis instance with the example dataset in `docker/db-graph`

- [ ] **Testing**: Implement and pass new tests for the new features/fixes, `cargo test`.
- [ ] **Performance**: Ensure new code has relevant performance benchmarks, `cargo bench`

PR #322, opened 2025-02-05
Feat: add user avatar endpoint

### LINKED ISSUE #319: Feat: optimized user avatar endpoint
To ease and speed up page loading in pubky app we can serve avatars size optimized bypassing the `/file` fetch by serving on a route that only requires knowledge of the `user_id` .

```rust
pub const USER_AVATAR_ROUTE: &str = concatcp!(USER_ROUTE, "/avatar");
```

This is an idea for implementation:

When a request is received we can:
1. Check for the `UserDetails` `image` field. 
2. Find the `file` by uri internally.
3. Serve the image file as response.

Ideally we keep a ~256px  `.webp` optimized avatar already preprocessed to be served. So this feature likely requires of: https://github.com/pubky/pubky-nexus/pull/165

SHAcollision:
Fixes #319

It's a prototype, but works.

TODO:
- [ ] Use #165 optimizations
- [ ] Unified pattern: all of the file serving endpoints should use either 1) Axum handlers or 2) middleware. Preferably an openapi documented endpoint.
- [ ] Simplify and reuse better.
- [x] Write tests.
- [ ] Write benchmark.

# Pre-submission Checklist


src/routes/v0/user/avatar.rs
```diff
@@ -0,0 +1,92 @@
...
+    // 7. Build Axum `Response` setting the correct Content-Type
+    let mut response = Response::new(data.into());
+    response.headers_mut().insert(
+        CONTENT_TYPE,
+        HeaderValue::from_str(&content_type)
+            .unwrap_or_else(|_| HeaderValue::from_static("application/octet-stream")),
+    );
+
+    Ok(response)
+}
```
SHAcollision, line 88: Alternatively, not serving `content-type` header, makes this much shorter as we don't need to retrieve the `FileDetails` and (might?) work?
```
pub async fn user_avatar_handler(Path(user_id): Path<String>) -> Result<Response> {
    info!("GET {USER_AVATAR_ROUTE} user_id:{}", user_id);

    // 1. Get user details
    let details = match UserDetails::get_by_id(&user_id)
        .await
        .map_err(|source| Error::InternalServerError { source })?
    {
        None => return Err(Error::UserNotFound { user_id }),
        Some(d) => d,
    };

    // 2. Check if user has image. If not, 404
    let Some(image_uri) = details.image else {
        return Err(Error::FileNotFound {});
    };

    // 3. Parse user_id + file_id from the "pubky://owner_id/file_id" style URI
    let parsed_uri = match ParsedUri::try_from(image_uri.as_str()) {
        Ok(parsed_uri) => parsed_uri,
        Err(e) => return Err(Error::InternalServerError { source: e.into() }),
    };

    // 4. Build the actual path to the file on disk
    let file_path = match parsed_uri.resource {
        Resource::File(file_id) => {
            let config = Config::from_env();
            format!("{}/{}/{}", config.file_path, parsed_uri.user_id, file_id)
        }
        _ => return Err(Error::FileNotFound {}),
    };

    // 5. Read the file bytes from disk
    let data = match tokio::fs::read(&file_path).await {
        Ok(buf) => buf,
        Err(err) => {
            warn!("Reading avatar file failed: {err}");
            return Err(Error::FileNotFound {});
        }
    };

    Ok(Response::new(data.into()))
}
```
Specially if we assume `image/webp` always

PR #327, opened 2025-02-07
feat: add relationship boolean in all tags related endpoints

tipogi:
# Pre-submission Checklist


src/models/tag/traits/taggers.rs
```diff
@@ -29,7 +29,7 @@ where
         pagination: Pagination,
         viewer_id: Option<&str>,
         depth: Option<u8>,
-    ) -> Result<Option<Taggers>, DynError> {
+    ) -> Result<Option<(Taggers, bool)>, DynError> {
```
SHAcollision, line 32:
```
/// # Returns
    /// A result containing a vector of taggers or an error.
```
Let's update the docstring so it is clearer what the new `bool` is used for

src/models/tag/traits/taggers.rs
```diff
@@ -45,16 +45,26 @@ where
...
-    ) -> Result<Option<Taggers>, DynError> {
-        Ok(Self::try_from_index_set(&key_parts, skip, limit, prefix).await?)
+    ) -> Result<Option<(Taggers, bool)>, DynError> {
+        let taggers = Self::try_from_index_set(&key_parts, skip, limit, prefix).await?;
+        if let Some(taggers) = taggers {
+            let mut is_member = false;
+            if let Some(member) = viewer_id {
+                let (_, member_found) = Self::check_set_member(&key_parts, member).await?;
+                is_member = member_found;
+            }
```
SHAcollision, line 64: NIT I think we could do something like this too to avoid the `let mut is_member = false;`, but same thing, just a matter of preference/style
```rust
          let is_member = match viewer_id {
                Some(member) => {
                    Self::check_set_member(&key_parts, member).await?.1
                }
                None => false,
            };
```

SHAcollision (approved):
I didn't dive deep, but overall the new logic seems to tackle achieve the wanted functionality without introducing much chaos.  I am not sure whether our existing implementation of `get_multiple_sets()` is optimal or there is better ways to construct the redis call, but we can explore this at another moment.

As this PR is tackling new functionality requested by the frontend, probably rather than a `chore:` is a `feat:` :D

LGTM!

PR #334, opened 2025-02-10
Fix: files processing fixed conflict

SHAcollision:
Hey @amirRamirfatahi I saw there is a nasty merge conflict between `main` and `files-processing` . I tried solving it here. Still 2 tests failing, not sure if mayde due to this unsuccessful merge :smile: 

Merge into #165 if this is useful.

# Pre-submission Checklist


SHAcollision:
Hey @amirRamirfatahi , both tests are now fixed. We were 1. not loading the avatar from the right new static path and 2. not creating the ID as a hashId for the blob on the "test_delete" test, but as a timestampId. 

If we request a `Small` variant for an avatar using the shortcut `{user_id}/avatar` endpoint, which is the ideal usecase, the `small` variant will return a 404. Maybe we can find a way to improve this.

SHAcollision:
closing as the conflict was already fixed in #165 and the fixes were moved.

PR #335, opened 2025-02-11
chore: fix blob ids, remove pubky-common

SHAcollision:
As we compute IDs from `pubky-app-specs` and `pkarr` crate exports `Keypair` we can remove the dependency on `pubky-common` .

Blobs were being `put` malformed into the homeserver and with "wrong" id as it does not pass Nexus validation anymore unless ID is a correct hash id.

# Pre-submission Checklist


tipogi (approved):
Good!! one crate less 🚀

PR #336, opened 2025-02-12
feat: add in user/post counts index the unique tags field

tipogi:
### Chosen Decisions

1. Why `size([...])` (Pattern Comprehension) is more efficient than `OPTIONAL MATCH`?
- `OPTIONAL MATCH` performs multiple database scans. Each `OPTIONAL MATCH` executes a separate traversal for each relationship type (e.g., _TAGGED_, _REPLIED_, _REPOSTED_), leading to extra lookups and slower performance
- `size([...])` evaluates all patterns in a single scan. It counts relationships inline within a single query pass, reducing redundant graph traversals and making execution significantly faster and more efficient
- `size([...])` cannot count distinct values. If we need to count unique tag labels, we must use `OPTIONAL MATCH` and `COUNT(DISTINCT t.label)`

### Pre-submission Checklist


src/models/user/counts.rs
```diff
@@ -10,14 +11,18 @@ use super::UserStream;
 /// Represents total counts of relationships of a user.
 #[derive(Serialize, Deserialize, ToSchema, Debug, Default)]
 pub struct UserCounts {
+    // The number of tags assigned to other entities (e.g. user, posts)
     pub tags: u32,
+    // User received tags counts
+    pub tagged: u32,
+    // Distinct tags where the user was referenced
+    pub unique_tagged: u32,
```
SHAcollision, line 19: Yeah, now I get what you had in mind when you first asked about names!? I also think that for consistency across models (`PostCounts` and `UserCounts`) we could rename here these: `tagged` <-> `tags` (switch them here) and  `unique_tagged`  -> `unique_tags` .

This way the meaning will always be consistent across all of our models:
- `counts.tags` : tags received.
- `counts.unique_tags` : unique tags received.  
- `counts.tagged` will be only exist for `UserCounts` and stand for "tags the user has created" (for obvious reasons, posts don't create tags and don't need a `tagged` count).

SHAcollision (commented):
Maybe it could be valuable to add some test edge cases (e.g., the same tag added by the same user twice) to ensure the unique count logic is robust.

src/db/graph/queries/get.rs
```diff
@@ -35,24 +35,26 @@ pub fn get_post_by_id(author_id: &str, post_id: &str) -> Query {
...
+        MATCH (u:User {id: $author_id})-[:AUTHORED]->(p:Post {id: $post_id})
+        WITH p
+        OPTIONAL MATCH (p)<-[t:TAGGED]-()
+        WITH p, 
+            size([(p)<-[:TAGGED]-() | 1]) AS tags_count, 
+            COUNT(DISTINCT t.label) AS unique_tags_count, 
+            size([(p)<-[:REPLIED]-() | 1]) AS replies_count, 
+            size([(p)<-[:REPOSTED]-() | 1]) AS reposts_count, 
+            EXISTS { (p)-[:REPLIED]->(:Post) } AS is_reply
+        RETURN
```
amirRamirfatahi, line 48: So I had to have a deeper dive into this and the whole conversation around `size([...])` vs `OPTIONAL MATCH`.  
Thankfully I learned a lot as I hadn't dived deep into the inner workings of neo4j so far. 
 While using `size()` here is definitely speeding things up by removing redundant scans, the size() function materializes lists in memory, so it has a high memory usage. I believe a better version for this query would be something like this:
 ```
MATCH (u:User {id: $author_id})-[:AUTHORED]->(p:Post {id: $post_id})  
        WITH p  
OPTIONAL MATCH (p)<-[t:TAGGED]-()  
WITH p, COUNT { (p)<-[:TAGGED]-() } AS tags_count, COUNT(DISTINCT t.label) AS unique_tags_count  
RETURN p IS NOT NULL AS exists,  
       { tags: tags_count,  
         unique_tags: unique_tags_count,  
         replies: COUNT { (p)<-[:REPLIED]-() },  
         reposts: COUNT { (p)<-[:REPOSTED]-() }  
       } AS counts,  
       EXISTS { (p)-[:REPLIED]->(:Post) } AS is_reply
```

File: src/db/graph/queries/get.rs [hunk 3]
```diff
@@ -292,50 +294,39 @@ pub fn user_counts(user_id: &str) -> neo4rs::Query {
...
+
+        // Count relationships to posts
+        size([(u)-[:AUTHORED]->(:Post) | 1]) AS posts,
+        size([(u)-[:AUTHORED]->(:Post)-[:REPLIED]->(:Post) | 1]) AS replies,
+        size([(u)-[:BOOKMARKED]->(:Post) | 1]) AS bookmarks,
+
+        // Count user and post tagging
+        size([(u)-[:TAGGED]->(:User) | 1]) AS user_tags,
+        size([(u)-[:TAGGED]->(:Post) | 1]) AS post_tags,
+        size([(:User)-[:TAGGED]->(u) | 1]) AS tags
```
amirRamirfatahi, line 315: same argument left for post_counts applies here

amirRamirfatahi (commented):
I left a comment regarding the new queries.  
Moreover, while I know that we're processing new events sequentially, we should consider that this is a privilege lost in scale when we're forced to index new events in parallel. In that case, the solution here is not immune to race conditions. 
Using the sorted set for unique tags and getting their cardinality would not have this problem. While it might not be an immediate issue, we should aim to have these as little as possible. (Also maybe a good case to anchor a comment with like "Race condition prone" so in future when we move to parallel processing, we know where else needs fixing)

amirRamirfatahi (approved):

PR #345, opened 2025-02-20
chore: refactor global variables related with pubky

tipogi:
# Pre-submission Checklist


SHAcollision (approved):
This is OK to merge or not, as you prefer. I don't see this as much improvement. And I don't like we are loading from `.env` file outside of our `Config` struct.

PR #359, opened 2025-03-06
chore: nexus crates

tipogi:
# Phase 2

In this phase, we focused on reorganizing the repository to achieve a more modular design by creating a multi-crate workspace. Below are the key accomplishments and ongoing tasks:

- [x] Create new crate `pubky-common`: Contains shared logic and utilities
- [x] Create new crate `pubky-api`: Exposes and handles all API-related functionality
- [x] Create new crate `pubky-watcher`: Implements the file or data-watching mechanisms
- [x] Refactor the integration tests by placing them in their own crate for improved maintainability
- [x] Introduced a main daemon, `nexusd`, that coordinates the underlying crates to deliver a unified runtime experience
- [x] Update github workflow for the new design
- [x] Refactor common _functions_, _structs_, and _enums_ into appropriately scoped crates
- [x] Create a `benches` crate to run performance tests
- [ ] ~~Separate graph queries into a dedicated crate (currently part of `nexus-common`)~~
- [x] Rename migration structs to follow Rust naming conventions (e.g., UpperCase)
- [ ] ~~Implement safe shutdown using `tokio::signal::ctrl_c().await?`~~
- [x] Refine `pub use` statements to avoid importing unnecessary modules.
 

nexus-watcher/src/events/handlers/file.rs
```diff
@@ -115,8 +119,15 @@ pub async fn del(user_id: &PubkyId, file_id: String, files_path: PathBuf) -> Res
...
+                DbError::GraphQueryFailed { message } => {
+                    EventProcessorError::GraphQueryFailed { message }
+                }
+                DbError::IndexOperationFailed { message } => {
+                    EventProcessorError::IndexWriteFailed { message }
+                }
+            })?;
         }
 
         let folder_path = format!("{}/{}", user_id, file_id);
```
SHAcollision, line 133: Unrelated to this PR. NIT: We could consider using `PathBuf` and `.join` to unsure path separators are handled correctly on different operating systems and avoid mistakes.

nexus-common/src/media/mod.rs
```diff
@@ -101,27 +80,53 @@ impl StaticProcessor {
...
+            }
+            _ => file.content_type.clone(),
+        }
+    }
+
+    pub fn get_file_urls_by_content_type(content_type: &str, path: &str) -> FileUrls {
+        let variants = Self::get_valid_variants_for_content_type(content_type);
+
+        FileUrls {
+            main: format!("{}/{}", path, FileVariant::Main),
```
SHAcollision, line 99: Nit and unrelated to this PR: Here too we could use `PathBuf` maybe? Also in `watcher/src/events/handlers/file.rs` ? Overall we could have a second look at where we are doing string interpolation to create a path and fix it.

SHAcollision (approved):
Congrats @tipogi ! This refactor is a significant move toward a clean, modular architecture by separating core concerns into distinct crates and moving common code into a shared module. Looks good to merge, we could merge to `main` as well.

I did not take a look in detail into changes as "everything has changed", I only have seen that:
- The new structure makes sense and is super solid :heavy_check_mark: 
- Tests still pass! :heavy_check_mark: 

Only thing I think we should do before merging:
- [x] We could consider establishing the versions of ALL crates as `0.3.2` so there is continuity after `0.3.1` (the one before the refactor), then proceed to make a new release and publish them all into `crates.io`. Why? Well, current Nexus API is `0.3.1` (this data is displayed in `/v0/info`), it is a bit awkward that Nexus API now displays version `0.1.0`.

I leave some suggestions for improvement that we could do in the future:
- Dependabot: I guess we need to edit `dependabot.yml` so it finds the new crates' `Cargo.toml` but maybe it works "as is".
- Improve configuration management using an established configuration crate like `config` or `figment`, this could help a lot at ingesting configurations without much boilerplate from `.toml`, environment, and CLI args. 
- Documentation: we could make each crate’s public API well documented. For example, adding a module-level documentation (using //!) at the top of each `lib.rs` file to explain the crate’s responsibilities. 
- We could dive deeper later on and examine case by case what should be `pub` and what should not be externally accessible for each of these crates. Let's say "clean up and prune external API). Same for `pubky-app-specs`, for most models we are accessing their data with `pub` fields but ideally we write getters for them so we can start building a stable API for them as well.
- I don't know what you think about this, maybe centralize error handling by defining a unified error type, maybe on `nexus-common` or maybe on each crate `watcher` / `api`:
Just an example:
```rust
// nexus-common/src/error.rs

use thiserror::Error;
use std::fmt;

/// A unified error type used throughout the application.
#[derive(Error, Debug)]
pub enum NexusError {
    #[error("Database error: {0}")]
    Database(#[from] DbError),

    #[error("Network error: {0}")]
    Network(#[from] reqwest::Error),

    #[error("Configuration error: {0}")]
    Config(String),

    #[error("General error: {0}")]
    General(String),

    #[error("Other error: {0}")]
    Other(#[from] Box<dyn std::error::Error + Send + Sync>),
}
```

Here take this  :crab:  with you :rocket:

PR #381, opened 2025-03-21
bump: pubky-core

tipogi:
# Pre-submission Checklist


SHAcollision (commented):
Nice!

I believe we are missing bumping `pubky` to 0.4.2 in `nexus-common` and we could also remove `pubky` as a dependency of `nexus-api` right?

PR #391, opened 2025-03-25
fix(models): allow stream of post in ascending order

SHAcollision:
Sorting post replies in ascending order helps in a great way to fix https://github.com/pubky/pubky-app/issues/1241
as @flaviomoceri requested, `order=ascending|descending` can not be requested by the client.

# Pre-submission Checklist

### DIFF
--- a/nexus-api/src/routes/v0/stream/posts.rs
+++ b/nexus-api/src/routes/v0/stream/posts.rs
@@ -106,8 +106,8 @@ pub async fn stream_posts_handler(
     }
 
     let source = query.source.unwrap_or_default(); // StreamSource::All is default
-    let sorting = query.sorting.unwrap_or_default(); // StreamSorting::Timeline) is default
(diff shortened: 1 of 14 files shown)


flaviomoceri (approved):

nexus-api/src/routes/v0/stream/posts.rs
```diff
@@ -104,10 +107,12 @@ pub async fn stream_posts_handler(
 
     let source = query.source.unwrap_or_default(); // StreamSource::All is default
     let sorting = query.sorting.unwrap_or_default(); // StreamSorting::Timeline) is default
+    let order = query.order.unwrap_or_default(); // StreamSorting::Timeline) is default
```
tipogi, line 110: SortOrder::Descending

SHAcollision: `Descending` is `#[Default]` , do you think it's better to be more explicit here?

PR #393, opened 2025-03-25
feat(api): add response compression with tower‑http

SHAcollision:
# Pre-submission Checklist

- Integrates CompressionLayer from tower‑http into the Axum stack
- Automatically compresses all responses to reduce payload size and improve performance
- Remove dev dependency `reqwest` that was not really needed as we can use `axum` types.

Request responses now transfer anywhere between 60%-90% more data. A `/stream/post` with 30 members transfer only `1.7Kb` vs `15.9Kb` before this PR on a modern browser using `zstd` . It also support `gzip`, `brotli` and `deflate` depending on browser's support.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -37,6 +37,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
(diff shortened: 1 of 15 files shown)


tipogi (approved):
nicee improvement 🚀

PR #394, opened 2025-03-26
204 response when filters do not match and index exists

tipogi:
# Pre-submission Checklist

### DIFF
--- a/nexus-api/src/routes/v0/user/tags.rs
+++ b/nexus-api/src/routes/v0/user/tags.rs
@@ -26,6 +26,7 @@ use utoipa::OpenApi;
     ),
     responses(
         (status = 200, description = "User tags", body = TagDetails),
+        (status = 204, description = "Tags not found"),
         (status = 404, description = "User not found"),
         (status = 500, description = "Internal server error")
     )
@@ -50,7 +51,15 @@ pub async fn user_tags_handler(
     )
     .await
     {
-        Ok(Some(tags)) => Ok(Json(tags)),
+        Ok(Some(tags)) => {
+            if tags.len() == 0 {
+                Err(Error::EmptyStream {
+                    message: "No tags found for the given criteria.".to_string(),
+                })
+            } else {
+                Ok(Json(tags))
+            }
+        },
         Ok(None) => Err(Error::UserNotFound { user_id }),
         Err(source) => Err(Error::InternalServerError { source }),
     }
--- a/nexus-common/src/models/tag/traits/collection.rs
+++ b/nexus-common/src/models/tag/traits/collection.rs
@@ -90,6 +90,7 @@ where
         {
             Some(tag_details) => Ok(Some(tag_details)),
             None => {
+                println!("Get from graph");
                 let graph_response = Self::get_from_graph(user_id, extra_param, None).await?;
                 if let Some(tag_details) = graph_response {
                     Self::put_to_index(user_id, extra_param, &tag_details, false).await?;
@@ -133,6 +134,7 @@ where
             ),
             false => (None, None),
         };
+        println!("skip_tags: {:?}, limit_tags: {:?}, limit_taggers: {:?}", skip_tags, limit_tags, limit_taggers);
         // Get related tags
         match Self::try_from_index_sorted_set(
             &key_parts,
@@ -146,6 +148,7 @@ where
         .await?
         {
             Some(tag_scores) => {
+                println!("{:?}", tag_scores);
                 let mut tags = Vec::with_capacity(limit_tags);
                 // TODO: Temporal fix. Should it delete SORTED SET value if score is 0?
                 for (label, score) in tag_scores.iter() {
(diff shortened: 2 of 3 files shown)


nexus-api/src/routes/v0/user/tags.rs
```diff
@@ -50,7 +51,15 @@ pub async fn user_tags_handler(
     )
     .await
     {
-        Ok(Some(tags)) => Ok(Json(tags)),
+        Ok(Some(tags)) => {
+            if tags.len() == 0 {
+                Err(Error::EmptyStream {
+                    message: "No tags found for the given criteria.".to_string(),
+                })
```
SHAcollision, line 58: `"No tags provided"` ??

SHAcollision (approved):
Did not test, but if it works as expected for you, let's go! Maybe one test case to cover the correctness of this fix could be added if anything.

nexus-api/src/routes/v0/utils.rs
```diff
@@ -0,0 +1,19 @@
...
+use axum::Json;
+
+/// It is used in Axum route handlers to streamline the logic for
+/// returning a JSON-wrapped response from a list of items, while gracefully handling the case
+/// where the list is empty
+/// # Arguments
+///
+/// * `items` - The vector of items to be returned in the JSON response
+/// * `model` - A string used in the error message to indicate which kind of item was expected
+pub fn as_json_or_error<T>(items: Vec<T>, model: &str) -> Result<Json<Vec<T>>> {
```
SHAcollision, line 11: I am not sure if I understand this weel. It seems on some endpoints (like notifications, search/tags, and user/tags), empty results are now treated as an error via `as_json_or_error` ? whereas in other parts of the code (for example, in some internal functions in the post stream), an empty vector is returned directly. It might be clearer to decide on one consistent approach for empty responses across the API: either always returning an empty array (with a 200/204 status) or returning an error that maps to a specific status code. I guess we always had this inconsistency.

nexus-api/src/routes/v0/utils.rs
```diff
@@ -0,0 +1,19 @@
...
+    if items.is_empty() {
+        Err(Error::EmptyStream {
+            message: format!("No {:?} found for the given criteria", model),
```
SHAcollision, line 14: `model` is a `&str` so `{:?}` (debug display) could be just `{}` to avoid unnecessary quotes.

nexus-api/src/routes/v0/tag/global.rs
```diff
@@ -130,7 +130,9 @@ pub async fn hot_tags_handler(Query(query): Query<HotTagsQuery>) -> Result<Json<
 
     match HotTags::get_hot_tags(query.user_id, query.reach, &input).await {
         Ok(Some(hot_tags)) => Ok(Json(hot_tags)),
```
SHAcollision, line 132: Do we need `as_json_or_error()` here?

nexus-common/src/models/tag/traits/collection.rs
```diff
@@ -90,6 +90,7 @@ where
         {
             Some(tag_details) => Ok(Some(tag_details)),
             None => {
+                println!("Get from graph");
```
SHAcollision, line 93: Leftover `println!`

SHAcollision (commented):
Nice work @tipogi ! :rocket: The changes improve caching logic, reduce unnecessary queries (e.g., for tags when counts are zero), and add valuable documentation, the main area for improvement is the consistency in handling “empty” results across different layers of the application. 

We should decide on a uniform pattern for empty responses and update both the API and internal function signatures accordingly.

nexus-api/src/routes/v0/utils.rs
SHAcollision: HTTP Semantics: In REST, an empty list is often acceptable. Many APIs simply return a 200 with an empty array or use a 204 “No Content” status without a body.

While this function avoids returning empty lists, the decision to “error out” might force clients to handle this as an exceptional condition rather than a normal state. Is an empty result really an error? I know we have debated this before, but it's kind of weird tbh. We should consider if it would be better to return a successful response with an empty JSON array.

The function name `as_json_or_error` is a bit vague. A more descriptive name (for example, `json_response_or_empty_error` ) might help.

The generic parameter T isn’t bounded by a trait (like Serialize), which might cause issues if used with types that aren’t serializable. Adding a constraint (T: Serialize) would be safer.

SHAcollision:
> The updated implementation now distinguishes between:
> 
> * The index not existing — triggers a graph lookup to populate the cache
> * The index existing but exhausted due to filters — skips the graph lookup and simply returns an empty result

Could the sorted set exist and be empty? If the sorted set is only created when it has at least 1 element we will always be querying the graph on every request for a post that has no tags. Do we need graph fallback for this at all?

nexus-api/src/routes/v0/utils.rs
tipogi: Good catch! The `fn xx_error` name was misleading, since the response isn't treated as an actual error — it just returns an empty array case as `204, No Content`. I believe we had already defined this behavior as our standard across the API, but it looks like some endpoints hadn't adopted it yet. Fixed that now!

PR #411, opened 2025-04-01
chore: fix influencers query

tipogi:
# Pre-submission Checklist

### DIFF
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -515,21 +515,23 @@ pub fn get_influencers_by_reach(
         format!(
             "
         {}
-        WHERE user.id = $user_id
+        WHERE user.id = $user_id  
+        WITH DISTINCT reach
 
-        OPTIONAL MATCH (others:User)-[follow:FOLLOWS]->(reach)
-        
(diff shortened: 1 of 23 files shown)


nexus-common/src/db/graph/queries/get.rs
```diff
@@ -515,21 +515,23 @@ pub fn get_influencers_by_reach(
...
+        CALL (reach) {{
+            MATCH (reach)-[tag:TAGGED]->(:Post)
+            RETURN count(tag) as tags_count
+        }}
+        CALL (reach) {{
+            MATCH (reach)-[:AUTHORED]->(post:Post)
+            RETURN count(post) as posts_count
+        }}
 
+        WITH reach, followers_count, tags_count, posts_count
```
SHAcollision, line 534: Looks like behaviour changed and the test for "following" reach influencers is failing.

What is the impact of this change in DB hits? How long was it taking before vs now for computing that big query the frontend was requesting?

SHAcollision: Answering myself, tested the old vs new query on Prod DB with my user id at latency went from `9800 ms` to `260 ms`

SHAcollision: We should anyway make sure that the Global queries (reach All), even if there is a `user_id`  get served from Redis

PR #421, opened 2025-04-16
fix(common): redis with deadpool

SHAcollision:
Same as #420 but using `deadpool-redis` instead for hopefully better scalability.

# Pre-submission Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -960,10 +960,34 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


SHAcollision:
Merging `deadpool` on favor of #420 as a single multuplexed connection saturates at ~40 req/sec while pooling scales better. Over ~190 req/sec  (maxing CPU usage) on the `nexus API` machine.

PR #426, opened 2025-04-24
feat: fetch posts in batch

tipogi:
# Pre-submission Checklist

### DIFF
--- a/nexus-api/src/routes/v0/stream/posts.rs
+++ b/nexus-api/src/routes/v0/stream/posts.rs
@@ -158,7 +158,7 @@ pub async fn stream_posts_by_ids_handler(
         request.post_ids.len()
     );
 
-    const MAX_POSTS: usize = 1000;
+    const MAX_POSTS: usize = 100;
 
     if request.post_ids.len() > MAX_POSTS {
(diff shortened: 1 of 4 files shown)


nexus-api/src/routes/v0/stream/posts.rs
```diff
@@ -128,9 +128,65 @@ pub async fn stream_posts_handler(
...
+pub async fn stream_posts_by_ids_handler(
+    Json(request): Json<PostStreamByIdsRequest>,
+) -> AppResult<Json<PostStream>> {
+    info!(
+        "POST {} post_ids size {:?}",
+        STREAM_POSTS_BY_IDS_ROUTE,
+        request.post_ids.len()
+    );
+
+    const MAX_POSTS: usize = 1000;
```
SHAcollision, line 161: It's a very big value maybe? What about `100` or so? The frontend should be aware of the maximum to avoid sending requests that will get `Error::InvalidInput`. Another option is that Nexus sanitises the list and crops it to 100 always without erroring.

SHAcollision (approved):
Looks good to go!! :rocket:

SHAcollision (approved):

PR #430, opened 2025-04-29
chore(common/api): replace redis crate for deadpool_redis::redis

SHAcollision:
# Pre-submission Checklist


nexus-api/Cargo.toml
```diff
@@ -20,7 +20,7 @@ opentelemetry = { workspace = true }
 opentelemetry_sdk = { workspace = true }
 pubky-app-specs = { workspace = true }
 nexus-common = { version = "0.4.0", path = "../nexus-common" }
-redis = { version = "0.29.5", features = ["tokio-comp", "json"] }
+deadpool-redis = "0.20.0"
```
tipogi, line 23: what about the `json` feature, don't we need to import? I can see by default it imports tokio runtime

SHAcollision: Good catch! It looks like `deadpool-redis` does not use the `json` feature: https://github.com/deadpool-rs/deadpool/blob/7623410e20c54ff9139e618cf18ce70e532d260e/redis/Cargo.toml#L56 so if we want to use it, then we need to keep the `redis` crate always on a compatible version (0.29). However... it looks like maybe we did not need the `"json"` feature? All tests are passing with this change :thinking:

SHAcollision: It all seems to be working in Staging so we are going to recklessly merge!

PR #433, opened 2025-05-05
chore: unify the configuration file to the daemon process

tipogi:
# Pre-submission Checklist


examples/README.md
```diff
@@ -0,0 +1,28 @@
+# Examples
+
+Reference implementations showing primary crates functionality:
```
SHAcollision, line 3:
```suggestion
Example usage of crates `nexus-api` and `nexus-watcher` :
```

nexus-api/src/main.rs
SHAcollision, line 1: Good, bye bye!

nexus-watcher/src/main.rs
SHAcollision, line 1: bye!

nexusd/src/migrations/builder.rs
```diff
@@ -1,36 +1,35 @@
...
+use nexus_common::file::ConfigLoader;
+use nexus_common::file::ConfigReader;
+use nexus_common::StackConfig;
 use nexus_common::StackManager;
-use nexus_common::{Config as StackConfig, ConfigLoader};
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use std::fmt::Debug;
 
-pub const CONFIG_FILE: &str = "./src/migrations/mconf_template.toml";
+pub const CONFIG_FILE: &str = ".pubky-nexus/migrations";
```
SHAcollision, line 10: Where is this?

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,95 @@
+use crate::types::DynError;
+use async_trait::async_trait;
+use serde::de::DeserializeOwned;
+use std::fmt::Debug;
+use std::path::{Path, PathBuf};
+use tracing::error;
+
+use super::ConfigLoader;
+
+pub const DEFAULT_HOME_DIR: &str = ".pubky-nexus";
```
SHAcollision, line 10: Is this the same as `~/.pubky-nexus` ?

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,95 @@
...
+const TEMPLATE_CONFIG_FILE: &str = "nexusd/conf_template.toml";
```
SHAcollision, line 11: It's hard to get what's going on here. These files are being READ at runtime, right? They might not be there (actually, they won't!). This only works with `cargo run` when executed on the project folder.

Instead, we want the whole files to be included at compile time, right? `const DEFAULT_CONFIG_TOML: &str = include!("nexusd/conf_template.toml"). In any case when someone simply pulls the `nexus-common` crate the `nexusd/conf_template.toml` path might not exist, this needs testing, but it's safer if the `.toml` is in `nexus-common` crate.

Is `conf_teample.toml` a sample file, or a defaults file? Let's be specific by naming it `sample.config.toml` or `default.config.toml` . If it is copied on initialization into the `~/.pubky-nexus` by `write_default_config_file` then it becomes the default config on a restart. Does it match also the default config when no config is provided?

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,95 @@
...
+        };
+        // Copy the file
+        std::fs::copy(PathBuf::from(file_path), config_file_path)?;
+        Ok(())
+    }
+
+    /// Reads the config file from the data directory
+    /// Creates a default config file if it doesn't exist
+    async fn read_config_file(expanded_path: PathBuf, migration_file: bool) -> Result<T, DynError> {
+        println!("nexusd reading the conf.toml file from {:?}", expanded_path);
```
SHAcollision, line 72: It will print this even if the file does not exist yet.

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    if path.starts_with("~/") {
+        if let Some(home) = dirs::home_dir() {
+            let without_home = path.strip_prefix("~/").expect("Invalid ~ prefix");
+            let joined = home.join(without_home);
+            return joined;
+        }
+    }
+    PathBuf::from(path)
+}
```
SHAcollision, line 37:
```suggestion
pub fn expand_home_dir(mut path: PathBuf) -> PathBuf {
    if let Some(first) = path.components().next() {
        if first == Component::Normal(OsStr::new("~")) {
            if let Some(home) = dirs::home_dir() {
                // drop the "~" prefix and re-join
                let without_tilde = path.iter().skip(1).collect::<PathBuf>();
                return home.join(without_tilde);
            }
        }
    }
    path
}
```

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,106 @@
...
+        expanded_path.join(CONFIG_FILE_NAME)
+    }
+
+    fn write_default_config_file(
+        config_file_path: &PathBuf,
+        is_migration_file: bool,
+    ) -> std::io::Result<()> {
+        // Make sure before write the file, the directory path exists
+        if let Some(parent) = config_file_path.parent() {
+            println!(
```
SHAcollision, line 55: let's use `info!` instead of `println!` to stick with the use of `tracing` crate.

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,106 @@
...
+    async fn read_config_file(
+        expanded_path: PathBuf,
+        is_migration_file: bool,
+    ) -> Result<T, DynError> {
+        let config_file_path = Self::get_config_file_path(&expanded_path);
+
+        if !config_file_path.exists() {
+            Self::write_default_config_file(&config_file_path, is_migration_file)?;
+        }
+        println!(
```
SHAcollision, line 83: `info!`

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,106 @@
...
+            })?;
+        Ok(config)
+    }
+}
+
+// ——————————————————————————————————————————————————————————————
+// Blanket impl so *any* `T` automatically gets a `ConfigReader<T>`
+// ——————————————————————————————————————————————————————————————
+#[async_trait]
+impl<T> ConfigReader<T> for T where T: ConfigLoader<T> + DeserializeOwned + Send + Sync + Debug {}
```
SHAcollision, line 106: This is an AI recommendation, not sure if this makes sense to you:

Right now you have:

```rust
pub trait ConfigReader<T> : ConfigLoader<T>
where
    T: DeserializeOwned + Send + Sync + Debug,
{ … }

#[async_trait]
impl<T> ConfigReader<T> for T 
where T: ConfigLoader<T> + DeserializeOwned + Send + Sync + Debug {}
```
You can collapse this by making ConfigReader non-generic:

```rust
#[async_trait]
pub trait ConfigReader: ConfigLoader + Sized
where
    Self::Output: DeserializeOwned + Send + Sync + Debug,
{
    /// Creates or reads the config at `base_dir`.
    async fn read_config(self, base_dir: PathBuf, is_migration: bool) 
        -> Result<Self::Output, DynError> { … }
}
```

Then implement once:

```rust
#[async_trait]
impl<L> ConfigReader for L
where
    L: ConfigLoader,
    L::Output: DeserializeOwned + Send + Sync + Debug,
{
    // ...your read_config implementation...
}
```
This reduces the double-use of T and makes the plumbing clearer.

nexus-watcher/src/events/processor.rs
```diff
@@ -36,6 +35,10 @@ impl EventProcessor {
     pub async fn test(homeserver_id: String) -> Self {
         let id = PubkyId::try_from(&homeserver_id).expect("Homeserver ID should be valid");
         let homeserver = Homeserver::new(id).await.unwrap();
+        info!(
+            "Watcher static files PATH is inside their own crate: {}",
```
SHAcollision, line 39: I don't think this can be understood.

Maybe we can change the message to "Watcher static files PATH during tests are stored inside of the watcher crate: {}", and then use the full path by using `PathBuf::from(FILES_DIR_TEST)` ?

examples/api/readme.md
SHAcollision, line 1: Empty file. Can be removed.

examples/README.md
```diff
@@ -0,0 +1,28 @@
...
+cargo run --bin watcher_example
+```
+
+## ⚙️ Configuration & Flags
+
+Both `api` and `watcher` binaries include a compile-time switch in `main.rs`:
+
+```rust
+// Toggle between embedded/default config vs. file-based config
+const FROM_FILE: bool = false;
```
SHAcollision, line 25: Let's avoid baking in `FROM_FILE` flags in source. Hard-coding const `FROM_FILE: bool = …` means every time someone wants to switch modes they must recompile. Instead we could read `--config` CLI flag.

Examples should “just work”. Point file paths in the example configs at `./examples/api/config.toml` etc., and in the example code do:
```
cargo run --example api_example -- --config examples/api
```
That way users don’t have to rename or copy files around.

examples/api/config.toml
```diff
@@ -1,11 +1,11 @@
 # Service name used for tracing, logging, and metrics in OpenTelemetry
```
SHAcollision, line 1: To make extra sure of `config.toml` are not confused we could rename this one to `api.config.toml` the other example/ one to `watcher.config.toml`

nexusd/src/cli.rs
```diff
@@ -1,34 +1,42 @@
...
+    dirs::home_dir().unwrap_or_default().join(DEFAULT_HOME_DIR)
+}
+
+/// Validate that the data_dir path is a directory.
+/// It doesnt need to exist, but if it does, it needs to be a directory.
+fn validate_config_dir_path(path: &str) -> Result<PathBuf, String> {
+    let path = expand_home_dir(PathBuf::from(path));
+    if path.exists() && path.is_file() {
+        return Err(format!("Given path is not a directory: {}", path.display()));
+    }
```
SHAcollision, line 36: We could suggest `“create with mkdir -p …” or “point to a directory”`

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::path::{Path, PathBuf};
+use tracing::error;
+
+use super::ConfigLoader;
+
+/// Path to default nexusd config file. Defaults to ~/.pubky-nexus
+pub const DEFAULT_HOME_DIR: &str = ".pubky-nexus";
+const TEMPLATE_CONFIG_FILE: &str = include_str!("../default.config.toml");
+const MIGRATION_TEMPLATE_CONFIG_FILE: &str =
+    include_str!("../../../../nexusd/src/migrations/default.config.toml");
```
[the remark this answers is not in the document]
tipogi: The pubky-common crate cannot create or run migrations—only the nexusd crate is responsible for executing migration tasks. This path lives in pubky-common solely because it’s used by its file-reading utilities

nexus-common/src/config/file/reader.rs
```diff
@@ -0,0 +1,106 @@
...
+use serde::de::DeserializeOwned;
+use std::fmt::Debug;
+use std::path::{Path, PathBuf};
+use tracing::error;
+
+use super::ConfigLoader;
+
+/// Path to default nexusd config file. Defaults to ~/.pubky-nexus
+pub const DEFAULT_HOME_DIR: &str = ".pubky-nexus";
+const TEMPLATE_CONFIG_FILE: &str = include_str!("../default.config.toml");
```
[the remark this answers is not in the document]
tipogi: I don’t see an issue. I placed it there because that folder contains everything related to configuration tasks

nexus-common/src/config/file/reader.rs
tipogi: The tracing won’t work in this case, because we have to read first the configuration file to determine the log verbosity level (info, error, etc.) before activating the tracing engine. Once we get that param, initializing the tracing engine with those settings is the responsibility of `init_stack`.

nexus-common/src/config/file/reader.rs
tipogi: same as above comment

examples/api/readme.md
tipogi: fixed in the child [PR](https://github.com/pubky/pubky-nexus/pull/440/files)

examples/README.md
tipogi: makes sense 👍

examples/api/config.toml
tipogi: I named it this way because the user only loads files called `config.toml`. If a file has any other name, it won’t be read—so users must adhere to this naming convention

nexus-common/src/config/file/reader.rs
SHAcollision: Ah, that makes sense!

examples/api/config.toml
SHAcollision: Ummm... Okay. We can leave it like that.

SHAcollision (approved):
Let's go!! :rocket:

PR #439, opened 2025-05-08
bug: post details attachments, wrong type

tipogi:
## Breaking change
- **Consistent Array Semantics**: After this change, `PostDetails.attachments` will always be an array—empty (`[]`) when there are no attachments, or populated with one or more entries when they exist. There will be no more ambiguity between `null` and `[]`.

- **Coordinated Deployment**: This PR must go live in lockstep with the corresponding update in `pubky.app`; otherwise, the frontend may encounter mismatched attachment formats and display inconsistent UI states

CC: @catch-21 @flaviomoceri @MiguelMedeiros 

### Pre-submission Checklist

### DIFF
--- a/nexus-common/src/models/post/details.rs
+++ b/nexus-common/src/models/post/details.rs
@@ -46,7 +46,16 @@ impl PostDetails {
         author_id: &str,
         post_id: &str,
     ) -> Result<Option<PostDetails>, DynError> {
-        if let Some(post_details) = Self::try_from_index_json(&[author_id, post_id], None).await? {
+        if let Some(mut post_details) =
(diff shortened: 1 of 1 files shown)


SHAcollision (approved):
ACK! 

It's an OK temporary solution so our API always returns empty vectors consistently. But we have to come back here and change PostDetails attachments to be a `Vec<String>` and perform a DB migration.

tipogi:
opened an issue, describing the solution, #441

flaviomoceri:
@catch-21 @MiguelMedeiros  after the merge of it please check here https://github.com/pubky/pubky-app/pull/1495

SHAcollision:
@tipogi should we close here?

Issue #446, opened 2025-05-16
feat(api): search user by pubky prefix match


SHAcollision:
In the spirit of #72, it will also be useful to have an endpoint to return full pubkey matches for prefixes entered into the search bar,

In order to complete this feature we need:
1. A new sorted set for lexicographical matches similar to the existing `Sorted:Users:Name` sorted set.
2. New logic to populate the sorted set index by the `nexus-watcher` when new `profile.json` are found (same as we do for the `Sorted:Users:Name`
3. A new "user search" endpoint and route (or combine with current search by username?)
4. New `nexus-api/tests` for the new endpoint.
5. New `nexus-api/benchmarks` for the new endpoint.

*Ideas*

This needs more research:
The `Sorted:Users:Name` does prefix matches on strings like `john carvalho:gujx6qd8ksydh1makdphd3bxu351d9b8waqka8hfg6q7hnqkxexo` for an input like `joh` then it parses the string and returns the pubky_ids behind the `:` separator. This just works, although we always return the results ordered alphabetically. The new prefix match is even more straight forward because we will be matching on the pubky_id itself. That means we could use the separator technique to store the "inverse_relevance" for that user `{pubky_id}:{inverse_relevance}` so the results are also naturally returned by relevance (because alphabetically, a lower number will come first). I'm not 100% sure this would work just as I imagine, so it needs some research :)

PR #450, opened 2025-05-19
Add tag autocompletion endpoint

### LINKED ISSUE #402: feat(api/watcher): add new endpoint for tag autocompletion
Same as we have a `user search` endpoint that helps with user autocompletion, we could have a `tag search` by prefix endpoint that can be used for:
- Suggest tags when searching for tags
- Suggest tags when writing a new tag. This way it is less likely new tags are created with mispelling or with a `-` instead of a `_` separator of a popular tag.

ok300:
This PR adds a `nexus-api` endpoint that can be used for tag auto-completion.

Commits:

* b9d88c315dad4009ab8c4ac309d5a82beafe385f : Rename `/search/tags` to `/search/posts/by_tag`
  * `/search/posts/` aligns this method with the existing `/search/users/` (first returns posts, second returns users)
  * the `/by_tag` extra path element aligns with some of the other APIs that contain `/by_ids`
* a4896891e42cdbb8d357847d7645cc21c0947666 : Add `/search/tags/by_prefix` endpoint
  * Adds a new Cypher query, model getter and API route that returns a list of tag labels by label prefix

Open TODOs:

* [x] Add `/search/tags/by_prefix` tests
* [x] Add `/search/tags/by_prefix` benchmark

Fixes #402


nexus-api/src/routes/v0/endpoints.rs
```diff
@@ -47,7 +47,11 @@ pub const STREAM_TAGS_REACH_ROUTE: &str =
 // -- SEARCH endpoints --
 const SEARCH_PREFIX: &str = concatcp!(VERSION_ROUTE, "/search");
 pub const SEARCH_USERS_ROUTE: &str = concatcp!(SEARCH_PREFIX, "/users");
-pub const SEARCH_TAGS_ROUTE: &str = concatcp!(SEARCH_PREFIX, "/tags/{label}");
+pub const SEARCH_POSTS_BY_TAG_ROUTE_BASE: &str = concatcp!(SEARCH_PREFIX, "/posts/by_tag");
+pub const SEARCH_POSTS_BY_TAG_ROUTE: &str = concatcp!(SEARCH_POSTS_BY_TAG_ROUTE_BASE, "/{tag}");
+pub const SEARCH_TAGS_BY_PREFIX_ROUTE_BASE: &str = concatcp!(SEARCH_PREFIX, "/tags/by_prefix");
+pub const SEARCH_TAGS_BY_PREFIX_ROUTE: &str =
+    concatcp!(SEARCH_TAGS_BY_PREFIX_ROUTE_BASE, "/{prefix}");
```
SHAcollision, line 54: If we don't expect more than one `"/posts/by_tag/*"` and more than one `"/tags/by_prefix/*"`, like in this case, we could collapse these into just:
```suggestion
pub const SEARCH_POSTS_BY_TAG_ROUTE: &str = concatcp!(SEARCH_PREFIX, "/posts/by_tag/{tag}");
pub const SEARCH_TAGS_BY_PREFIX_ROUTE: &str = concatcp!(SEARCH_PREFIX, "/tags/by_prefix/{prefix}");
```
This avoid unnecessary verbosity. I see other `const`s like `STREAM_POSTS_ROUTE` and `STREAM_POSTS_BY_IDS_ROUTE` also seem to have this pitfall, would also remove the unnecessary redundancy there as well.

nexus-api/src/routes/v0/search/mod.rs
```diff
@@ -21,7 +23,8 @@ pub struct SearchApiDoc;
 impl SearchApiDoc {
     pub fn merge_docs() -> utoipa::openapi::OpenApi {
         let mut combined = users::SearchUsersApiDocs::openapi();
-        combined.merge(tags::SearchTagPostsApiDocs::openapi());
+        combined.merge(posts::SearchPostsByTagApiDocs::openapi());
+        combined.merge(tags::SearchTagsByPrefixApiDocs::openapi());
```
SHAcollision, line 27: Nice, it's complete, exposes the new API on swagger-ui correctly :ok_hand:

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -386,6 +386,17 @@ fn stream_reach_to_graph_subquery(reach: &StreamReach) -> String {
     }
 }
 
+pub fn get_tags_by_label_prefix(label_prefix: &str) -> Query {
+    query(
+        "
+        MATCH ()-[t:TAGGED]->()
+        WHERE t.label STARTS WITH $label_prefix
```
SHAcollision, line 393: This might be pretty expensive operation. Hopefully it never needs to be executed, but when it does (re-indexing, or so) It might benefit from a new index in Neo4J. 

Worth benchmarking with and without this DB index:
```
// Insert in nexus-common/src/db/graph/setup.rs indexes[]
"CREATE INDEX taggedLabelIndex IF NOT EXISTS FOR ()-[r:TAGGED]-() ON (r.label)",
```

nexus-common/src/models/tag/global.rs
```diff
@@ -174,6 +174,17 @@ impl Taggers {
...
     }
 
+    pub async fn get_tags_by_label_prefix(
+        label_prefix: &str,
+        // user_id: &str,
+        // reach: StreamReach,
+        // skip: usize,
+        // limit: usize,
+    ) -> Result<Option<Vec<String>>, DynError> {
+        let query = queries::get::get_tags_by_label_prefix(label_prefix);
```
SHAcollision, line 184: This works, but it will be inefficient and will likely not scale as the graph grows.

Recommending tags should happen always from Redis cache. I suggest we use the same structure we use on the `search/user.rs` endpoint and `models/user/search.rs` model. We have `UserSearch::get_by_name()` where we use a Redis Sorted Set where every `key, value` are `username, 0` . If a sorted set in redis has all values of `0` redis allows very optimised lexicographical prefix matching ( that by implemented the `RedisOps` trait will be available as `Self::try_from_index_sorted_set_lex()` ).

Like in most other models, only if the Redis index is missing, we could fallback to the graph query. However, as you see in `UserSearch`, we don't even fallback to graph: the sorted set existence is taken for granted and falling back to graph for a query that will take place so often ("J", "Jh", "Jho", "Jhon" in short seconds) is not worth it (too expensive).

This is what makes feat #402 harder than #446 and why I think starting with #446 is a good idea as. By copying the "search by username prefix" structure to make a "search by pubky_id key prefix you see the full pipe to create this Redis set, populate it as new information gets indexed by the `nexus-watcher` and then use it on the API.

SHAcollision (commented):
Hey @ok300 the PR is looking good! You did not even miss any detail on the route decorator for swagger-ui :ok_hand: . 

I this endpoint will benefit A LOT for copying what we implemented in `UserSearch` model `/models/user/search.rs` for efficient serving from cache and never hit the graph DB.

nexus-common/src/models/tag/search.rs
```diff
@@ -1,163 +1,56 @@
...
-    async fn add_to_global_sorted_set(query: Query, index_key: [&str; 4]) -> Result<(), DynError> {
-        let mut result;
-        {
-            let graph = get_neo4j_graph()?;
-
-            let graph = graph.lock().await;
-            result = graph.execute(query).await?;
+        let tag_labels: Option<Vec<String>> = retrieve_from_graph(get_tags(), "tag_labels").await?;
+        for tag_label in tag_labels.unwrap_or_default() {
+            Self::put_index_sorted_set(&TAGS_LABEL, &[(0.0, &tag_label)], None, None).await?;
```
tipogi, line 22: What about, if we use directly `Self::put_to_index` directly. Like this when ever we change the internal `put_to_index` that function would not be affected

nexus-common/src/models/tag/search.rs
```diff
@@ -1,163 +1,56 @@
...
-                    None,
-                    None,
-                )
-                .await?;
-            }
-        }
-        Ok(())
+    pub async fn put_to_index(tag_label: &str) -> Result<(), DynError> {
+        let elements = [(0.0, tag_label)];
+        Self::put_index_sorted_set(&TAGS_LABEL, &elements, Some("Sorted"), None).await
```
tipogi, line 49: the third argument in that case, it is not need in the `SortedSet` because the prefix argument is a value that is not common in that data structure index. For example, one of the cases would be RetryManager [index](https://github.com/pubky/pubky-nexus/blob/e73a762b532f5aa9aa2fa58f9da6c9b318e9a06d/nexus-watcher/src/events/retry/event.rs#L66). We do not want that it will start as a `Sorted`, we define like `RetryManager::Events`. All the Sorted Set data structures, start as `Sorted` thats why in that case, there is not need to add that param. Look [here](https://github.com/pubky/pubky-nexus/blob/72cf4e47a7248572caf91cadd45a0930da03448e/nexus-common/src/db/kv/traits.rs#L553)

nexus-common/src/models/tag/search.rs
```diff
@@ -1,163 +1,56 @@
...
-        // Make sure that post does not have more taggers with that tag. Post:Taggers:user_id:post_id:label
-        if label_taggers.is_none() {
-            let key_parts = [&TAG_GLOBAL_POST_TIMELINE[..], &[tag_label]].concat();
-            let post_key = format!("{}:{}", author_id, post_id);
-            Self::remove_from_index_sorted_set(None, &key_parts, &[&post_key]).await?;
-        }
-        Ok(())
+    // TODO Needed? When are tags removed?
+    pub async fn del_from_index(tag_label: &str) -> Result<(), DynError> {
+        Self::remove_from_index_sorted_set(Some("Sorted"), &TAGS_LABEL, &[tag_label]).await
```
tipogi, line 54: Same reason here for the `prefix` param

nexus-watcher/src/events/handlers/tag.rs
```diff
@@ -26,32 +27,26 @@ pub async fn sync_put(
...
-        other => Err(format!(
-            "The tagged resource is not Post or User resource. Tagged resource: {:?}",
-            other
-        )
-        .into()),
-    }
+    }?;
+
+    // After successful put_sync of the tagged resource, we index the tag label
+    TagSearch::put_to_index(&tag.label).await
```
tipogi, line 49: I assume you left it out to avoid duplicating the same function—but is it worth it? It isn’t wrong, but we may lose the parallelism that `tokio::join!` provides

nexus-common/src/models/tag/search.rs
ok300: Good idea, done in d2d390b717ff57f6c63925510692fdd94989cc0c

nexus-api/src/routes/v0/endpoints.rs
ok300: Done in 84e97ef844eb9ed091f0d6503be2e663852d1558

nexus-common/src/models/tag/search.rs
ok300: Interesting, wasn't aware of that. Done 0242847bc91a02164bc0c2d20e45700467e6994e

nexus-common/src/models/tag/search.rs
ok300: Done 0242847bc91a02164bc0c2d20e45700467e6994e

nexus-watcher/src/events/handlers/tag.rs
ok300: Actually, I left it out (i.e. called it after `put_sync`) to make sure it's only called on successful `put_sync`.

If something were to fail and we already indexed the new tag, we'd have to remove it.

nexus-common/src/db/graph/queries/get.rs
ok300: I added a benchmark in bfd23d3d3877caba0646b8dd4e7edb48cc984b44 and ran

```
cargo bench -p nexus-api --bench search
```

with and without that index.

The results were essentially the same:

- with index:
```
tag_search/he           time:   [570.54 µs 580.16 µs 588.39 µs]
                        change: [-7.6137% +0.8434% +7.7533%] (p = 0.87 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low severe
  2 (2.00%) low mild
  1 (1.00%) high mild
  2 (2.00%) high severe
```

- without the index:
```
tag_search/he           time:   [557.03 µs 575.59 µs 588.94 µs]
                        change: [-6.8058% -0.4279% +9.1004%] (p = 0.94 > 0.05)
                        No change in performance detected.
Found 11 outliers among 100 measurements (11.00%)
  6 (6.00%) low severe
  3 (3.00%) low mild
  1 (1.00%) high mild
  1 (1.00%) high severe
```

nexus-watcher/src/events/handlers/tag.rs
tipogi: make sense! can you commented out the reason to be out?

nexus-common/src/db/graph/queries/get.rs
ok300: Interesting, I didn't destroy the DB between runs, I only re-generated the mock data with `db mock`.

I re-ran the benchmark with destroying and re-building the DB before each run:

```
cargo run -p nexusd -- db clear
cargo run -p nexusd -- db mock
```

and I'm getting similar results.

With `taggedLabelIndex`:

```
tag_search/he           time:   [595.08 µs 600.09 µs 605.35 µs]
                        change: [-2.0248% -0.3378% +1.1342%] (p = 0.69 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low mild
  4 (4.00%) high mild
  1 (1.00%) high severe
```

Without `taggedLabelIndex`:

```
tag_search/he           time:   [589.27 µs 593.28 µs 597.46 µs]
                        change: [-2.0811% -0.7088% +0.6329%] (p = 0.32 > 0.05)
                        No change in performance detected.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild
  1 (1.00%) high severe
```

Maybe its because `TagSearch::get_by_label` used in the benchmark doesn't hit the neo4j, but its resolved by the redis sorted set?

tipogi (approved):
LGTM 🚀

nexus-watcher/src/events/handlers/tag.rs
tipogi: until we do not have a rollback strategy, both of them has it own pros and cons. I do not have a preference

nexus-common/src/models/tag/search.rs
```diff
@@ -1,163 +1,56 @@
...
-                    &key_parts,
-                    &[(post_details.indexed_at as f64, &member_key)],
-                    None,
-                    None,
-                )
-                .await?;
-            }
-        }
-        Ok(())
+    pub async fn put_to_index(tag_label: &str) -> Result<(), DynError> {
```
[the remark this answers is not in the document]
ok300: Good idea, done: 23971a515c66939866995a0c732ea3cb78779d18

nexus-common/src/db/graph/queries/get.rs
ok300: I ran the `nexus-api` benchmarks with and without that index, it didn't find any queries negatively impacted by removing it.

However I browsed around and found at least one query that does matches on `tag.label`:

https://github.com/pubky/pubky-nexus/blob/54768db5be1ef5b0cfbf5019b4220b2d02709913/nexus-common/src/db/graph/queries/get.rs#L402-L403

Should I keep the index for this query, and possibly future ones that access `tag.label`?

Or indeed better to remove it?

Asking because the index was there before this PR, so I figure it must have served a purpose.

SHAcollision (approved):
LGTM! :rocket:

Congrats on first feature merged :balloon:

PR #458, opened 2025-05-26
Speed up local responsiveness by removing unnecessary recompilations

ok300:
Before this PR, running any command that involved `nexus-api` code, like

```sh
cargo run -p nexusd -- db clear
```

resulted every time in ~7s of recompilation, even when no code was changed.

This is happening because the `build.rs` lines

```rust
println!("cargo:rerun-if-changed=.git/HEAD");
println!("cargo:rerun-if-changed=.git/refs/heads");
```

referenced files and folders in the workspace root's `.git` directory, from within a workspace member.

The path was incorrect, which led `cargo` to believe these files changed (cause they are missing => they were likely removed), which caused re-compilation on every run.

This PR fixes the paths, which removes unnecessary re-compilations when running commands locally.

### DIFF
--- a/nexus-api/build.rs
+++ b/nexus-api/build.rs
@@ -12,8 +12,4 @@ fn main() {
 
     // Write the commit hash to an environment variable
     println!("cargo:rustc-env=GIT_COMMIT_HASH={}", git_hash);
-
-    // Re-run build.rs if .git/HEAD changes
-    println!("cargo:rerun-if-changed=../.git/HEAD");
-    println!("cargo:rerun-if-changed=../.git/refs/heads");
 }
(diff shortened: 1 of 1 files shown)


ok300:
On second thought, I think entirely removing the git checks is even better (done in 0716dbb77e668e33177c8de886382d2a8ef87272).

Essentially what those 2 lines say is "hey cargo, if there is a new commit, or the branch was changed, then re-compile on next run".

What `cargo` is anyway doing is: if there is a code change, then re-compile on next run.

So, unless we want to trigger a re-compile for local git operations that don't result in a code change (like amend, squash, etc), which doesn't make a lot of sense to do, then sticking with the default cargo behavior looks like a good solution.

What do you think? Is this indeed better, or I missing something?

PR #463, opened 2025-05-27
feat(watcher): wip handle moderators delete tags

SHAcollision:
This PR adds the watcher handling needed to de-index content based on moderated tags policy and pubky_id of moderation source of trust.


examples/watcher/main.rs
```diff
@@ -34,19 +34,23 @@ async fn main() -> Result<(), DynError> {
         None => {
             let homeserver =
                 PubkyId::try_from("8um71us3fyw6h8wbcxb5ar3rwusy1a6u49956ikzojg3gcwd1dty").unwrap();
+            let moderation_id =
+                PubkyId::try_from("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa").unwrap();
```
SHAcollision, line 38: It's a placeholder that imitates what we do for `homeserver` . None of these two is right or good practice. But let's tackle on next PRs.

nexus-watcher/src/events/handlers/tag.rs
```diff
@@ -33,6 +36,16 @@ pub async fn sync_put(
     match parsed_uri.resource {
         // If post_id is in the tagged URI, we place tag to a post.
         Resource::Post(post_id) => {
+            // Moderate content
+            if tagger_id == moderation.id && moderation.tags.contains(&tag.label) {
```
SHAcollision, line 40: possibly best to extract to a new `fn`

nexus-watcher/tests/utils/moderator_key.pkarr
```diff
@@ -0,0 +1,2 @@
+pubky.org/recovery
```
SHAcollision, line 1: Interesting, never noticed that recovery files have a plaintext inscription with a broken link :)

nexus-watcher/src/events/mod.rs
```diff
@@ -150,7 +151,7 @@ impl Event {
                 handlers::bookmark::sync_put(user_id, bookmark, bookmark_id).await?
             }
             (PubkyAppObject::Tag(tag), Resource::Tag(tag_id)) => {
-                handlers::tag::sync_put(tag, user_id, tag_id).await?
+                handlers::tag::sync_put(tag, user_id, tag_id, moderation).await?
```
tipogi, line 154: what about if we handle here if it is `moderation_tag` and the owner of that tag is the `moderation pubky`? Like this we do not mix the scopes of tag handler for common tags and for moderated tags

```rust
(PubkyAppObject::Tag(tag), Resource::Tag(tag_id)) => {
   if tagger_id == moderation.id && moderation.tags.contains(&tag.label) {
       moderation::tag::xxxx.await?
   } else. {
       handlers::tag::sync_put(tag, user_id, tag_id, moderation).await?
   }
}
```

tipogi: why don't we create a notification for that user that its post was deleted because it violates some rule of the app

dzdidi: It is done by moderation bot

tipogi: but having already the tags specified and knowing the behaviour that it will have in that scenarios, does it make sense to pass that responsibility to moderation service (I guess creating the post referencing the user) when nexus could create that notification when the tag arrives?

tipogi: also if we want to have more specific information about the bot, we could query to the moderation service and get more broad info

SHAcollision: All of these ideas were discussed a few weeks ago.

Your proposal would not work for notifying content that the moderation service directly deletes on the homeserver: it does not create tags on it, it simply deletes it and a new `DEL pubky://` event line is created. Therefore for those it must create a post anyway to notify the user.

> we could query to the moderation service
Why create more API when we are building a communication app and a whole set of tools for this. It's best if the moderation user creates the "notifications" directly as posts in the homeserver, it makes no assumptions and works using pubky-app as intended.

nexus-watcher/src/events/handlers/tag.rs
SHAcollision: Revisited this, and we really don't gain much with it.

nexus-watcher/src/events/mod.rs
SHAcollision: > what about if we handle here if it is `moderation_tag` and the owner of that tag is the `moderation pubky`? Like this we do not mix the scopes of tag handler for common tags and for moderated tags
> 
> ```rust
> (PubkyAppObject::Tag(tag), Resource::Tag(tag_id)) => {
>    if tagger_id == moderation.id && moderation.tags.contains(&tag.label) {
>        moderation::tag::xxxx.await?
>    } else. {
>        handlers::tag::sync_put(tag, user_id, tag_id, moderation).await?
>    }
> }
> ```

Pushed a refactor along this lines. Take a look. it's more verbose, but cleaner.

nexus-watcher/src/events/moderation.rs
```diff
@@ -0,0 +1,43 @@
...
+    // Tags to be moderated (tagged content is deleted)
+    pub tags: Vec<String>,
+}
+
+impl Moderation {
+    pub async fn should_delete(&self, tag: &PubkyAppTag, tagger_id: PubkyId) -> bool {
+        tagger_id == self.id && self.tags.contains(&tag.label)
+    }
+
+    pub async fn apply_moderation(tag: PubkyAppTag) -> Result<(), DynError> {
```
ok300, line 18: Maybe rename `tag` to `moderaton_tag` or `mod_tag`, to avoid confusion with the moderated tag in the `Resource::Tag` branch below.

ok300 (approved):

ok300 (approved):

PR #466, opened 2025-05-28
Add nexus static files to `.gitignore`

ok300:
This is a cosmetic PR that improves dev experience.

Each time mock data is added and the tests are run, this generates quite a few static files in `nexus-api` and `nexus-watcher`. The git client shows them as untracked files, which clutters the UI and risks accidentally including them in a commit.

So this PR adds these static files to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@ docker/.env
 docker/.database*
 # Files
 /examples/static
+/nexus-api/static
+/nexus-watcher/static
 # Build artifacts
 /target
 # System-specific files


tipogi (commented):
in which scenarios does it create that files?

SHAcollision (approved):
Looks good, I believe to ave seen these also being created after the recent refactor: not sure exactly what are the conditions for them to be created in that path though

ok300:
@tipogi they show up after I run

```
cargo nextest run
```

Some tests focus on files under `static/files/`, like:

https://github.com/pubky/pubky-nexus/blob/7d6b845e16893cc1991df491019ca130cb5074a8/nexus-api/tests/files/image.rs#L18-L37

tipogi (approved):

PR #467, opened 2025-05-28
Add autocomplete endpoint: user search by PK prefix

### LINKED ISSUE #446: feat(api): search user by pubky prefix match

ok300:
This PR adds a new endpoint `/search/users//by_id/{prefix}` that returns a list of User IDs based on the given ID prefix.

Fixes #446

### DIFF
--- a/nexus-common/src/models/user/search.rs
+++ b/nexus-common/src/models/user/search.rs
@@ -1,11 +1,14 @@
 use super::UserDetails;
 use crate::db::RedisOps;
+use crate::models::create_zero_score_tuples;
 use crate::{models::traits::Collection, types::DynError};
 use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
 pub const USER_NAME_KEY_PARTS: [&str; 2] = ["Users", "Name"];
+pub const USER_ID_KEY_PARTS: [&str; 2] = ["Users", "ID"];
 
+/// List of user IDs
 #[derive(Serialize, Deserialize, ToSchema, Default)]
 pub struct UserSearch(pub Vec<String>);
 
@@ -13,12 +16,12 @@ impl RedisOps for UserSearch {}
 
 impl UserSearch {
     pub async fn get_by_name(
-        name: &str,
+        name_prefix: &str,
         skip: Option<usize>,
         limit: Option<usize>,
     ) -> Result<Option<Self>, DynError> {
         // Perform the lexicographical range search
-        let elements = Self::get_from_index(name, skip, limit).await?;
+        let elements = Self::get_from_index_name(name_prefix, skip, limit).await?;
 
         // If elements exist, process them to extract user_ids
         if let Some(elements) = elements {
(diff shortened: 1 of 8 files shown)


nexus-api/src/routes/v0/endpoints.rs
```diff
@@ -46,7 +46,9 @@ pub const STREAM_TAGS_REACH_ROUTE: &str =
 
 // -- SEARCH endpoints --
 const SEARCH_PREFIX: &str = concatcp!(VERSION_ROUTE, "/search");
-pub const SEARCH_USERS_ROUTE: &str = concatcp!(SEARCH_PREFIX, "/users");
+const SEARCH_USERS_ROUTE: &str = concatcp!(SEARCH_PREFIX, "/users");
+pub const SEARCH_USERS_BY_NAME_ROUTE: &str = concatcp!(SEARCH_USERS_ROUTE, "/by_name/{prefix}");
```
tipogi, line 50: I guess that endpoint is a breaking change. if it is, we have to be careful merging

nexus-common/src/models/user/search.rs
```diff
@@ -38,22 +40,48 @@ impl UserSearch {
...
+        skip: Option<usize>,
+        limit: Option<usize>,
+    ) -> Result<Option<Self>, DynError> {
+        // Perform the lexicographical range search
+        let elements = Self::get_from_id_index(id_prefix, skip, limit).await?;
+
+        Ok(elements.map(UserSearch))
+    }
+
+    pub async fn get_from_name_index(
```
tipogi, line 54: I’m not sure about that, but what if we follow the model’s pattern for retrieving data from indexes (`get_from_index`)? In that case, we could append the index name at the end—e.g., `get_from_index_name`. It might not be very descriptive, but at least the first part clearly indicates the pattern. Same for `get_from_index_id`

nexus-common/src/models/user/search.rs
```diff
@@ -67,46 +95,34 @@ impl UserSearch {
...
-            None,
-            None,
-        )
-        .await
+        Self::put_index_sorted_set(&USER_NAME_KEY_PARTS, &Self::tuples(&pairs), None, None).await?;
+        Self::put_index_sorted_set(&USER_ID_KEY_PARTS, &Self::tuples(&ids), None, None).await
+    }
+
+    /// Create tuples for the sorted set, by including the 0.0 score for each element and converting it to a &str
+    fn tuples(strings: &[String]) -> Vec<(f64, &str)> {
```
tipogi, line 116: Maybe name it as `create_zero_score_tuple` and add a comment related?

```rust
/// Create tuples with a 0.0 score for each element, forcing the sorted set to support lexicographical search
fn create_zero_score_tuple(strings: &[String]) -> Vec<(f64, &str)> {
```

tipogi (commented):
for my side, the implementation is ready, missing the tests in the watcher side. Could you add some?
- In user creation is missing the control of both indexes: `USER_NAME_KEY_PARTS` and `USER_ID_KEY_PARTS` (raw.rs)
- In deletion, we have to add the new key

SHAcollision (commented):
Nice work! I left a few minor comments

nexus-api/src/routes/v0/search/users.rs
```diff
@@ -56,6 +54,49 @@ pub async fn search_users_handler(Query(query): Query<SearchQuery>) -> Result<Js
...
+    )
+)]
+pub async fn search_users_by_id_handler(
+    Path(prefix): Path<String>,
+    Query(query): Query<SearchQuery>,
+) -> Result<Json<UserSearch>> {
+    let id_prefix = prefix;
+    if id_prefix.trim().is_empty() {
+        return Err(Error::invalid_input("ID cannot be empty"));
+    }
```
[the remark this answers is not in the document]
ok300: Done: 824ef8c7a2a89ba2a396678107692da7a16338fa

tipogi (approved):
what ever you decide about SHA comment, for my side also ready!

PR #472, opened 2025-06-02
User lookup: handle case when username includes semicolon

ok300:
The `Sorted:Users:Name` index in Redis contains contais entries of the form `user_name:user_pk`.

Since usernames can themselves contain semicolon, extracting the PK from `user_name:user_pk` has to account for that as well.

This PR updates the way the user PK is extracted from such an entry, by reading the PK as the last element when the entry is split by semicolon.


tipogi (approved):

SHAcollision (approved):
LGTM

PR #485, opened 2025-06-11
Bump `clap` dependency

ok300:
This PR bumps `clap` to the latest version and moves its definition in the workspace `Cargo.toml`.

Supersedes #478


tipogi (approved):
good catch!

PR #489, opened 2025-06-26
Fix influencer score discrepancy

### LINKED ISSUE #488: bug(service): influencer list disagreement redis vs graph
When looking into the "Influencer" list of users currently, by merely looking into number of posts, number of tags and number of followers, I know my user (SHACollision) should be 3rd, in front of @SeverinAlexB 

Image

When verifyin in Redis. That is, in fact, true. My user is third, instead of fourth:

Image

However, the response from nexus places me fourth https://nexus.pubky.app/v0/stream/users?viewer_id=operrr8wsbpr3ue9d4qj41ge1kcc6r7fdiy6o3ugjrrhi4y77rdo&source=influencers&limit=5

My hypothesis for this behaviour is that this  query is ALWAYS hitting the graph and the computation graph vs index is slightly different. It should NEVER hit the graph because it's very expensive. The graph computation of influencers was added on #234 and I believe it's almost guaranteed this bug did not exist before #234.

Ideally, influencers scores are always static in Redis and not computed in the graph on a request->response cycle. If the graph is needed to have faithful scoring, it should be a daily midnight cron.

ok300:
This PR:
- Changes the influencer lookup, so that for the default `Timeframe::AllTime`, the sorted set is used instead of a cache.
  - As a consequence of this, the influencer score is updated with every user interaction, so the new influencer sorting is immediately visible in the UI, on next refresh.
- Fixes a discrepancy in influencer score calculation, where sometimes a slightly different formula was used.

Fixes #488

### DIFF
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -539,7 +539,7 @@ pub fn get_influencers_by_reach(
         WITH DISTINCT reach
 
         CALL (reach) {{
-            MATCH (reach)-[follow:FOLLOWS]->(others:User)
+            MATCH (others:User)-[follow:FOLLOWS]->(reach)
             RETURN count(DISTINCT follow) as followers_count
         }}
         CALL (reach) {{
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -556,7 +556,7 @@ pub fn get_influencers_by_reach(
         WITH reach, followers_count, tags_count, posts_count
         WITH {{
             id: reach.id,
-            score: (tags_count + posts_count) * sqrt(followers_count)
+            score: (tags_count + posts_count) * sqrt(followers_count + 1)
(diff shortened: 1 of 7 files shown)


tipogi (approved):

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -594,7 +594,7 @@ pub fn get_global_influencers(skip: usize, limit: usize, timeframe: &Timeframe)
              COUNT(DISTINCT post) AS posts_count
         WITH {
             id: user.id,
-            score: (tags_count + posts_count) * sqrt(followers_count + 1)
+            score: (tags_count + posts_count) * sqrt(followers_count)
```
SHAcollision, line 597: Interesting. I did not know the `+ 1` was here. It makes sense that if a user has 0 followers, his score is absolute 0 even though he might have a million of posts. It's the best way to protect from users climbing up the leaderboard that are just spamming.

SHAcollision (approved):
Looks good to me :rocket:

PR #502, opened 2025-07-11
Use own config file names for `ApiConfig`, `WatcherConfig`

ok300:
This PR
- changes the `ApiConfig` config filename from `config.toml` to `api-config.toml`
- changes the `WatcherConfig` config filename from `config.toml` to `watcher-config.toml`
- updates relevant examples and docs
- adds error handling to `expand_home_dir`, which is used in the config file path handling
- removes the `ConfigReader` trait and moves its methods to `DaemonConfig`. Its `read_config_file` / `write_default_config_file` logic, which tries to read the config at a given location and writes a default config file if nothing is found, only applies to `DaemonConfig` as there are no default config files for `ApiConfig` or `WatcherConfig`, which used to derive this trait

### DIFF
--- a/nexus-common/src/config/stack.rs
+++ b/nexus-common/src/config/stack.rs
@@ -4,7 +4,6 @@ use std::{fmt::Debug, path::PathBuf};
 
 use super::{file::validate_and_expand_path, Level, LOG_LEVEL};
 
-/// Custom deserializer: take a String, expand `~`, clean up `.`/`..`, return PathBuf.
 fn deserialize_and_expand<'de, D>(deserializer: D) -> Result<PathBuf, D::Error>
 where
     D: Deserializer<'de>,


examples/watcher/main.rs
```diff
@@ -1,16 +1,16 @@
...
+    db::DatabaseConfig, file::try_expand_home_dir, get_files_dir_pathbuf, types::DynError,
+    StackConfig, WatcherConfig, LOG_LEVEL,
 };
 use nexus_watcher::{NexusWatcher, NexusWatcherBuilder};
 use pubky_app_specs::PubkyId;
 use std::path::PathBuf;
 
 #[derive(Parser)]
-#[command(about = "Example Nexus API server", long_about = None)]
+#[command(about = "Example Nexus Watcher server", long_about = None)]
```
SHAcollision, line 11: nice catches

nexus-common/src/config/mod.rs
```diff
@@ -1,10 +1,32 @@
...
+        .clone()
+}
+
 // All the tests run inside their own crate therefore the default directory does not apply
 pub const FILES_DIR_TEST: &str = "./static/files";
+static FILES_DIR_TEST_PATHBUF: OnceLock<PathBuf> = OnceLock::new();
+pub fn get_files_dir_test_pathbuf() -> PathBuf {
+    FILES_DIR_TEST_PATHBUF
+        .get_or_init(|| {
+            try_expand_home_dir(PathBuf::from(FILES_DIR_TEST))
```
SHAcollision, line 25: Is`try_expand_home_dir()` needed here? The `./static/files` is relative to where tests were run. 

Maybe we want to use `try_expand_home_dir()` to validate it's a folder and not a file. But it's a bit confusing that the function is named `try_expand_home_dir()` as I would expect it to return an `Error` if there is no `~` to expand. Maybe renaming or documenting is necessary.

SHAcollision (commented):
Hey @ok300 the PR looks good! Just left a couple of comments.

nexus-watcher/src/builder.rs
```diff
@@ -95,18 +97,29 @@ impl NexusWatcher {
         NexusWatcherBuilder::default()
     }
 
-    /// Loads the configuration from a file and starts the Watcher
+    /// Loads the [WatcherConfig] from [WATCHER_CONFIG_FILE_NAME] in the given path and starts the Nexus Watcher.
+    ///
+    /// If no [WatcherConfig] file is found, it defaults to [NexusWatcher::start_from_daemon].
```
SHAcollision, line 102: nice

nexus-common/src/config/mod.rs
ok300: Good points.

I agree, it makes sense to keep this check even for the relative path used in tests, to check it's not a file.

And I also agree `try_expand_home_dir` is not the clearest name. I renamed it to `validate_and_expand_path` with a more accurate rustdoc.

PR #509, opened 2025-07-21
Correctly handle tags containing emojis in tag search

ok300:
This PR fixes the tag search issue where tags with unicode characters (like emojis) were not included in the search results.

Fixes https://github.com/pubky/pubky-app/issues/1660


tipogi (approved):
🚀

PR #513, opened 2025-07-25
feat(watcher): allow any testnet hostname

SHAcollision:
# Pre-submission Checklist
This PR  uses latest https://github.com/pubky/pubky-core/pull/189 pubky-client and should allow setting custom `hostnames` for testnet enabled Nexus. This is needed for the dockerized stack to run on platforms other than Linux.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -734,17 +734,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "console_log"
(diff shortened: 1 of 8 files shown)


SpontaneousOverthrow (approved):
Everything is working, but i mentioned that the `testnet_host` is required config option, is that ok?

PR #530, opened 2025-08-14
Propagate use of graph util methods

ok300:
This PR propagates the use of the graph util methods:
- `retrieve_from_graph<T>`
  - which returns `Option<T>`
- `fetch_row_from_graph`
  - which returns `Option<Row>`
- `fetch_all_rows_from_graph`
  - which returns `Vec<Row>`

---

Open questions:
- [x] Rename `retrieve_from_graph` to `fetch_key_from_graph`?


SHAcollision (approved):
Did not dive in depth, but looks convenient and passes all tests :check:

PR #535, opened 2025-08-19
feat: multiple homeserver processing

tipogi:
# Pre-submission Checklist

### DIFF
--- a/nexus-watcher/src/events/processor.rs
+++ b/nexus-watcher/src/events/processor.rs
@@ -12,7 +12,7 @@ use pubky_app_specs::PubkyId;
 use std::error::Error;
 use std::path::PathBuf;
 use tokio::sync::watch::Receiver;
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, warn};
 
 pub struct EventProcessor {
     pub homeserver: Homeserver,
@@ -59,11 +59,13 @@ impl EventProcessor {
         }
     }
 
-    pub async fn from_config(config: &WatcherConfig) -> Result<Self, DynError> {
-        let homeserver = Homeserver::get_by_id(config.homeserver.clone())
+    pub async fn from_config(config: &WatcherConfig, homeserver_id: String) -> Result<Self, DynError> {
+        let homeserver_id = PubkyId::try_from(homeserver_id.as_str()).map_err(|e| DynError::from(e))?;
+        let homeserver = Homeserver::get_by_id(homeserver_id)
             .await?
             .ok_or("Homeserver not found")?;
         let limit = config.events_limit;
+        warn!("----- Homeserver limit: {}", limit);
         let files_path = config.stack.files_path.clone();
         let tracer_name = config.name.clone();
 
@@ -107,7 +109,7 @@ impl EventProcessor {
                 self.process_event_lines(shutdown_rx, lines).await?;
             }
         }
-
+        warn!("----- EventProcessor::run finished: {:?}", self.homeserver.id);
         Ok(())
     }
 
--- a/nexus-watcher/src/builder.rs
+++ b/nexus-watcher/src/builder.rs
@@ -1,19 +1,13 @@
 use std::path::PathBuf;
-
-use crate::events::processor::EventProcessor;
-
 use nexus_common::db::{DatabaseConfig, PubkyClient};
-use nexus_common::file::ConfigLoader;
 use nexus_common::types::DynError;
 use nexus_common::utils::create_shutdown_rx;
-use nexus_common::{DaemonConfig, Level, StackConfig};
+use nexus_common::{Level, StackConfig};
 use nexus_common::{StackManager, WatcherConfig};
 use pubky_app_specs::PubkyId;
 use tokio::sync::watch::Receiver;
-use tokio::time::Duration;
-use tracing::{debug, error, info};
 
-pub const WATCHER_CONFIG_FILE_NAME: &str = "watcher-config.toml";
+use crate::NexusWatcher;
 
 #[derive(Debug, Default)]
 pub struct NexusWatcherBuilder(pub WatcherConfig);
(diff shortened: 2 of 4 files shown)


nexus-common/src/models/homeserver.rs
```diff
@@ -25,6 +26,14 @@ impl Homeserver {
         }
     }
 
+    /// Mutates the cursor of the homeserver
+    pub fn mutate_cursor(&self, cursor: String) -> Self {
```
ok300, line 30: Maybe this can be renamed to `persist_cursor` and include the calls to `put_to_index` / `put_to_graph`?

nexus-common/src/models/homeserver.rs
```diff
@@ -65,6 +74,25 @@ impl Homeserver {
             },
         }
     }
+
+    /// Verifies if homeserver exists, or persists it if missing
+    pub async fn verify_or_persist(homeserver_id: PubkyId) -> Result<(), DynError> {
```
ok300, line 79: `persist_if_unknown` might be a better name?

nexus-common/src/models/homeserver.rs
```diff
@@ -121,4 +149,12 @@ mod tests {
 
         Ok(())
     }
+
+    #[test]
+    fn test_homeserver_cursor_mutation() {
```
ok300, line 154: The test should check if the mutation was persisted.

nexus-watcher/src/events/processor.rs
```diff
@@ -14,6 +14,47 @@ use std::path::PathBuf;
...
+        let homeserver_id = PubkyId::try_from(homeserver_id).map_err(DynError::from)?;
+        let homeserver = Homeserver::get_by_id(homeserver_id)
+            .await?
+            .ok_or("Homeserver not found")?;
+
+        Ok(EventProcessor {
+            homeserver,
+            limit: self.limit,
+            files_path: self.files_path.clone(),
+            tracer_name: self.tracer_name.clone(),
```
ok300, line 52: Shouldn't the `tracer_name` be different for each processor? 🤔

nexus-watcher/src/events/processor.rs
```diff
@@ -14,6 +14,47 @@ use std::path::PathBuf;
...
+            moderation: self.moderation.clone(),
```
ok300, line 53: To avoid cloning (each time, for each processor, for each run), maybe `EventProcessor` can have `Arc` pointers, like so:

```rust
pub struct EventProcessor {
    pub homeserver: Homeserver,
    limit: u32,
    pub files_path: Arc<PathBuf>,
    pub tracer_name: String,
    pub moderation: Arc<Moderation>,
}
```

Then its just a matter of cloning the pointers.

nexus-watcher/src/events/processor.rs
```diff
@@ -59,34 +100,7 @@ impl EventProcessor {
...
-            homeserver,
-            limit,
-            files_path,
-            tracer_name,
-            moderation,
-        })
-    }
-
-    pub async fn run(&mut self, shutdown_rx: Receiver<bool>) -> Result<(), DynError> {
+    pub async fn run(&self, shutdown_rx: Receiver<bool>) -> Result<String, DynError> {
```
ok300, line 103: A simple rustdoc describing what it does and what it returns would be helpful.

nexus-watcher/src/events/processor.rs
```diff
@@ -117,7 +130,7 @@ impl EventProcessor {
     /// using the current cursor and a specified limit. It retrieves new event
     /// URIs in a newline-separated format, processes it into a vector of strings,
     /// and returns the result.
-    async fn poll_events(&mut self) -> Result<Option<Vec<String>>, DynError> {
+    async fn poll_events(&self) -> Result<Option<Vec<String>>, DynError> {
```
ok300, line 133: Nice, if `mut` isn't needed, then it's good to remove it.

ok300, line 133: Would it simplify things if instead of returning `Option<Vec<String>>` it would return `Vec<String>`?

nexus-watcher/src/events/processor.rs
```diff
@@ -158,21 +171,27 @@ impl EventProcessor {
     /// # Parameters
     /// - `lines`: A vector of strings representing event lines retrieved from the homeserver.
     pub async fn process_event_lines(
-        &mut self,
+        &self,
         shutdown_rx: Receiver<bool>,
         lines: Vec<String>,
-    ) -> Result<(), DynError> {
+    ) -> Result<String, DynError> {
+        println!("Processing {:?} event lines", lines);
```
ok300, line 178: `println!` -> `tracing::info!`

nexus-watcher/src/events/processor.rs
```diff
@@ -158,21 +171,27 @@ impl EventProcessor {
...
+        let mut new_cursor = String::new();
         for line in &lines {
             if *shutdown_rx.borrow() {
                 info!("Shutdown detected, exiting event processing loop");
-                return Ok(());
+                return Ok(new_cursor);
             }
 
+            // Cursor is the last line of the event list
             if line.starts_with("cursor:") {
```
ok300, line 187: You can combine both these checks into one:

```rust
if line.starts_with("cursor:") {
    if let Some(cursor) = line.strip_prefix("cursor: ") {
        // ...
    }
} 
```
=>

```rust
if let Some(cursor) = line.strip_prefix("cursor: ") {
    // ...
}
```

nexus-common/src/models/homeserver.rs
tipogi: https://github.com/pubky/pubky-nexus/pull/535/commits/ae536df85346387eaf323e097acb27df114fbda3

nexus-common/src/models/homeserver.rs
tipogi: https://github.com/pubky/pubky-nexus/pull/535/commits/a3709191ffec7d84cff5a94c379e41868ad1df6b

nexus-watcher/src/events/processor.rs
tipogi: thought about that but as the observability layer implementation is paused, we can tackle when we focus in its implementation

PR #542, opened 2025-08-20
Do not recommend deleted users in Who To Follow feed

### LINKED ISSUE #523: [Bug]: Do not recommend on Who To Follow users with [DELETED] username
Describe the bug
Under " Who to Follow" section in Pubky app, recommendations have been given for accounts that have been already deleted.

More context from Chris:
These nodes in the graph still exist and for some reason, when they deleted their content from the frontend, not every post/reply was deleted. Therefore, they still have enough all time activity to be shown on the recommendations

Image
Image

ok300:
Fixes #523


nexus-common/src/db/graph/queries/get.rs
```diff
@@ -899,7 +901,7 @@ pub fn recommend_users(user_id: &str, limit: usize) -> neo4rs::Query {
         MATCH (potential)-[:AUTHORED]->(post:Post)
         WITH potential, COUNT(post) AS post_count
         WHERE post_count >= 5
-        RETURN potential.id AS recommended_user_id
+        RETURN potential.id AS recommended_user_id, potential.name AS recommended_user_name
```
tipogi, line 904: what about if we do that filtering in that layer and add another condition in the `WHERE` clause? Like this we could skip the loop to filter users that has not _[DELETED]_ as name. Each option has its pros and cons but from now, I think we could delegate that work to the graph. What do you think?

tipogi (commented):
as always, cleaning the house a bit 😆 🔝

nexus-common/src/db/graph/queries/get.rs
ok300: 50/50

I went with filtering outside the query because that's how I saw it done in other places, so I assume this is the preferred pattern. I also think generally filtering by `[DELETED]` is somewhat brittle, so if we want to change it or improve it at a later time, it's easier to do if this is done in the code than if it's done in queries.

So for these reasons, I think it's better to keep filtering outside the queries. Would you agree? Or are there other better reasons to go for the query filtering?

PR #586, opened 2025-09-26
Bump `pubky-app-specs`, use new URI builder utilities

ok300:
This PR bumps the `pubky-app-specs` dependency and replaces hardcoded URIs with `xyz_uri_builder` utils.


tipogi (approved):
go!! good clean up!

PR #609, opened 2025-11-03
feat(api): add stream of only post keys

SHAcollision:
- Added a `PostKeyStream` model alongside shared helpers so post streams can surface raw keys without hydrating full post views.
- Re-exported the new stream type and wired a `/v0/stream/posts/keys` route with OpenAPI documentation to deliver lightweight post key streams.
- Added a Redis-aware integration test that compares the key-only stream with the full post stream.


nexus-webapi/benches/streams_benches/keys.rs
ok300: Maybe rename this to `post_keys.rs` as `keys.rs` could mean many things (public keys, etc)

ok300 (commented):
One short NIT, otherwise LGTM.

PR #611, opened 2025-11-05
GET `/events` endpoint

dzdidi:
# Pre-submission Checklist

# Changes:
- [x] **watcher** - move events persistence to common
- [x] **common** - persist event lines in redis sorted by synchronization timestamp
- [x] **webapi** - add rest API to read events

### DIFF
--- a/nexus-watcher/src/events/mod.rs
+++ b/nexus-watcher/src/events/mod.rs
@@ -1,9 +1,8 @@
 use nexus_common::db::PubkyConnector;
-use nexus_common::models::event::EventProcessorError;
+use nexus_common::models::event::{Event, EventProcessorError, EventType};
 use nexus_common::types::DynError;
-use pubky_app_specs::{ParsedUri, PubkyAppObject, Resource};
-use serde::{Deserialize, Serialize};
-use std::{fmt, path::PathBuf, sync::Arc};
+use pubky_app_specs::{PubkyAppObject, Resource};
+use std::sync::Arc;
 use tracing::debug;
 
 pub mod handlers;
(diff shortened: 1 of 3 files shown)


nexus-watcher/src/events/mod.rs
```diff
@@ -91,16 +95,32 @@ impl Event {
...
+        self.store_event().await
+    }
+
+    /// Stores event at redis as a member of stored set.
+    /// Sorting is done by synchronization timestamp.
+    async fn store_event(&self) -> Result<(), DynError> {
+        // We use sync timestamp instead of homeserver cursor (creation timestamp)
+        // because we want to ensure that events are processed in the order they
+        // were received and they can not be inserted in between already processed events
+        let ts_ms = SystemTime::now().duration_since(UNIX_EPOCH)?.as_millis() as f64;
```
ok300: From a different angle: what happens if a few events are processed in the same millisecond on a very fast Nexus machine? Sorting doesn't match processing order anymore.

nexus-watcher/src/events/mod.rs
```diff
@@ -91,16 +95,32 @@ impl Event {
...
+        let line = format!("{} {}", self.event_type, self.uri);
+
+        let elements = vec![(ts_ms, line.as_str())];
+
+        Event::put_index_sorted_set(&["Events"], &elements, None, None).await
```
ok300: Are events only inserted?

Or is there some condition where entries will be deleted? For example after some time, or after they've been checked by moderation.

nexus-watcher/src/events/mod.rs
```diff
@@ -91,16 +95,32 @@ impl Event {
...
+    pub async fn handle(&self, moderation: Arc<Moderation>) -> Result<(), DynError> {
         match self.event_type {
-            EventType::Put => self.handle_put_event(moderation).await,
-            EventType::Del => self.handle_del_event().await,
-        }
+            EventType::Put => self.handle_put_event(moderation).await?,
+            EventType::Del => self.handle_del_event().await?,
+        };
+
+        self.store_event().await
```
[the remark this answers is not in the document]
dzdidi: If I understand correctly:
> When an event cannot be processed when it's received (for example, because it depends on an entity that's not yet known), 

means that either `self.handle_put_event().await?` or `self.handle_del_event().await?` will fail. Thus event will be stored if and only if it has been successfully processed.

nexus-watcher/src/events/mod.rs
dzdidi: see above

dzdidi: also milliseconds it what I see retry manager is using.

nexus-common/src/models/event/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+        limit: usize,
+    ) -> Result<Vec<(String, f64)>, DynError> {
+        let result = Event::try_from_index_sorted_set(
+            &["Events"],
+            Some(cursor),
+            None,
+            None,
+            Some(limit + 1),
+            SortOrder::Descending,
+            None,
```
SHAcollision: Pagination boundary is inverted. `get_events_from_redis` calls `try_from_index_sorted_set` with `start=Some(cursor)` and `end=None`, but the trait implementation swaps those positions before hitting Redis. I am not sure how this does affect, I guess the default `cursor` of `0` the query becomes “max score 0, min -∞”, so a new call returns nothing even if events exist, and follow-up pagination keeps walking backwards? It somehow looks like a logic bug `/v0/events` maybe worth a double check.

nexus-common/src/models/event/mod.rs
```diff
@@ -0,0 +1,139 @@
...
+    /// Sorting is done by synchronization timestamp.
+    pub async fn store_event(&self) -> Result<(), DynError> {
+        // We use sync timestamp instead of homeserver cursor (creation timestamp)
+        // because we want to ensure that events are processed in the order they
+        // were received and they can not be inserted in between already processed events.
+        //
+        // At the same time the timestamp is the same as it is used in homeserver in order
+        // to simplify migration from homeserver /events endpoint to nexus /events endpoint
+        // aiming to make them mutually replaceable from the consumer standpoint.
+        let ts_ms = SystemTime::now().duration_since(UNIX_EPOCH)?.as_micros() as f64;
```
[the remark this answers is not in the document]
dzdidi: With list there is no expiration so clean up will need to be done manually. 
Regarding loss... Truncation happens only in one place.  The conversions flow is: `microseconds ts (u128) - (truncates and produces closest possible float) -> redis key (f64) - (rounds towards zero) -> api response (u64) - (no cast)  -> api request (u64) - (closes possible float) -> redis key (f64)`. Yes, the loss due truncation is possible at `ts` closer to higher boundaries of `u64`  in ~500 years. Loss due rounding is also possible but only at close enough numbers. Events are stored one by one with full event processing between them so I think there is enough time to avoid collision

nexus-webapi/src/routes/v0/events.rs
```diff
@@ -86,7 +86,7 @@ fn assemble_page(items: Vec<(String, f64)>) -> EventsList {
 }
 
 fn parse_query(q: &EventsQuery) -> Result<(usize, f64), Error> {
-    let limit = q.limit.unwrap_or(500);
+    let limit = q.limit.unwrap_or(500).min(1000);
```
ok300: Was this meant to be `.max(1000)`?

dzdidi:
For some reason GH decided that my replies to your comments were review and I needed to explicitly submit them :facepalm:

nexus-common/src/models/event/mod.rs
```diff
@@ -0,0 +1,125 @@
...
+                    message: format!("Unknown resource in URI: {uri}"),
+                }
+                .into())
+            }
+            // Known resources not handled by Nexus
+            Resource::LastRead | Resource::Feed(_) | Resource::Blob(_) => return Ok(None),
+            _ => (),
+        };
+
+        let event_line = line.to_string();
```
dzdidi: Not sure if this is enough or if `files_path` should also be included

ok300 (approved):

SHAcollision (approved):
Lgtm

PR #612, opened 2025-11-10
Feat: Remove cross compiler toolchain for aarch64

SpontaneousOverthrow:
# Pre-submission Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 # ========================
 # Build Stage
 # ========================
-FROM rust:1.87.0-alpine3.22 AS builder
+FROM rust:1.90.0-alpine3.22 AS builder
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,10 +1,10 @@
 # ========================
 # Build Stage
 # ========================
-FROM rust:1.87.0-alpine3.22 AS builder
+FROM rust:1.90.0-alpine3.22 AS builder
 
 # Build platform argument (x86_64 or aarch64) (default: x86_64)
-ARG TARGETARCH=x86_64
+ARG TARGETARCH
```
catch-21: can remove all `TARGETARCH`

SpontaneousOverthrow: want to keep it now for debugging, will delete when we have a tested pipeline for building/pushing images to both registries.

PR #614, opened 2025-11-12
Expose `MAX_HOMESERVERS_PER_RUN` as configurable in `WatcherConfig`

ok300:
This PR renames and exposes `MAX_HOMESERVERS_PER_RUN` as a new config field, `WatcherConfig::monitored_homeservers_limit`.

This allows a Nexus instance to limit how many homeservers are monitored on each event processor run. Setting this config to `1` limits monitoring only to the default homeserver (supersedes #613).

The PR also improved the rustdoc of a few related `WatcherConfig` fields.

### DIFF
--- a/nexus-common/default.config.toml
+++ b/nexus-common/default.config.toml
@@ -18,7 +18,9 @@ testnet = false
 testnet_host = "localhost"
 # Synonym homeserver pubky
 homeserver = "8um71us3fyw6h8wbcxb5ar3rwusy1a6u49956ikzojg3gcwd1dty"
+# Maximum number of events to fetch per run from each homeserver
 events_limit = 50
+# Maximum number of monitored homeservers. If set to 1, only the default homeserver is monitored.
 monitored_homeservers_limit = 50
 watcher_sleep = 5000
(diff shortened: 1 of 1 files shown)


nexus-common/default.config.toml
```diff
@@ -19,6 +19,7 @@ testnet_host = "localhost"
 # Synonym homeserver pubky
 homeserver = "8um71us3fyw6h8wbcxb5ar3rwusy1a6u49956ikzojg3gcwd1dty"
 events_limit = 50
+monitored_homeservers_limit = 50
```
SHAcollision: Needs a comment explaining what this is and what it does. Ideally `event_limit` has a comment too.

SHAcollision (approved):
LGTM 🚀

nexus-common/default.config.toml
ok300: Done: 2669389ddc18feaa6191fa9f895d633a62420b49

tipogi (approved):
Nice! Like the rustdoc how is pointing to config  📝

PR #642, opened 2025-12-03
chore: user not indexed in bootstrap

tipogi:
## Summary

Enables bootstrap endpoint to work during sign-up by returning a valid response for non-indexed users instead of 404. Adds `indexed: bool` flag to indicate user status.

## Key Changes

1. **Added `indexed` field** to `Bootstrap` struct
2. **Changed return type** from `Result<Option<Bootstrap>>` to `Result<Bootstrap>` - always returns a response
3. **API endpoint** now returns 200 OK instead of 404 for non-indexed users
4. **Conditional logic** - skips user recommendations for non-indexed users
5. **Updated tests** - verifies non-indexed user behavior instead of expecting 404

## Improvements

✅ **Better UX** - Works during sign-up flow, clients can populate UI with global data  
✅ **Consistent API** - Always returns Bootstrap payload with clear `indexed` status  
✅ **Graceful degradation** - Non-indexed users get influencers/hot tags, skip personalized recommendations  
✅ **Better error handling** - Avoids unnecessary API calls and potential errors

# Pre-submission Checklist

### DIFF
--- a/nexus-common/src/models/bootstrap.rs
+++ b/nexus-common/src/models/bootstrap.rs
@@ -52,19 +52,19 @@ impl Bootstrap {
     pub async fn get_by_id(user_id: &str, view_type: ViewType) -> Result<Self, DynError> {
         let mut bootstrap = Self::default();
         let mut user_ids = HashSet::new();
-        let mut viewer_id = None;
+        let mut maybe_viewer_id = None;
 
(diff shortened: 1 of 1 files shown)


ok300 (commented):
Small naming NIT, otherwise LGTM.

PR #662, opened 2026-01-02
Prevent event reindexing (on same HS) by only allowing a cursor to change incrementally

ok300:
This PR brings in https://github.com/ok300/pubky-nexus/pull/12 .


SHAcollision (commented):
Hey @ok300 , nice work!

Two concerns:
1. Backwards compatibility: I am not sure about this, but maybe the previous Homeserver cursors were previously stored in Redis as strings because of their crockford encoding (e.g., `"0000000000000"`) ?. With the field now defined as `u32` and deserialized directly from Redis, any existing cursor will fail to deserialize, causing `get_from_index`/`validate_cursor_change` to error before processing. Maybe a data migration or a custom deserializer that accepts both strings and numbers is needed to avoid runtime failures during rollout. Have you tried running this fix over an existing homeserver?

2. Cursor range maybe be too small?: The new `u32` type limits cursor values to ~4.29B. If homeserver cursors can exceed this (the former 13-digit likely did), `.parse::<u32>()` will reject valid cursors and prevent further updates. Is the cursor on `pubky-core/pubky-homeserver` also `u32` ? We could consider `u64` (or string+numeric parsing) if we might sometimes expect larger cursors.

SHAcollision (approved):
Given what we've learned on #860 I'd say we need to merge here and deploy to staging with higher urgency. LGMT.

nexus-common/src/models/homeserver.rs
```diff
@@ -94,6 +118,19 @@ impl Homeserver {
         }
     }
 
+    async fn validate_cursor_change(id: &str, new_cursor: u64) -> RedisResult<()> {
+        // If we already indexed a value, reject cursors going below it to prevent reindexing past events
+        if let Some(hs_from_index) = Self::get_from_index(id).await? {
```
aintnostressin, line 123: It uses the same `json::get` fn as described in linked issue. If it returns `None` then validation pass.

SHAcollision: We've done this as hotfix via https://github.com/pubky/pubky-nexus/pull/861

SHAcollision:
Extracted the `get_by_id` graph-fallback removal into #861 to ship the urgent fix on its own. This PR keeps the `validate_cursor_change` monotonicity invariant, which is complementary to #861 (it covers the in-flight-batch overwrite at `processor.rs:111-115` and any future backward write).

SHAcollision:
Confirming the current gap this PR addresses, with the exact site: `process_user_events` persists the **last** event's cursor id, not the max, and the write has no monotonicity guard.

- `let cursor_id = stream_event.cursor.id();` then `latest_cursor = Some(cursor_id);` (`service/indexer/key_based.rs:316-338`), i.e. the last event in the batch.
- `UserHsCursor::write` calls `put_index_sorted_set` with the new score unconditionally (`user/cursor.rs:62-65`); `read` only guards against negative scores.

Since the cursor id originates entirely from the homeserver-supplied `StreamEvent`, a HS can rewind the cursor (forcing reindexing, the #860/#865 class) or fast-forward it (skipping events). The fix should clamp the write to `max(stored, new)` and track the max over the batch.

ok300:
Superseded by #997

PR #663, opened 2026-01-02
`Post`, `Tagged`: do not update `indexed_at` on HS re-indexing

ok300:
This PR brings in https://github.com/ok300/pubky-nexus/pull/13 .


ok300:
@SHAcollision your comment from https://github.com/ok300/pubky-nexus/pull/13#issuecomment-3574920546

> Can we prevent the `Tagged` relationship's `indexed_at` from updating on reindexing as well? Looks like this also broke on the reindexing event of last week.

is addressed in 8a033a966e63a7a6cc933842a99bcf105d0451b1

SHAcollision (approved):
LGTM!

PR #669, opened 2026-01-06
Update dependabot

ok300:
This PR consolidates all cargo dependency updates into a single weekly PR instead of creating individual PRs per dependency.


tipogi (approved):
Nice!

PR #671, opened 2026-01-07
test: use `EphemeralTestnet` singleton in `nexus-watcher` tests

ok300:
This PR simplifies test setup in the `nexus-watcher` suite by only initializing the test stack and the `EphemeralTestnet` once, then re-using them across the `nexus-watcher` tests.


tipogi (approved):

SHAcollision (approved):
IMO good to merge. But i thought it might be relevant to raise for you this Codex comment:

> 
> 3. **`create_external_test_homeserver()` depends on global initialization order**
>    - The helper now errors if `SHARED_TESTNET` is not initialized, which is correct but fragile. If any future test calls this helper without first going through `WatcherTest::setup()`, it will return an error (`SHARED_TESTNET not initialized`) and fail unexpectedly.
>    - Recommendation: either make `create_external_test_homeserver()` lazily initialize the testnet, or document the required initialization order clearly.
>

ok300:
Unfortunately this has no effect on test execution, because `nextest` runs each test in a separate process.

This means the singleton is only visible to one test due to process separation, which means the tests still end-up spinning up an `EphemeralTestnet` per test.

PR #675, opened 2026-01-09
Remove unnecessary Mutex wrapper around the Neo4j Graph handle

ok300:
This PR removes the superfluous `Mutex` wrapper around `Graph`, because the `neo4rs::Graph` instance already handles connection pooling internally, is cloneable and thread-safe.

This PR also removes the `Arc` around it, since `Graph` internally consists of `Arcs`.

Finally, it removes an inner nested `OnceCell` in the connector.


SHAcollision (approved):
LGTM

PR #678, opened 2026-01-09
Partition CI tests

ok300:
This PR adds nextest partitioning to split tests across multiple CI runners.


ok300:
The setup and rust binary caching becomes way too complicated, not worth it.

PR #683, opened 2026-01-12
Fix integer overflow in `get_range`

ok300:
`nexus_common::db::kv::index::lists::get_range` contained the following:

```rust
    let skip = skip.unwrap_or(0);
    let limit = limit.unwrap_or(usize::MAX);

    let start = skip as isize;
    let end = start + (limit as isize) - 1;      // can overflow
```

The last line and more specifically casting `limit as isize` can overflow because `limit` can be `usize::MAX` and `usize::MAX` > `isize::MAX`.

This PR addresses this by changing the `get_range` method signature to directly use `isize`. The callers use smaller values in the range of `u32`, which can safely be cast to `isize` when calling this method.

### DIFF
--- a/nexus-webapi/tests/events/mod.rs
+++ b/nexus-webapi/tests/events/mod.rs
@@ -15,7 +15,7 @@ async fn test_events_huge_cursor() -> Result<()> {
         .await?;
     assert_eq!(res.status(), 200);
     let body = res.text_body()?;
-    assert!(body.contains("cursor:"), "response must contain a cursor line");
+    assert!(body.contains("cursor:"), "response must have a cursor line");
 
     Ok(())
 }
@@ -48,7 +48,7 @@ async fn test_events_large_cursor_with_limit() -> Result<()> {
         .await?;
     assert_eq!(res.status(), 200);
     let body = res.text_body()?;
-    assert!(body.contains("cursor:"), "response must contain a cursor line");
+    assert!(body.contains("cursor:"), "response must have a cursor line");
 
     Ok(())
 }


nexus-common/src/db/kv/traits.rs
```diff
@@ -276,8 +276,8 @@ pub trait RedisOps: Serialize + DeserializeOwned + Send + Sync {
     /// Returns an error if the operation fails, such as if the Redis connection is unavailable.
     async fn try_from_index_list(
         key_parts: &[&str],
-        skip: Option<usize>,
-        limit: Option<usize>,
+        skip: Option<isize>,
+        limit: Option<isize>,
```
SHAcollision: Same here, it now accepts negative values.

ok300: Done: 16632fdd67a0f70d9a98132616cef907ab926295

nexus-webapi/src/routes/v0/events.rs
```diff
@@ -29,17 +29,17 @@ impl std::fmt::Display for EventsList {
...
 }
 
 #[utoipa::path(
     get,
     path = EVENTS_ROUTE,
     tag = "Events",
     params(
-        ("cursor" = u64, Query, description = "Cursor"),
-        ("limit" = usize, Query, description = "Limit the number of results, (default 500, maximum 1000)")
+        ("cursor" = u32, Query, description = "Cursor"),
```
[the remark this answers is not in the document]
ok300: Kept `cursor` as `u64`: 905daef0540a2b9e457504f6169778e5137264b1

PR #690, opened 2026-01-15
Replace hardcoded DB defaults with values from `default.config.toml`

ok300:
This PR replaces hardcoded DB params with those read and parsed from `default.config.toml`.

This was expanded to include `StackConfig` and its fields, some of which were also hardcoded (log level, `FILES_DIR`) but for which `default.config.toml` already defined the same defaults.


tipogi (approved):

ok300:
This PR is an incomplete solution to the described problem of duplicated configs.

There are different kinds of "config value defaults", defined in different places:
- field defaults defined programatically
- field defaults defined in `default.config.toml`
- field serde defaults for new config fields which may be missing from `config.toml` (like `default_neo4j_user`)

The tricky part is the 3rd category of defaults, especially when combined with an optional field like `stack.db.neo4j.slow_query_logging_threshold_ms`: in such a case, does the absence of this from `config.toml` mean its intended value is `None`? Or should it then default to a programatic default?

IMO a more comprehensive solution is needed, for example removing the 3rd category of defaults from above (e.g. if a new config field is introduced, offer no "serde default", instead require that every `config.toml` explicitly sets it).

PR #691, opened 2026-01-16
fix: `put_index_list` de-duplicates when appending to list

ok300:
This PR addresses a TODO in `put_index_list` which mentioned the method only inserts and doesn't check for duplicates.

The fix is a modification to `lists::put` to use `LREM + RPUSH` pattern in a pipeline to prevent duplicate entries.

For each value being added:
1. `LREM` removes all existing occurrences from the list (count=0 removes all)
2. `RPUSH` appends the value to the end of the list

Note: If a value already exists, it will be removed from its original position and re-added at the end of the list.

### DIFF
--- a/nexus-common/src/db/kv/index/lists.rs
+++ b/nexus-common/src/db/kv/index/lists.rs
@@ -2,10 +2,12 @@ use crate::db::get_redis_conn;
 use crate::types::DynError;
 use deadpool_redis::redis::AsyncCommands;
 
-/// Adds elements to a Redis list.
+/// Adds elements to a Redis list with deduplication.
 ///
-/// This function appends elements to the specified Redis list. If the list doesn't exist,
-/// it creates a new list.
+/// This function appends elements to the specified Redis list while ensuring no duplicates.
+/// For each value, it first removes any existing occurrences, then appends the value to the end.
+/// This makes the operation idempotent - calling it multiple times with the same value results
+/// in only one occurrence at the end of the list.
 ///
 /// # Arguments
 ///
@@ -22,7 +24,16 @@ pub async fn put(prefix: &str, key: &str, values: &[&str]) -> Result<(), DynErro
     }
     let index_key = format!("{prefix}:{key}");
     let mut redis_conn = get_redis_conn().await?;
-    let _: () = redis_conn.rpush(index_key, values).await?;
+
+    // Use a pipeline to remove duplicates then append each value
+    // LREM removes all existing occurrences (0 = all), then RPUSH appends to end
+    // This is atomic from the perspective of other Redis clients
+    let mut pipe = deadpool_redis::redis::pipe();
+    for value in values {
+        pipe.lrem(&index_key, 0, value).rpush(&index_key, value);
+    }
+    let _: () = pipe.query_async(&mut redis_conn).await?;
+
     Ok(())
 }
 
--- a/nexus-common/src/db/kv/traits.rs
+++ b/nexus-common/src/db/kv/traits.rs
@@ -244,15 +244,12 @@ pub trait RedisOps: Serialize + DeserializeOwned + Send + Sync {
         let prefix = Self::prefix().await;
         let key = key_parts.join(":");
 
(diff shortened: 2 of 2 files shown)


nexus-common/src/db/kv/index/lists.rs
```diff
@@ -22,7 +24,16 @@ pub async fn put(prefix: &str, key: &str, values: &[&str]) -> Result<(), DynErro
     }
     let index_key = format!("{prefix}:{key}");
     let mut redis_conn = get_redis_conn().await?;
-    let _: () = redis_conn.rpush(index_key, values).await?;
+
+    // Use a pipeline to remove duplicates then append each value
+    // LREM removes all existing occurrences (0 = all), then RPUSH appends to end
+    // This is atomic from the perspective of other Redis clients
```
tipogi: That comment is not strictly correct in Redis terms:

- A pipeline sends multiple commands in one batch over the same connection, reducing round trips.
- But it does not make them atomic. Other clients’ commands can still be interleaved between these commands on the server.

To get true atomicity in Redis, you’d need `MULTI` / `EXEC` (transaction) or a Lua script (`EVAL`).

tipogi (approved):

SHAcollision (commented):
Two comments from me: 
1. A much better alternative for list uses that should not have duplicates and where order does not matter might be to use a Redis Set instead. So we might simply identify these use cases (if any) and migrate them to a set?
2. I am not 100% sure that we never want duplicates in ANY of our uses of Redis Lists :thinking: Certainly this makes our implementation of "put_to_list" stop behaving as a regular list, so we need to account for that when we want to implement new features that should behave as lists in the future. In fact I see how we are using the List type for `/events` , this is a use case where we do expect and want duplicates. This is the only List usage we have I believe.
image

IMO the TODO was on the wrong side and we probably should not merge this. How about we just remove the TODO? :smile: 

Codex review:
> ### ⚠️ Ordering considerations (behavior change)
> The new behavior changes ordering for any value that already exists in the list; it gets moved to the tail. That is a meaningful semantic change and should be confirmed as acceptable. If list order is significant (e.g., chronological or priority ordering), this change could be surprising.
> 
> ### ⚠️ Partial update visibility / atomicity
> `LREM` + `RPUSH` in a pipeline is not atomic. Other clients can observe transient states where the item has been removed but not yet re-added. If atomicity matters, this should be done inside a Lua script or a MULTI/EXEC transaction. A pipeline improves throughput, but does not provide isolation.
> 
> ### ⚠️ Duplicate handling within a single batch
> If the same value appears multiple times in the input (e.g., due to a caller bug), the current described approach will still append duplicates. If true de-duplication is required, the input list should be de-duplicated before issuing Redis commands.
> 
> ## Performance Review
> - `LREM` is O(N) per value. If these lists are large or if a large batch is appended, total cost can become O(N*M). Depending on dataset size, this might be acceptable but should be acknowledged.
> - Pipelining mitigates round-trip cost but doesn’t change algorithmic cost on Redis.
> 
> ## Tests / Validation
> No tests were described. This change affects persistence semantics and ordering, so targeted tests would be valuable:
> - Re-indexing does not create duplicates.
> - Existing values move to the end (if that’s intended).
> - No duplicate values remain after a batch insert.
> - Concurrency scenario test (if possible), or at least documentation of atomicity limits.
> 
> ## Merge Readiness
> **Not yet ready.** The deduplication behavior is good, but I recommend clarifying and/or addressing the following before merge:
> 1. Confirm that “move-to-end” semantics are desired.
> 2. Address atomicity, or explicitly document that transient removal is acceptable.
> 3. Add tests to lock in the new behavior and prevent regressions.

ok300:
Looking at the review comments above, specifically

> In fact I see how we are using the List type for `/events` , this is a use case where we do expect and want duplicates.

and https://github.com/pubky/pubky-nexus/pull/691#discussion_r2703232304

it occurs to me this is not really a good change, so I will close it.

There are valid reasons for having duplicates in a list. If a caller wants no duplicates, they can use sets.

PR #699, opened 2026-01-22
chore: same test order in readme and ci

aintnostressin:
This change reflects test order defined in github actions. 

`nexus-webapi` tests share the same  state as `nexus-watcher` tests, if env is not cleaned up between runs then following errors appear:

```
  stderr ───

    thread 'tags::hot::test_global_hot_tags_with_today_timeframe' (953172) panicked at nexus-webapi/tests/tags/hot.rs:52:5:
    assertion `left == right` failed
      left: Number(4)
     right: 3
```

```
  stderr ───

    thread 'stream::user::influencers::test_global_influencers_with_today_timeframe' (952936) panicked at nexus-webapi/tests/stream/user/influencers.rs:143:9:
    Expected user ID not found: omynbjw4ksjc4at5gretyoatw1g5h53tkee5z55fh69sng1d3jpy
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

This state dependency doesn't cause problems in current CI setup (yet). 
This PR doesn't fix underlying problem and it should be corrected in separate PR if needed.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -76,10 +76,10 @@ jobs:
       - name: Run nexus-common unit tests
         run: cargo nextest run -p nexus-common --no-fail-fast
 
-      # This test suite uses the TEST_PUBKY_CONNECTION_STRING env variable
       - name: Run API integration tests
         run: cargo nextest run -p nexus-webapi --no-fail-fast
 
+      # This test suite uses the TEST_PUBKY_CONNECTION_STRING env variable
       - name: Run WATCHER integration tests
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -76,10 +76,10 @@ jobs:
       - name: Run nexus-common unit tests
         run: cargo nextest run -p nexus-common --no-fail-fast
 
+      # This test suite uses the TEST_PUBKY_CONNECTION_STRING env variable
```
ok300: Are you sure? For me, the `nexus-webapi` tests run fine locally even without postgres running at all.

Issue #701, opened 2026-01-23
In `Bootstrap` payload, muted users should only be reflected as user IDs


ok300:
#650 added initial support for muted users in the `Bootstrap` response.

However, the muted users are added both to
* `bootstrap.ids.muted` , which stores user IDs of muted users for the given viewer
* `bootstrap.users` , which stores the full user details (`UserView`)

Please update the bootstrap API handler to _not_ include the full details of muted users in the response.

PR #710, opened 2026-01-30
fix: exclude muted users from bootstrap user views

aintnostressin:
Muted users now appear only as ids.

closes: #701

### DIFF
--- a/nexus-webapi/tests/user/bootstrap.rs
+++ b/nexus-webapi/tests/user/bootstrap.rs
@@ -1,5 +1,3 @@
-//use std::collections::HashSet;
-
 use std::collections::HashSet;
 
 use crate::utils::get_request;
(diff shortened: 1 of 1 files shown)


nexus-common/src/models/bootstrap.rs
```diff
@@ -296,18 +298,12 @@ impl Bootstrap {
...
     }
 
-    async fn add_muted(
-        &mut self,
-        user_ids: &mut HashSet<String>,
-        maybe_viewer_id: Option<&str>,
-    ) -> Result<(), DynError> {
+    async fn add_muted(&mut self, maybe_viewer_id: Option<&str>) -> Result<(), DynError> {
         if let Some(viewer_id) = maybe_viewer_id {
             if let Ok(Some(muted_ids)) = Muted::get_by_id(viewer_id, None, None).await {
```
aintnostressin: shall we limit it ?

PR #720, opened 2026-02-11
feat: add optional attachment metadata to post stream api

aintnostressin:
Add `include_attachment_metadata` flag to post stream endpoints, similar to #707

### DIFF
--- a/nexus-webapi/src/models/post.rs
+++ b/nexus-webapi/src/models/post.rs
@@ -75,8 +75,7 @@ impl PostStreamDetailed {
             .cloned()
             .collect();
 
-        // Single batched fetch, then build a lookup by database key (owner_id, file_id)
-        // This is more robust than URI string matching as it's immune to normalization differences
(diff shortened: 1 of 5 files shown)


nexus-webapi/src/models/post.rs
```diff
@@ -12,6 +15,13 @@ pub struct PostViewDetailed {
 }
 
 impl PostViewDetailed {
+    pub fn from(view: PostView, attachments_metadata: Vec<FileDetails>) -> Self {
+        Self {
+            view,
+            attachments_metadata,
+        }
+    }
```
tipogi: Nit: rename `PostViewDetailed::from(...)` to `PostViewDetailed::new(...)`.
In Rust, `from` is strongly associated with the `From` trait — a standard single-argument conversion. `new` is the standard name for "build me one of these," carries no baggage, and makes the intent immediately clear.

tipogi (commented):
apart of that, created a new [PR](https://github.com/pubky/pubky-nexus/pull/728) that might be useful to merge in that PR

Issue #714, opened 2026-02-04
Recommended users cache key mismatch


aintnostressin:
There is a cache key mismatch for recommended users set. 

key used for populating cache: `Cache:Users:Recommended:{user_id}`
key used for retrieving from cache: `Cache:Users:Recommended:Cache:Recommended:{user_id}`

Side effects: 
* application always returns up-to-date recommended users list
* cache is populated/checked without any real benefit but with additional overhead

We either **fix keys** or **remove cache**.

If we decide to fix keys, we will also need to implement proper cache eviction logic (for example upon user removal) to satisfy unit tests like `test_delete_recommended_user`

PR #722, opened 2026-02-13
fix: recommended users cache 

aintnostressin:
Fixes recommended users cache keys mismatch so redis cache will be used instead of neo4j if exists.                                                                                                                                                   
                  
  * fixes cache keys mismatch
  * filters out deleted users from recommended users set (extra redis hit)
  * self-heals cache by removing deleted users via SREM

fixes: #714

### DIFF
--- a/nexus-watcher/tests/event_processor/users/del_with_relations.rs
+++ b/nexus-watcher/tests/event_processor/users/del_with_relations.rs
@@ -4,7 +4,7 @@ use crate::{
 };
 use anyhow::Result;
 use chrono::Utc;
-use nexus_common::models::user::{UserCounts, UserStream, UserView};
+use nexus_common::models::user::{UserCounts, UserStream, UserView, USER_DELETED_SENTINEL};
 use pubky::Keypair;
 use pubky_app_specs::{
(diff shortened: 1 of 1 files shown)


nexus-common/src/models/user/mod.rs
```diff
@@ -23,3 +23,8 @@ pub use stream::{
 pub use tags::ProfileTag;
 pub use tags::UserTags;
 pub use view::UserView;
+
+/// Sentinel value used to mark deleted users in the system.
+/// When a user with relationships is deleted, their name field is set to this value
+/// instead of fully removing their profile data.
+pub const USER_DELETED_SENTINEL: &str = "[DELETED]";
```
ok300: NIT: since you introduced this constant, it would make sense to propagate its use in other places that hardcode "[DELETED]" for a deleted user:
- `test_delete_user_with_relationships`, when it asserts
  - `user_details.name, "[DELETED]"` in 2 places
  - `user_view.details.name, "[DELETED]"` in 2 places

ok300 (approved):
Looks good! Added just one comment.

PR #723, opened 2026-02-13
feat(bootstrap): remove muted user ids

aintnostressin:


nexus-common/src/models/bootstrap.rs
```diff
@@ -47,7 +46,11 @@ pub struct BootstrapIds {
     /// Recommended users for the given user ID
     pub recommended: Vec<String>,
     pub hot_tags: Vec<HotTag>,
-    /// User IDs muted by the given user
+    /// DEPRECATED: Always empty. Muted users are now managed client-side.
```
ok300: @tipogi should the field in the bootstrap response `Bootstrap.ids.muted` be removed or only deprecated?

PR #727, opened 2026-02-16
feat(bootstrap): remove taggers user views

aintnostressin:


nexus-common/src/models/bootstrap.rs
```diff
@@ -327,23 +280,13 @@ impl Bootstrap {
...
     ///
-    /// # Parameters
-    /// - `user_ids: &mut HashSet<String>` A mutable reference to a set of user IDs
-    ///
-    async fn add_global_hot_tags(
-        &mut self,
-        user_ids: &mut HashSet<String>,
-    ) -> Result<(), DynError> {
+    async fn add_global_hot_tags(&mut self) -> Result<(), DynError> {
         let hot_tag_filter = HotTagsInputDTO::new(Timeframe::Today, 40, 0, 20, None);
```
tipogi: in the UI, we just show max 5 tags and taggers limit should be also 5. In that case, IMO we could add the `userView` for that taggers

tipogi (approved):

PR #728, opened 2026-02-17
refactor: make file_key_from_uri return Option instead of Vec

tipogi:
## Summary

- Changes `FileDetails::file_key_from_uri` return type from `Vec<String>` to
  `Option<(String, String)>`, replacing the silent fallback (returning a vec with
  an empty `file_id`) with an explicit `None` for malformed or non-file URIs.
- Updates all call sites across the codebase to handle the new `Option` return:
  - **`avatar.rs`**: replaces manual `keys.len() != 2` guard with
    `.ok_or(Error::InternalServerError {...})?`
  - **`file/details.rs`**: replaces unchecked `file_key[0]`, `file_key[1]`
    indexing with `.ok_or(Error::FileNotFound {})?`
  - **`file/list.rs`**: adds `filter_map` to skip invalid URIs with a warning,
    uses `[&str; 2]` instead of `Vec<&str>`
  - **`post.rs`**: simplifies the distribution lookup in `from_post_views` from a
    10-line `if file_key.len() == 2` block to
    `FileDetails::file_key_from_uri(uri)?; metadata_by_key.get(&key).cloned()`.
    Hardens `fetch_attachment_metadata` to filter out malformed URIs upfront with
    logged warnings, and uses fixed-size `[&str; 2]` arrays for DB key refs.

## Why

PR #720 introduced the post stream attachment metadata feature but still relied
on the old `Vec<String>` return from `file_key_from_uri`, which silently produced
bad keys for malformed URIs (e.g., a non-file resource URI would yield
`(user_id, "")` and pass through to the DB). This follow-up makes every failure
path explicit and impossible to misuse at the type level.

## Pre-submission Checklist

### DIFF
--- a/nexus-webapi/src/routes/v0/file/details.rs
+++ b/nexus-webapi/src/routes/v0/file/details.rs
@@ -18,6 +18,7 @@ use utoipa::OpenApi;
     ),
     responses(
         (status = 200, description = "File Details", body = FileDetails),
+        (status = 400, description = "Malformed file URI"),
         (status = 404, description = "File not found"),
         (status = 500, description = "Internal server error")
     )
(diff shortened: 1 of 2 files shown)


nexus-webapi/src/routes/v0/file/details.rs
```diff
@@ -25,8 +25,9 @@ use utoipa::OpenApi;
 pub async fn file_details_handler(Path(file_uri): Path<String>) -> Result<Json<FileDetails>> {
     debug!("GET {FILE_ROUTE} file_uri:{}", file_uri);
 
-    let file_key = FileDetails::file_key_from_uri(&file_uri);
-    let files = FileDetails::get_by_ids(&[&[&file_key[0], &file_key[1]]]).await?;
+    let (owner_id, file_id) =
+        FileDetails::file_key_from_uri(&file_uri).ok_or(Error::FileNotFound {})?;
```
aintnostressin: URI is not valid - maybe we should return `Bad Request` instead of `File not found` ?

ok300 (approved):

Issue #731, opened 2026-02-18
Deleted users should not appear in recommendations


aintnostressin:
* Influencers
* Recommended
* Most followed

these user streams should not contain users that are deleted.

PR #732, opened 2026-02-18
fix: remove deleted users from recommended sets

aintnostressin:
This PR applies the same strategy as we have for recommended users.

1. Filter out users which are marked as '[DELETED]' from result set.
2. If  deleted user is coming from cache - remove it from cache.

fixes: #731

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,8 @@ cargo run -p nexusd -- api
 
 5. **Access Redis and Neo4j UIs and Swagger endpoint**:
    - Swagger UI: [http://localhost:8080/swagger-ui](http://localhost:8080/swagger-ui)
-   - Redis Insight: [http://localhost:8001/redis-stack/browser](http://localhost:8001/redis-stack/browser)
+   - Redis Insight: [http://localhost:5540/0/browser](http://localhost:5540/0/browser)
+     - Note: on first run, an error popup is shown and a TOS popup. After you accept the TOS, the link will work.
    - Neo4J Browser: [http://localhost:7474/browser/](http://localhost:7474/browser/)
 
 ## 📈 Observability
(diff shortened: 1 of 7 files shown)


nexus-common/src/models/user/influencers.rs
```diff
@@ -99,8 +100,8 @@ impl Influencers {
         timeframe: &Timeframe,
     ) -> Result<Option<Influencers>, DynError> {
         let cached_influencers = Influencers::get_from_global_cache(skip, limit, timeframe).await?;
-        if cached_influencers.is_some() {
-            return Ok(cached_influencers);
+        if let Some(cached) = cached_influencers {
+            return Influencers::filter_deleted(cached, Some(timeframe)).await;
```
ok300: Wouldn't it make more sense to apply `filter_deleted` inide `Influencers::get_from_global_cache`?

Then `get_global_influencers` would essentially remain unchanged.

In addition, all future callers of `Influencers::get_from_global_cache` will auto-apply `filter_deleted`.

ok300 (commented):
It seems to me the last commit 33674bdcf381cac9e53734cdae0e9eaab2371ce1 has a subtle unintended consequence: by not using `filter_deleted` anymore, it also doesn't call `remove_deleted_from_global_cache` anymore.

Is that intended?

aintnostressin:
The test is failing because some users are not properly indexed due to #742 . 

[get_influencers](https://github.com/pubky/pubky-nexus/blob/main/nexus-common/src/models/user/influencers.rs#L62-L71) function is non detereministic thus making tests fail from time to time. 

The failure happens when graph query returns ids of users not present in redis - such users are skipped (deleted), causing `assert_eq!(user_bootstrap_respose.ids.influencers.len(), 3);` to fail.

PR #744, opened 2026-02-20
Fix: use correct content type when serving video

### LINKED ISSUE #743: Serve video media with correct Content-Type header
Webkit doesn't allow pubky-app users to view videos in posts. It seems this is due to the content type header `Content-Type: application/octet-stream` being used for `/static/files/...` response for partial chunks of content being retreived. We need nexus to instead use `Content-Type: video/mp4` or whatever the actual video format is. This has been proven in pubky-app by adding a proxy route to modify this response header.

See internal thread for details: https://synonymworkspace.slack.com/archives/C031F2DC1JM/p1771419070042259

ok300:
This PR ensures video files are served with the correct `Content-Type` header.

This used to not work because of an HTTP success check that was too strict: it only injected the video content type header for `200 OK`, however video streaming returns `206 Partial Content`. Extending this check with a more generic `response.status().is_success()` covers the streaming response code as well.

Fixes #743

### DIFF
--- a/nexus-common/src/media/processors/video.rs
+++ b/nexus-common/src/media/processors/video.rs
@@ -131,11 +131,4 @@ mod tests {
         let result = VideoProcessor::get_content_type_for_variant(&file, &FileVariant::Main);
         assert_eq!(result, "video/mp4");
     }
-
-    #[test]
-    fn test_main_variant_preserves_quicktime_content_type() {
-        let file = make_file("video/quicktime");
-        let result = VideoProcessor::get_content_type_for_variant(&file, &FileVariant::Main);
-        assert_eq!(result, "video/quicktime");
-    }
 }


nexus-common/src/media/processors/video.rs
```diff
@@ -103,3 +106,29 @@ impl VideoProcessor {
...
+    fn test_main_variant_preserves_original_content_type() {
+        let file = make_file("video/webm");
+        let result = VideoProcessor::get_content_type_for_variant(&file, &FileVariant::Main);
+        assert_eq!(result, "video/webm");
+    }
+
+    #[test]
+    fn test_main_variant_preserves_mp4_content_type() {
+        let file = make_file("video/mp4");
+        let result = VideoProcessor::get_content_type_for_variant(&file, &FileVariant::Main);
```
aintnostressin: Hmm both test cases are testing the same branch - should here be used other file variant ?

aintnostressin (approved):

PR #755, opened 2026-02-27
`nexus-watcher`: use minimal testnet

ok300:
This PR replaces `EphemeralTestnet`, which was initialized in every test, with a minimal variant of `pubky_testnet::Testnet`.


nexus-watcher/tests/event_processor/utils/watcher.rs
```diff
@@ -113,10 +113,9 @@ impl WatcherTest {
 
         // WARNING: testnet initialization is time expensive, we only init one per process
         // TODO: Maybe we should create a single testnet network (singleton and push there more homeservers)
-        let mut testnet = EphemeralTestnet::builder()
-            .with_http_relay()
-            .build()
-            .await?;
+        // This can be further sped up by using Testnet::new_unseeded() with pubky-testnet 0.7.x
+        let mut testnet = pubky_testnet::Testnet::new().await?;
```
aintnostressin: `pubky_testnet::Testnet::new().await?;` -> Testnet::new().await?;

PR #788, opened 2026-04-02
`TEventProcessor`: expand logic of `handle_error`

ok300:
This PR more explicitly implements the logic of `TEventProcessor::handle_error`, making it re-usable across most implementations of the trait.

The `RetryManager` will have to overwrite it, but the two other implementations can just re-use this default impl.

This PR also removed the older / deprecated `handle_error` implementation from `HsEventProcessor`.


tipogi (approved):

nexus-common/src/models/event/errors.rs
```diff
@@ -116,4 +116,19 @@ impl EventProcessorError {
...
         Self::Generic(source.to_string())
     }
+
+    /// Returns whether or not this is an infrastructure error.
+    ///
+    /// If true, the caller has to stop indexing and persist the cursor
+    /// at the point where this error occurred. This is because, since
+    /// it's an infrastructure error, it is likely that trying to process
+    /// the next event line would result in the same issue.
+    pub fn is_infrastructure(&self) -> bool {
```
aintnostressin, line 130: should we call it `transient` ? 

`InternalError` doesn't fit into infrastructure scope.

nexus-common/src/models/event/errors.rs
```diff
@@ -116,4 +116,19 @@ impl EventProcessorError {
     pub fn generic(source: impl std::fmt::Display) -> Self {
         Self::Generic(source.to_string())
     }
+
+    /// Returns whether or not this is an infrastructure error.
+    ///
+    /// If true, the caller has to stop indexing and persist the cursor
```
aintnostressin: This is just suggestion - personally I would avoid making such claims here - it's up to implementation to handle it.

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -129,15 +129,30 @@ pub trait TEventProcessor: Send + Sync + 'static {
...
-    async fn handle_error(&self, _event: &Event, _error: EventProcessorError) {}
+    /// Called in the event processing loop.
+    ///
+    /// It re-throws the erorr, if the caller should break the processing loop.
+    async fn handle_error(
+        &self,
+        _event: &Event,
+        error: EventProcessorError,
+    ) -> Result<(), EventProcessorError> {
+        if matches!(error, EventProcessorError::MissingDependency { .. }) {
```
aintnostressin, line 133: Nice! This is aligns with my plan.

aintnostressin (approved):

nexus-common/src/models/event/errors.rs
ok300: Done, removed `InternalError`: e02991874724db44bd67c8348d192db19ab2a52f

tipogi (approved):

aintnostressin (approved):

Ari4ka (approved):

PR #791, opened 2026-04-03
fix: remove stale `HOSTED_BY` edge if no HS is found for user

ok300:
This PR removes the `HOSTED_BY` edge, if the HS resolver task founds the user has no published HS mapping.


tipogi (approved):

aintnostressin (approved):
Good change 👍 

I'm only wodnering if we should keep relation as long as previously found record is valid (timestamp + ttl). 

So even if it's not on DHT - it's still valid (was signed).

ok300:
As discussed, there is already a `TTL` for the `HOSTED_BY` edge.

PR #793, opened 2026-04-03
feat: bookmarks events handler partial recovery

aintnostressin:
Prepares bookmark handlers for partial recovery.

  - Reorders sync_del to read the bookmark target from the graph before deleting the edge, ensuring data survives for retries if Redis ops fail (graph-last pattern)                                                                                    
  - Guards counter decrement behind a Redis index existence check, preventing double-decrement on retry
  - Adds get_bookmark_target graph query and Bookmark::get_target_from_graph to read without side effects

### DIFF
--- a/nexus-watcher/src/events/handlers/bookmark.rs
+++ b/nexus-watcher/src/events/handlers/bookmark.rs
@@ -60,9 +60,11 @@ pub async fn del(user_id: PubkyId, bookmark_id: String) -> Result<(), EventProce
 
 pub async fn sync_del(user_id: PubkyId, bookmark_id: String) -> Result<(), EventProcessorError> {
     // 1. Read target from graph WITHOUT deleting the edge
-    let (post_id, author_id) = Bookmark::get_target_from_graph(&user_id, &bookmark_id)
-        .await?
-        .ok_or(EventProcessorError::SkipIndexing)?;
+    let Some((post_id, author_id)) =
+        Bookmark::get_target_from_graph(&user_id, &bookmark_id).await?
+    else {
+        return Ok(());
+    };
 
     // 2. Guard counter decrement: only decrement if bookmark still exists in Redis index
     let existed_in_index = Bookmark::get_from_index(&author_id, &post_id, &user_id)
--- a/nexus-watcher/src/events/handlers/bookmark.rs
+++ b/nexus-watcher/src/events/handlers/bookmark.rs
@@ -59,16 +59,27 @@ pub async fn del(user_id: PubkyId, bookmark_id: String) -> Result<(), EventProce
 }
 
 pub async fn sync_del(user_id: PubkyId, bookmark_id: String) -> Result<(), EventProcessorError> {
-    let deleted_bookmark_info = Bookmark::del_from_graph(&user_id, &bookmark_id).await?;
-    // Ensure the bookmark exists in the graph before proceeding
-    let (post_id, author_id) = match deleted_bookmark_info {
+    // 1. Read target from graph WITHOUT deleting the edge
+    let (post_id, author_id) = match Bookmark::get_target_from_graph(&user_id, &bookmark_id).await?
+    {
         Some(info) => info,
         None => return Err(EventProcessorError::SkipIndexing),
     };
 
+    // 2. Guard counter decrement: only decrement if bookmark still exists in Redis index
+    let existed_in_index = Bookmark::get_from_index(&author_id, &post_id, &user_id)
+        .await?
+        .is_some();
+
+    // 3. Redis cleanup (idempotent)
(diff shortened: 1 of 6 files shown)


nexus-watcher/src/events/handlers/bookmark.rs
```diff
@@ -59,16 +59,27 @@ pub async fn del(user_id: PubkyId, bookmark_id: String) -> Result<(), EventProce
...
 pub async fn sync_del(user_id: PubkyId, bookmark_id: String) -> Result<(), EventProcessorError> {
-    let deleted_bookmark_info = Bookmark::del_from_graph(&user_id, &bookmark_id).await?;
-    // Ensure the bookmark exists in the graph before proceeding
-    let (post_id, author_id) = match deleted_bookmark_info {
+    // 1. Read target from graph WITHOUT deleting the edge
+    let (post_id, author_id) = match Bookmark::get_target_from_graph(&user_id, &bookmark_id).await?
+    {
         Some(info) => info,
         None => return Err(EventProcessorError::SkipIndexing),
     };
```
ok300:
```suggestion
    let (post_id, author_id) = Bookmark::get_target_from_graph(&user_id, &bookmark_id)
        .await?
        .ok_or(EventProcessorError::SkipIndexing)?;
```

nexus-watcher/src/events/handlers/bookmark.rs
```diff
@@ -59,16 +59,27 @@ pub async fn del(user_id: PubkyId, bookmark_id: String) -> Result<(), EventProce
...
 
+    // 2. Guard counter decrement: only decrement if bookmark still exists in Redis index
```
ok300, line 69: The guard is designed to prevent double-decrement on retry, it it only handles the case where both the Redis index removal and the counter decrement succeed before the graph delete failed.

If `del_from_index` (step 3) succeeds, but the process crashes before `UserCounts::decrement` runs, then on retry `existed_in_index` will be `false` and the counter will never be decremented.

nexus-watcher/tests/event_processor/bookmarks/del_idempotent.rs
```diff
@@ -0,0 +1,151 @@
...
+    let bookmark_id = bookmark.create_id();
+    let bookmark_path = bookmark.hs_path();
+
+    test.put(&user_kp, &bookmark_path, bookmark).await?;
+
+    // Delete through normal event flow
+    test.del(&user_kp, &bookmark_path).await?;
+
+    // Verify fully deleted state
+    assert!(find_post_bookmark(&user_id, &post_id, &user_id)
```
ok300: The test conflates the post author and the bookmarker, which means it doesn't cover the case when post author != bookmarker, which is a more realistic scenario. It would be better if the test creates 2 separate users, `author_id` and `bookmarker_id`.

nexus-watcher/src/events/handlers/bookmark.rs
aintnostressin: That's true - 0 or 1 decrements.

nexus-watcher/tests/event_processor/bookmarks/del_idempotent.rs
aintnostressin: done [606a14a](https://github.com/pubky/pubky-nexus/pull/793/commits/606a14a019399755cf423c046faa152f03024f50)

nexus-watcher/src/events/handlers/bookmark.rs
aintnostressin: done [606a14a](https://github.com/pubky/pubky-nexus/pull/793/commits/606a14a019399755cf423c046faa152f03024f50)

nexus-watcher/src/events/handlers/bookmark.rs
```diff
@@ -59,16 +59,25 @@ pub async fn del(user_id: PubkyId, bookmark_id: String) -> Result<(), EventProce
...
-    let deleted_bookmark_info = Bookmark::del_from_graph(&user_id, &bookmark_id).await?;
-    // Ensure the bookmark exists in the graph before proceeding
-    let (post_id, author_id) = match deleted_bookmark_info {
-        Some(info) => info,
-        None => return Err(EventProcessorError::SkipIndexing),
-    };
+    // 1. Read target from graph WITHOUT deleting the edge
+    let (post_id, author_id) = Bookmark::get_target_from_graph(&user_id, &bookmark_id)
+        .await?
+        .ok_or(EventProcessorError::SkipIndexing)?;
```
tipogi: I guess with that error type, we might be careful because in the actual impl. `RetryManager` will ingest again. Maybe you are planing something for the future `RetryManager`

tipogi (approved):

ok300 (approved):

PR #798, opened 2026-04-06
chore: add `TEventProcessor::instance_name()`

ok300:
This PR addresses TODOs pointing to the need to define a way to get a `TEventProcessor` instance's name.


tipogi (approved):
Nice

PR #800, opened 2026-04-06
feat: user-key-based event indexing

tipogi:
## Pre-submission Checklist

## Description

Implements per-user event indexing for non-default homeservers via `KeyBasedEventProcessor`, which connects to external homeservers using the pubky SDK's event stream API.

## Changes

- **`indexer/key_based.rs`** — Full `run_internal` implementation that:
  - Resolves monitored users on the homeserver and reads their cursors from Redis
  - Opens a **per-user** event stream (one user per subscription) with bounded execution via `.limit()`
  - Checkpoints cursors to Redis after each user's event loop completes (best-effort)
  - Graceful shutdown via `shutdown_rx` (checked between users and inside each event loop)
  - Propagates Redis cursor-read errors instead of silently rewinding to 0
  - Tracing instrumentation (`#[instrument]`) on `resolve_users_with_cursors` and `process_user`
  - Infrastructure errors abort the homeserver run; non-infrastructure errors skip the user

- **`nexus-common/src/models/event/mod.rs`** — `Event::from_stream_event` constructor that maps the SDK's `StreamEvent` directly to the internal `Event` struct, bypassing the string round-trip through `parse_event`

- **`runner/key_based.rs`** — `KeyBasedEventProcessorRunner` wires config (`limit`, `shutdown_rx`, `moderation`, `files_path`, backoff) and builds `KeyBasedEventProcessor` instances per homeserver

- **`nexus-watcher/Cargo.toml`** — `futures` dependency (workspace) for `StreamExt`


nexus-common/src/models/event/mod.rs
```diff
@@ -94,6 +95,45 @@ impl Event {
...
+            Resource::Unknown => {
+                return Err(EventProcessorError::InvalidEventLine(format!(
+                    "Unknown resource in URI: {uri}"
+                )))
+            }
+            Resource::LastRead | Resource::Feed(_) | Resource::Blob(_) => return Ok(None),
+            _ => (),
+        };
+
+        let event_line = format!("{event_type} {uri}");
```
aintnostressin: Can we have a 'line' as it was returned by the homeserver instead of reconstructing it ?

nexus-watcher/src/service/runner/key_based.rs
```diff
@@ -78,8 +78,10 @@ impl TEventProcessorRunner for KeyBasedEventProcessorRunner {
 
         Ok(Arc::new(KeyBasedEventProcessor {
             homeserver,
+            limit: self.limit.try_into().unwrap_or(u16::MAX),
```
aintnostressin, line 81: Is `u16::MAX` here safe to use? 65k is a lot of events in a single response

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -29,13 +43,154 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
+        &self,
+        hs_id: &str,
+    ) -> Result<Vec<(PublicKey, EventCursor)>, EventProcessorError> {
+        let user_ids = user_hs_resolver::get_user_ids_by_homeserver(hs_id).await?;
+        debug!("Resolved {} user(s) on HS {hs_id}", user_ids.len());
+
+        let mut users = Vec::with_capacity(user_ids.len());
+        for user_id in &user_ids {
+            let user_pk: PublicKey = user_id.parse().map_err(|_| {
+                EventProcessorError::client_error(format!("Invalid user public key: {user_id}"))
```
aintnostressin: Should we only log this ?
This will block monitoring other users from this HS if one pk is invalid.

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -29,13 +43,154 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
+            })?;
+            let cursor = Self::read_user_cursor(user_id, hs_id).await?;
```
aintnostressin, line 111: Same here - maybe only error/warn and let it proceed with other users ?

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -29,13 +43,154 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
+        Ok(())
+    }
 
+    /// Flushes all accumulated in-memory cursors to Redis in one pass.
+    async fn flush_cursors(
+        cursors: &HashMap<String, u64>,
+        hs_id: &str,
+    ) -> Result<(), EventProcessorError> {
+        for (user_id, cursor) in cursors {
+            Self::write_user_cursor(user_id, hs_id, *cursor).await?;
```
aintnostressin: I think we should try to write as many cursors as possible instead of failing on first error.

nexus-common/src/models/event/mod.rs
tipogi: we cannot do that one because event `list` and event `stream` endpoints return different result type. For our case, we get that output for each event
```txt
event: PUT
data: pubky://<pk>/pub/...
data: cursor: 42
data: content_hash: <base64>
```

nexus-watcher/src/service/runner/key_based.rs
tipogi: I think the code never will reach that one becase `WatcherConfig` default configuration will add the default value. But now that you raise that point, we need another limit that is different from the default HS because key based indexing 50 events might be enough as `limit` while in the default one should be bigger, 500 for example

nexus-watcher/src/service/indexer/key_based.rs
tipogi: 02a41d9e

nexus-watcher/src/service/indexer/key_based.rs
tipogi: deprecated after c7546c3

nexus-watcher/src/service/indexer/key_based.rs
tipogi: deprecated after c7546c3

nexus-common/src/models/user/details.rs
```diff
@@ -14,6 +14,12 @@ use utoipa::ToSchema;
 
 pub const USER_HS_CURSOR: [&str; 2] = ["Users", "Homeservers"];
 
+/// Builds the Redis key path for per-user homeserver cursor storage:
+/// `["Users", "Homeservers", <user_id>]`.
+pub fn user_hs_cursor_key(user_id: &str) -> Vec<&str> {
```
ok300, line 19: Doesn't the per-user-cursor have to also be stored per-HS?

If a user has cursor `X` on HS1 and switches to HS2, they will have a new cursor on HS2 (`0`). When switching back to HS1, the user's cursor will now be `X`.

ok300: Nevermind, it's a sorted set where the key is the HS ID, so it is already also per-HS.

nexus-watcher/src/service/runner/key_based.rs
tipogi: https://github.com/pubky/pubky-nexus/pull/801

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -39,13 +91,195 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
+        let result: Result<(), EventProcessorError> = async {
+            for stream_event in stream_events {
+                if *self.shutdown_rx.borrow() {
+                    debug!(hs_id = %hs_id, user = %user_id, "Shutdown detected; exiting event loop");
+                    break;
+                }
+
+                let cursor_id = stream_event.cursor.id();
+
+                match Event::from_stream_event(&stream_event, self.files_path.clone()) {
```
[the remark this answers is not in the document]
ok300: Clarified behavior in 902e0a82e90b825a2bd6b5906b5ec156981f0d87

nexus-common/src/models/event/mod.rs
```diff
@@ -95,40 +103,61 @@ impl Event {
...
+        let event_type: EventType = stream_event.event_type.clone().into();
+
+        let uri = stream_event.resource.to_pubky_url();
+        debug!("New stream event: {event_type} {uri}");
+
+        let event_line = format!("{event_type} {uri}");
+        match Self::parse_event_parts(event_type, uri, event_line, files_path)? {
+            ParseResult::Parsed(event) => Ok(Some(event)),
+            ParseResult::Skipped | ParseResult::UnrecognizedUri { .. } => Ok(None),
+        }
```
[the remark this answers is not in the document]
ok300: Logged handling of `UnrecognizedUri` in 609c16dde3037dedb6e7688bf39479e6e0b77c61

nexus-common/src/config/watcher.rs
```diff
@@ -175,6 +192,37 @@ fn default_hs_resolver_sleep() -> u64 {
...
+    D: Deserializer<'de>,
+{
+    let val = u16::deserialize(deserializer)?;
+    if val > MAX_EVENTS_LIMIT {
+        let err_msg = format!("events_limit ({val}) exceeds max ({MAX_EVENTS_LIMIT})");
+        Err(D::Error::custom(err_msg))
+    } else {
+        Ok(val)
+    }
+}
```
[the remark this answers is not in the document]
ok300: Done: ee91d6f61af460b7594780ca7350d7b73a08ced9

ok300:
@greptile please review again

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -1,23 +1,75 @@
...
+    ) -> Result<Vec<StreamEvent>, EventProcessorError> {
+        let pubky = PubkyConnector::get()?;
+        let mut stream = pubky
+            .event_stream_for(hs_pk)
+            .add_users(vec![(user_pk, Some(cursor))])?
+            .limit(limit)
+            .path("/pub/")
+            .subscribe()
+            .await
+            .inspect_err(|e| error!("Failed to subscribe to event stream: {e:?}"))?;
```
aintnostressin, line 53: Have you checked how long this will take ? It subscribes so from my understanding this will wait for new events to come.

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -39,13 +91,213 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
+            .inspect_err(|e| error!("Failed to resolve users for HS {hs_id}: {e:?}"))?;
+
+        if users.is_empty() {
+            debug!("No users on HS {hs_id}, skipping");
+            return Ok(());
+        }
+
+        info!("Found {} users on HS {hs_id}", users.len());
+
+        // TODO: Process users concurrently (bounded semaphore) to reduce per-HS latency
```
aintnostressin: Please resolve this todo

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -39,13 +91,213 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
+        hs_id: &str,
+    ) -> Result<Vec<(PublicKey, EventCursor)>, EventProcessorError> {
+        let user_ids = user_hs_resolver::get_user_ids_by_homeserver(hs_id).await?;
+        debug!("Resolved {} user(s) on HS {hs_id}", user_ids.len());
 
-        // TODO Implement: fetch events per user
+        let mut users = Vec::with_capacity(user_ids.len());
+        for user_id in &user_ids {
+            let Ok(user_pk) = user_id.parse::<PublicKey>() else {
+                error!("Invalid user public key '{user_id}' on HS {hs_id}, skipping");
```
aintnostressin: We could have a metric + alert on that so data can be repaired.

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -39,13 +91,213 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
+                continue;
+            };
+            let cursor = Self::read_user_cursor(user_id, hs_id).await?;
```
aintnostressin: If we load users with one query then maybe it's also possible to fetch cursors in one query ?

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -39,13 +91,213 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
+            .await?;
+
+        let user_id = user_pk.z32();
+        let (latest_cursor, result) = self
+            .process_user_events(hs_id, &user_id, stream_events)
+            .await;
+
+        if let Some(cursor_val) = latest_cursor {
+            if let Err(write_err) = Self::write_user_cursor(&user_id, hs_id, cursor_val).await {
+                // TODO: Queue failed cursor writes in the retry manager so they
```
aintnostressin: Please resolve this todo

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -1,23 +1,75 @@
...
+        let mut stream = pubky
+            .event_stream_for(hs_pk)
+            .add_users(vec![(user_pk, Some(cursor))])?
+            .limit(limit)
+            .path("/pub/")
+            .subscribe()
+            .await
+            .inspect_err(|e| error!("Failed to subscribe to event stream: {e:?}"))?;
+
+        let mut events = Vec::new();
```
aintnostressin, line 55: You can use `with_capacity` to pre-allocate.

aintnostressin (changes requested):

nexus-watcher/src/service/indexer/key_based.rs
ok300: `EventStreamBuilder::subscribe()` is an inaccurate method name. It is backward compatible with the previous client, which did an HTTP GET and just returned the event lines.

See the rustdoc of `EventStreamBuilder::live()`:

https://github.com/pubky/pubky-core/blob/88d194ffd2365c9f519a43c123da236d16d8dddd/pubky-sdk/src/actors/event_stream.rs#L249-L267

> Without this flag (default): Stream only delivers historical events and closes.

In other words, if `subscribe()` is used without `live()` (which is our case here) it will only perform an HTTP GET and close.

ok300: Added clarifying comments: 4e86c005297f482e578bc797958367cf75b63bf3

nexus-watcher/src/service/indexer/key_based.rs
ok300: This is a parallelization suggestion for the future, when there will be many users sharing a 3rd party HS:

```
// TODO: Process users concurrently (bounded semaphore) to reduce per-HS latency
//       when many users share a non-default homeserver.
```

Now it's not the case and sequential handling is good enough.

nexus-watcher/src/service/indexer/key_based.rs
aintnostressin: Or even better   `let events = stream.try_collect::<Vec<_>>().await?;`

nexus-watcher/src/service/indexer/key_based.rs
ok300: I think that's overkill.

Can be added later, if we really think its a high risk.

nexus-watcher/src/service/indexer/key_based.rs
ok300: Theoretically yes, but IMO that would be premature. This is only for users on external HSs, of which there is maybe 1 per HS or very few.

Added TODO to batch call this in the future: 5c1cd72617937a06dd0e9175edebd52e855a0498

nexus-watcher/src/service/indexer/key_based.rs
ok300: This was added before Retry Manager was ready.

What is the correct way to track this as a `RetryEvent`?

nexus-watcher/src/service/indexer/key_based.rs
ok300: Unfortunately the stream length isn't available before the stream is fully consumed with `stream.next()` calls.

nexus-watcher/src/service/indexer/key_based.rs
aintnostressin: How would we know that's a right time to improve it ?
Do we have any metrics for example?

nexus-watcher/src/service/indexer/key_based.rs
aintnostressin: If that's not a problem then please change error -> warn

nexus-watcher/src/service/indexer/key_based.rs
aintnostressin: I see no proper way of resolving this with `RetryEvent`.
We can either:
1. Stop here , add inline retry with exponential backoff - do not continue unless succeded
2. Accept failure and continue (remove todo)

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -1,23 +1,78 @@
...
+        hs_pk: &PublicKey,
+        user_pk: &PublicKey,
+        cursor: EventCursor,
+        limit: u16,
+    ) -> Result<Vec<StreamEvent>, EventProcessorError> {
+        let pubky = PubkyConnector::get()?;
+
+        // We are building the stream without the live flag, so it performs an HTTP GET and closes.
+        // See rustdoc of EventStreamBuilder::live()
+        let mut stream = pubky
```
aintnostressin, line 46: I'm running locally against prod homeserver.

After a couple of thousand events indexed it detected references to users on staging home server. 
It started to index those users but at least two problems are found during this process

1. If homeserver responds with http status code 404, nexus will keep on trying to fetch events for that user (but will fail again) - we should add some backoff
2. After a 5 requests (all failed with 404) sixth request fails with 429 Too Many Request which effectively blocks other users from being indexed


```
2026-05-21T08:30:06.059435Z  INFO event_processor.run:dx.user_events.process: pubky::actors::event_stream:362: Subscribing to event stream for 1 user(s) on homeserver pubkyufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=uudfeafc1c6dhxxnaiyuzsy3yy79i1ikpb8syht46qpnx4ksi6ho
2026-05-21T08:30:06.120291Z ERROR event_processor.run:dx.user_events.process: nexus_watcher::service::indexer::key_based:51: Failed to subscribe to event stream: Request(Server { status: 404, message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=uudfeafc1c6dhxxnaiyuzsy3yy79i1ikpb8syht46qpnx4ksi6ho
2026-05-21T08:30:06.120368Z ERROR event_processor.run: nexus_watcher::service::indexer::key_based:136: Non-infrastructure user error; continuing with next user hs_id=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=uudfeafc1c6dhxxnaiyuzsy3yy79i1ikpb8syht46qpnx4ksi6ho action="skip_user" error=PubkyClientError(NotFound404 { message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy
2026-05-21T08:30:06.120408Z  INFO event_processor.run:dx.user_events.process: pubky::actors::event_stream:362: Subscribing to event stream for 1 user(s) on homeserver pubkyufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=pxnu33x7jtpx9ar1ytsi4yxbp6a5o36gwhffs8zoxmbuptici1jy
2026-05-21T08:30:06.178419Z ERROR event_processor.run:dx.user_events.process: nexus_watcher::service::indexer::key_based:51: Failed to subscribe to event stream: Request(Server { status: 404, message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=pxnu33x7jtpx9ar1ytsi4yxbp6a5o36gwhffs8zoxmbuptici1jy
2026-05-21T08:30:06.178494Z ERROR event_processor.run: nexus_watcher::service::indexer::key_based:136: Non-infrastructure user error; continuing with next user hs_id=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=pxnu33x7jtpx9ar1ytsi4yxbp6a5o36gwhffs8zoxmbuptici1jy action="skip_user" error=PubkyClientError(NotFound404 { message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy
2026-05-21T08:30:06.178533Z  INFO event_processor.run:dx.user_events.process: pubky::actors::event_stream:362: Subscribing to event stream for 1 user(s) on homeserver pubkyufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=cn7sk399cyfotgyqez3mekcwwktgr3rynyjxp9ci6s63iw6kk7go
2026-05-21T08:30:06.230238Z ERROR event_processor.run:dx.user_events.process: nexus_watcher::service::indexer::key_based:51: Failed to subscribe to event stream: Request(Server { status: 404, message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=cn7sk399cyfotgyqez3mekcwwktgr3rynyjxp9ci6s63iw6kk7go
2026-05-21T08:30:06.230311Z ERROR event_processor.run: nexus_watcher::service::indexer::key_based:136: Non-infrastructure user error; continuing with next user hs_id=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=cn7sk399cyfotgyqez3mekcwwktgr3rynyjxp9ci6s63iw6kk7go action="skip_user" error=PubkyClientError(NotFound404 { message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy
2026-05-21T08:30:06.230357Z  INFO event_processor.run:dx.user_events.process: pubky::actors::event_stream:362: Subscribing to event stream for 1 user(s) on homeserver pubkyufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=9arfi37owcrdywc9zqw3m5uc7gd5gqu1yfuykzo66od6tcayqk9y
2026-05-21T08:30:06.287821Z ERROR event_processor.run:dx.user_events.process: nexus_watcher::service::indexer::key_based:51: Failed to subscribe to event stream: Request(Server { status: 404, message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=9arfi37owcrdywc9zqw3m5uc7gd5gqu1yfuykzo66od6tcayqk9y
2026-05-21T08:30:06.287910Z ERROR event_processor.run: nexus_watcher::service::indexer::key_based:136: Non-infrastructure user error; continuing with next user hs_id=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=9arfi37owcrdywc9zqw3m5uc7gd5gqu1yfuykzo66od6tcayqk9y action="skip_user" error=PubkyClientError(NotFound404 { message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy
2026-05-21T08:30:06.287968Z  INFO event_processor.run:dx.user_events.process: pubky::actors::event_stream:362: Subscribing to event stream for 1 user(s) on homeserver pubkyufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=sazaz5yowshbkfbsqi4n47rbyn8xbmokmu8kft59iyjazjxo4jeo
2026-05-21T08:30:06.348674Z ERROR event_processor.run:dx.user_events.process: nexus_watcher::service::indexer::key_based:51: Failed to subscribe to event stream: Request(Server { status: 404, message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=sazaz5yowshbkfbsqi4n47rbyn8xbmokmu8kft59iyjazjxo4jeo
2026-05-21T08:30:06.348774Z ERROR event_processor.run: nexus_watcher::service::indexer::key_based:136: Non-infrastructure user error; continuing with next user hs_id=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=sazaz5yowshbkfbsqi4n47rbyn8xbmokmu8kft59iyjazjxo4jeo action="skip_user" error=PubkyClientError(NotFound404 { message: "Event stream request failed with status 404 Not Found" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy
2026-05-21T08:30:06.348841Z  INFO event_processor.run:dx.user_events.process: pubky::actors::event_stream:362: Subscribing to event stream for 1 user(s) on homeserver pubkyufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=uxni6dn45bbnd7mw6ypf3swoyey9wjntmjo4h1ph9xab1jfhp8do
2026-05-21T08:30:06.398284Z ERROR event_processor.run:dx.user_events.process: nexus_watcher::service::indexer::key_based:51: Failed to subscribe to event stream: Request(Server { status: 429, message: "Event stream request failed with status 429 Too Many Requests" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy homeserver=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=uxni6dn45bbnd7mw6ypf3swoyey9wjntmjo4h1ph9xab1jfhp8do
2026-05-21T08:30:06.398405Z ERROR event_processor.run: nexus_watcher::service::indexer::key_based:136: Non-infrastructure user error; continuing with next user hs_id=ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy user=uxni6dn45bbnd7mw6ypf3swoyey9wjntmjo4h1ph9xab1jfhp8do action="skip_user" error=PubkyClientError(RequestFailed { is_transport: false, message: "Event stream request failed with status 429 Too Many Requests" }) service=KeyBasedEventProcessor with HS ID: ufibwbmed6jeq9k4p583go95wofakh9fwpp4k734trq79pd9u1uy
```

ok300: > If homeserver responds with http status code 404, nexus will keep on trying to fetch events for that user (but will fail again) - we should add some backoff

What 404 usually means is that, considering (A) the high-level view of all events for that particular resource:

`PUT_1`, `PUT_2`, ..., `PUT_n`, `DEL_(n+1)`

and considering (B) we only fetch and process parts of these events, paginating with `cursor` and `limit`:

- Page 1: `PUT_1`
- Page 2: `PUT_2`, `PUT_3`
- ...
- Page x: ... `DEL_(n+1)`

this means that (C) all "event pages" up to Page x (the one with the `DEL` event) will face a 404 error when the event processor tries to fetch and parse the event resource, because it doesn't exist.

So the correct way to handle 404 is to not retry them, not add any special handling (backoff, etc) but just log that it's a 404 and let the cursor advance.

ok300: > After a 5 requests (all failed with 404) sixth request fails with 429 Too Many Request which effectively blocks other users from being indexed

This could use a simple, light backoff.

The tricky part here is that different external HSs can be configured differently, so we have to dynamically adjust to the 429.

I think a good solution is that inside `KeyBasedEventProcessor` , the `event_source` needs a way to handle delays between the HS API requests. Could have a new inner field `hs_api_delay_sec` and some interface methods:

```
fn get_delay_between_api_reqs();
fn inc_delay_between_api_reqs(); // which would increment by 1s
```

Then the `event_source` could could `sleep` for this delay just before calling the HS API inside `fetch_events`.

Then finally, `KeyBasedEventProcessor` could react to the 429 error, for example inside the user loop. Something like so

```
for (user_pk, cursor) in &users {
  ...
   if let Err(err) = self.process_user(..) {
    if err.is_429) {
      self.event_source.increment_delay_between_api_reqs();
    }
  }
  ...
}
```

What do you think?

aintnostressin: On the 404 - it comes from event-stream fetch, not resource fetch. This may indicate that user is no longer present on this homeserver - PKDNS may say something else but we should handle it instead of hammering it each run.

On the 429 - we already have a backoff logic - we just need to detect 429 and throw error.  Wouldn't that be simpler ?

nexus-watcher/src/service/indexer/key_based.rs
aintnostressin: Ok, so please either remove this todo or create an issue if any action needs to be taken.

nexus-watcher/src/service/indexer/key_based.rs
ok300: Removed: f406eed94c17ac8e798b206dc49e144f8f6689a0

nexus-watcher/src/service/indexer/key_based.rs
ok300: The impact is so small, and a 3rd party HS would have to have so many users to impact indexing by even 5 - 10 seconds, that this won't matter for a long time.

However to keep this moving, it's done: 4b3c6ddb3eb01b3c2d17a290f9ff17633f9e1781

nexus-watcher/src/service/indexer/key_based.rs
ok300: Done: 3d07fcca40794f1c88d66b37442e30253034cf3d

nexus-watcher/src/service/indexer/key_based.rs
ok300: Done, replaced TODO with #885

nexus-watcher/src/service/indexer/key_based.rs
ok300: > 404

PKDNS is the authoritative mapping, so If PKDNS says this user is on that HS, I don't see how we can "handle" the 404 in a proper way. It could be the HS is temporarily offline, or that republishing temporarily failed.

In my view, it is fine to log these 404s and retry on next run. This is only "hammering" the HS if the `watcher_sleep` , the delay between key-based processor runs, is too low:

https://github.com/pubky/pubky-nexus/blob/2aeede0ac21958dbbabde441851d8edc0777c80e/nexus-watcher/src/service/mod.rs#L118-L121

How would you suggest to handle the 404s?

> 429

We don't have a 429 backoff logic, we have a backoff logic in general for external HSs that fail to respond or otherwise timeout.

Added 429 backoff + retry for the per-user calls of the key-based processor: e64c194438cd7b820889cb931db10ffc0902a5a0

nexus-common/src/db/kv/index/sorted_sets.rs
```diff
@@ -37,11 +37,43 @@ pub const SORTED_PREFIX: &str = "Sorted";
...
 
+/// Checks multiple (key, member) pairs in Redis sorted sets using a single
+/// pipeline of `ZSCORE` commands.
+///
+/// `keys` and `members` are parallel slices of the same length — each
+/// `(keys[i], members[i])` pair produces one `ZSCORE` call. Returns scores in
+/// the same order, with `None` for absent members.
+pub async fn check_members(
+    prefix: &str,
+    keys: &[&str],
```
aintnostressin: kyes and members looks like a good candidate for a slice of tuples -> then there is no need for len check

nexus-common/src/db/kv/traits.rs
```diff
@@ -546,6 +546,25 @@ pub trait RedisOps: Serialize + DeserializeOwned + Send + Sync {
...
 
+    /// Checks multiple (key, member) pairs across Redis sorted sets in a single
+    /// pipeline round-trip.
+    ///
+    /// `keys_parts` and `members_parts` are parallel slices — each pair is joined
+    /// and passed as one `ZSCORE`. Returns scores in the same order, with `None`
+    /// for absent members.
+    async fn check_sorted_set_members(
+        prefix: Option<&str>,
+        keys_parts: &[&[&str]],
```
aintnostressin: this also looks like a good candidate for a slice of tuples

nexus-watcher/src/service/indexer/key_based.rs
aintnostressin: It may be authoritative mapping but pkdsn record + absence of user on HS should not affect other users on that homeserver. Right now there can be hundreds fake pkdns records pointing to homeserver and on each run indexer will hit only 404 + 429 errors.

proposed solution: shuffle user list + mute for a while users that have pkdsn record and hs responded with 404

nexus-watcher/tests/service/key_based_event_processor.rs
```diff
@@ -370,6 +370,45 @@ async fn key_based_processor_continues_after_non_infrastructure_fetch_error() ->
...
+            ])
+            .await,
+    );
+    let handler = create_mock_handler(Ok(()), None);
+    let processor = processor(homeserver, handler.clone(), source.clone());
+
+    let started_at = Instant::now();
+    processor.run().await?;
+
+    assert!(
```
aintnostressin: this should simply count how many times function was called.

aintnostressin (approved):
Looks ok.

Nothing blocking anymore from my side. 
Small improvements can be introduced in follow-up PRs.

nexus-common/src/db/kv/index/sorted_sets.rs
ok300: Done: 3feb7cfbecc49ff5fcf9cb1d5a9b93385222a343

nexus-common/src/db/kv/traits.rs
ok300: Done: 3feb7cfbecc49ff5fcf9cb1d5a9b93385222a343

nexus-watcher/tests/service/key_based_event_processor.rs
ok300: Done: 1241224a6bfa71aa3931ffca84eb9ecbcf8a18c8

nexus-watcher/src/service/indexer/key_based.rs
ok300: Opened PR for 404 handling: #891

PR #802, opened 2026-04-07
feat: follow events handlers partial recovery

aintnostressin:
Prepares follow handlers for partial recovery.

  - DEL handler: graph-last ordering — Runs Redis SREM cleanup first, then deletes the FOLLOWS edge last. On retry after partial failure, the graph edge survives so the handler re-enters and re-runs the idempotent Redis cleanup.                    
  - DEL handler: guarded counters/notifications — Uses Followers::check_in_index to verify the follower is still in the Redis set before decrementing UserCounts or sending lost-follow notifications, preventing double-decrement on retry.            
  - DEL handler: friendship state captured early — Friends::check runs before any deletes so were_friends remains correct even if cleanup is interrupted mid-flight.                                                                                    
  - PUT handler: retry recovery — When put_to_graph returns Updated (edge already exists from a prior attempt), re-runs idempotent Followers/Following put_to_index writes to recover from partial failures where graph succeeded but Redis didn't.     
  Skips counters and notifications (prefer 0 over N).                                                                                                                                                                                                   
  - Idempotent no-op on replay — Follow DEL silently swallows the graph OperationOutcome, so replays of an already-deleted relationship return Ok(()).                                                                                                  
  - Adds del_idempotent.rs and put_idempotent.rs integration tests covering retry scenarios.

### DIFF
--- a/nexus-watcher/src/events/handlers/follow.rs
+++ b/nexus-watcher/src/events/handlers/follow.rs
@@ -91,7 +91,7 @@ pub async fn sync_del(
     follower_id: PubkyId,
     followee_id: PubkyId,
 ) -> Result<(), EventProcessorError> {
-    // Check friendship while graph edge still exists
+    // Check friendship while Redis follow sets are still populated
     let were_friends = Friends::check(&follower_id, &followee_id).await?;
 
     // Guard counters/notifications: only run if still in Redis index (first attempt).
--- a/nexus-watcher/src/events/handlers/follow.rs
+++ b/nexus-watcher/src/events/handlers/follow.rs
@@ -19,8 +19,22 @@ pub async fn sync_put(
     // SAVE TO GRAPH
     // (follower_id)-[:FOLLOWS]->(followee_id)
     match Followers::put_to_graph(&follower_id, &followee_id).await? {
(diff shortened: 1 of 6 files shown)


nexus-watcher/src/events/handlers/follow.rs
```diff
@@ -77,41 +91,41 @@ pub async fn sync_del(
...
     followee_id: PubkyId,
 ) -> Result<(), EventProcessorError> {
-    match Followers::del_from_graph(&follower_id, &followee_id).await? {
-        // Both users exists but they do not have that relationship
-        OperationOutcome::Updated => Ok(()),
-        OperationOutcome::MissingDependency => Err(EventProcessorError::SkipIndexing),
-        OperationOutcome::CreatedOrDeleted => {
-            // Check if the users are friends. Is this a break? :(
-            let were_friends = Friends::check(&follower_id, &followee_id).await?;
+    // Check friendship while graph edge still exists
```
ok300:
```suggestion
    // Check friendship while Redis follow sets are still populated
```

These checks perform an index, not a graph, lookup.

ok300 (approved):

tipogi (approved):

Issue #806, opened 2026-04-09
chore: move URI utils to `pubky-app-specs`


ok300:
The following URI utils introduced in #777 belong in `pubky-app-specs`, so they should be moved there:
- `nexus_common::models::resource::normalize_uri`
- `nexus_common::models::resource::normalize_parsed_uri`

ok300:
Also include `resource_id.rs` introduced in #849

Issue #735, opened 2026-02-19
Event handler errors that cannot be covered by `RetryManager`


ok300:
The various branches of `handle_put_event` can fail for all sorts of reasons that cannot be addressed later by `RetryManager`, like:
- missing permissions (to write a file)
- file already exists (when saving attachment file)
- etc

We need a clean way to propagate and handle these errors. The main distinction is these cannot be covered by `RetryManager`.

PR #809, opened 2026-04-20
feat: RetryEvent processor

aintnostressin:
This PR implements a full retry pipeline for failed watcher events, replacing the previous `// TODO save event in retry manager` stub.

   ### What's new

   **Retry infrastructure** (`nexus-watcher/src/events/retry/`)
   - `RetryScheduler` — enqueues a failed event onto the Redis retry queue with an initial backoff delay; called by `handle_error` in `TEventProcessor`
   - `RetryStore` trait with two backends: `RedisRetryStore` (production) and `InMemoryRetryStore` (per-test isolation so parallel tests don't step on each other's queue state)
   - `RetryProcessor` — a periodic background task (10 s tick) that drains events whose `next_retry_at <= now` in batches of 100, applies exponential backoff on re-failure, and dead-letters after configurable max retries


nexus-common/src/config/watcher.rs
```diff
@@ -35,6 +50,42 @@ pub const MODERATED_TAGS: [&str; 6] = [
     "il_adult_nu_sex_act",
 ];
 
+/// Retry configuration settings
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct EventRetryConfig {
```
ok300, line 56: The fields + their annotations + their rustdocs are too dense.

It would help readability to separate them by newline:

```rust
    /// Transient error retry limit before dead-letter
    #[serde(default = "default_max_retries")]
    pub max_retries: u32,

    /// Safety net for homeservers that disappear silently (no DEL events, content just gone)
    #[serde(default = "default_max_dependency_retries")]
    pub max_dependency_retries: u32,
```

nexus-common/src/models/event/homeserver_parsed_uri.rs
```diff
@@ -0,0 +1,246 @@
...
+
+/// Parsed URI representation that can handle both:
+/// 1. Standard pubky-app-specs URIs (pubky://<user_id>/pub/pubky.app/...)
+/// 2. Universal tag URIs from other apps (pubky://<user_id>/pub/<app>/tags/<tag_id>)
+///
+/// This is needed because homeservers may return events from applications other than
+/// pubky.app (e.g., eventky.app, mapky), and ParsedUri from pubky-app-specs strictly
+/// requires the app path to be "pubky.app".
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum HomeserverParsedUri {
```
ok300, line 13: This is not Nexus-specific. In fact, it looks like it belongs in `pubky-app-specs`, as an extension or revamp of `ParsedUri`.

That could also cover other "universal tag follow-ups" like #806 .

nexus-common/default.config.toml
```diff
@@ -38,6 +38,20 @@ moderated_tags = [
     "il_adult_nu_sex_act",
 ]
 
+# Event retry configuration
```
ok300, line 41: Looking at the other fields set in other `*.config.toml` files, it seems default fields are supposed to be un-commented and set to their default values.

So to be consistent, these new config fields, here and in `watcher-config.toml`, should be un-commented.

nexus-watcher/src/events/retry/event.rs
```diff
@@ -31,59 +34,41 @@ impl RedisOps for RetryEvent {
...
-        let user_id = &event_uri.user_id;
-        let event_resource = &event_uri.resource;
-
-        match event_uri.resource.id() {
-            Some(id) => format!("{user_id}:{event_resource}:{id}"),
-            None => format!("{user_id}:{event_resource}"),
-        }
+    /// Generates an index key from a parsed URI in the format `"{pubkyId}:{resource}"`.
+    /// Returns `None` if the URI cannot be parsed as a valid `HomeserverParsedUri`.
+    pub fn generate_index_key(parsed_uri: HomeserverParsedUri) -> Option<String> {
```
ok300: No parsing happens inside the method anymore, so there's no possibility it could return `None`.

Why not then directly return `String`?

nexus-watcher/src/events/handlers/tag.rs
```diff
@@ -382,6 +389,7 @@ pub async fn del(
     let maybe_row = fetch_row_from_graph(query).await?;
 
     let Some(row) = maybe_row else {
+        info!("THere is not tag ...");
```
ok300: Committed by mistake?

nexus-watcher/src/events/handlers/universal_tag.rs
```diff
@@ -36,6 +40,13 @@ pub async fn try_handle(
     })
 }
 
+/// Extract the app name from a universal tag URI for logging purposes.
+fn extract_app_from_uri(uri: &str) -> String {
```
ok300, line 44: Another candidate for `pubky-app-specs`.

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub shutdown_rx: Receiver<bool>,
+    pub config: RetryProcessorConfig,
+    /// Persistence backend for retry events. Production wiring uses
+    /// [`RedisRetryStore`]; tests swap in an in-memory store for isolation.
+    pub store: Arc<dyn RetryStore>,
+    /// Provided only to satisfy [`TEventProcessor::retry_scheduler`]. The retry
+    /// processor bypasses [`TEventProcessor::handle_error`] (it calls the event
+    /// handler directly and reschedules via [`Self::schedule_retry`]), so the
+    /// scheduler is never actually invoked from this processor.
+    pub retry_scheduler: Arc<RetryScheduler>,
```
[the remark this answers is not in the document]
aintnostressin: done

nexus-watcher/src/events/retry/event.rs
```diff
@@ -113,4 +119,49 @@ impl RetryEvent {
...
+            None,             // min_score (end → min in get_range)
+            Some(0),          // skip
+            limit,
+            SortOrder::Ascending,
+            Some(RETRY_MANAGER_PREFIX),
+        )
+        .await
+        .map_err(|e| EventProcessorError::generic(format!("Failed to fetch retry events: {}", e)))
+    }
 }
```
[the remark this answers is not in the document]
aintnostressin: done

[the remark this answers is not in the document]
ok300: The fix proposed by greptile ( 2297419346bed55be1e26741d934ec29d5a39b18 ) is more confusing than the args before.

You now have to read 6 lines of comments to understand ..what? To understand one is a start and the other is an end param, which you can also tell by reading the `try_from_index_sorted_set` method signature.

I'd say it's best to revert this commit and just ignore greptile here.

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,285 @@
...
+                    return Ok(());
+                }
+
+                // Schedule retry with backoff
+                self.schedule_retry(&retry_event, resource_key, &e).await?;
+
+                // Propagate infrastructure errors to stop the batch
+                if e.is_infrastructure() {
+                    return Err(e);
+                }
```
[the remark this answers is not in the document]
aintnostressin: done

nexus-common/default.config.toml
aintnostressin: done [b079f9a](https://github.com/pubky/pubky-nexus/pull/809/commits/b079f9a946b71066a3fbab25f8de080ade9d4280)

nexus-common/src/models/event/homeserver_parsed_uri.rs
aintnostressin: It's tricky because if URL doesn't fall into any know pubky-app-spec category it's being marked as `UniversalTag`. 

`UniversalTag` by definition is not pubky app specific so I decided to keep it in Nexus because it seems like only nexus is aware of it. 

I suggest to keep it (imho this is an improvement) and change later if needed.

nexus-watcher/src/events/handlers/universal_tag.rs
aintnostressin: Same as https://github.com/pubky/pubky-nexus/pull/809#discussion_r3111199359

nexus-watcher/src/events/retry/event.rs
aintnostressin: done [d63c12c](https://github.com/pubky/pubky-nexus/pull/809/commits/d63c12c2e5d672407ba58390b8b246bebf6b498e) , simplifies a lot

nexus-watcher/src/events/handlers/tag.rs
aintnostressin: removed [d63c12c](https://github.com/pubky/pubky-nexus/pull/809/commits/d63c12c2e5d672407ba58390b8b246bebf6b498e)

nexus-common/default.config.toml
ok300: `examples/watcher/watcher-config.toml` still has new configs commented.

nexus-watcher/src/events/handlers/universal_tag.rs
```diff
@@ -91,25 +106,21 @@ async fn handle_del(info: AppTagInfo) -> Result<(), EventProcessorError> {
...
     // Case-insensitive scheme check per RFC 3986 (safe UTF-8 access)
-    let rest = match uri.get(..8) {
-        Some(prefix) if prefix.eq_ignore_ascii_case("pubky://") => &uri[8..],
-        _ => return None,
-    };
+    let rest = uri
+        .strip_prefix("pubky://")
+        .or_else(|| uri.strip_prefix("PUBKY://"))
+        .or_else(|| uri.strip_prefix("Pubky://"))
+        .or_else(|| to_ascii_lower_prefix(uri, "pubky://"))?;
```
ok300:
```suggestion
    let rest = to_ascii_lower_prefix(uri, "pubky://")?;
```

nexus-watcher/src/events/handlers/universal_tag.rs
```diff
@@ -1,10 +1,13 @@
...
 use nexus_common::models::event::{EventProcessorError, EventType};
-use pubky_app_specs::{PubkyAppTag, PubkyId};
-use tracing::debug;
+use pubky_app_specs::PubkyId;
+use tracing::{debug, instrument};
 
 use super::tag;
 
+/// Path segment separating the app name from the tags section.
+const TAGS_SEPARATOR: &str = "/tags/";
```
ok300, line 9: This `const` is only used once, so it makes little sense to extract and reference it.

Better to directly use `"/tags/"` explicitly where needed.

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,313 @@
...
+    pub shutdown_rx: Receiver<bool>,
+    pub config: RetryProcessorConfig,
+    /// Persistence backend for retry events. Production wiring uses
+    /// [`RedisRetryStore`]; tests swap in an in-memory store for isolation.
+    pub store: Arc<dyn RetryStore>,
+}
+
+/// Configuration for the retry processor
+#[derive(Debug, Clone)]
+pub struct RetryProcessorConfig {
```
ok300: This is a 1:1 mirror of `EventRetryConfig`, they have identical fields.

Maybe this can be avoided and `EventRetryConfig` used directly where needed:

```diff
pub struct RetryProcessor {
     pub files_path: PathBuf,
     pub event_handler: Arc<dyn EventHandler>,
     pub shutdown_rx: Receiver<bool>,
-    pub config: RetryProcessorConfig,
+    pub config: EventRetryConfig,
     /// Persistence backend for retry events. Production wiring uses
     /// [`RedisRetryStore`]; tests swap in an in-memory store for isolation.
     pub store: Arc<dyn RetryStore>,
}
```

nexus-common/src/config/watcher.rs
```diff
@@ -35,6 +50,42 @@ pub const MODERATED_TAGS: [&str; 6] = [
...
+    pub max_backoff_secs: u64,
+    /// Base for MissingDependency polling backoff (seconds)
+    #[serde(default = "default_initial_missing_dep_backoff_secs")]
+    pub initial_missing_dep_backoff_secs: u64,
+    /// Backoff ceiling for MissingDependency (seconds)
+    #[serde(default = "default_max_missing_dep_backoff_secs")]
+    pub max_missing_dep_backoff_secs: u64,
+}
+
+impl Default for EventRetryConfig {
```
ok300, line 71: Since there is an `impl Default` for `EventRetryConfig`, then the struct can avoid specifying the default methods for each field:

```diff
  #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+ #[serde(default)]
  pub struct EventRetryConfig {
      /// Transient error retry limit before dead-letter
-     #[serde(default = "default_max_retries")]
      pub max_retries: u32,
      /// Safety net for homeservers that disappear silently (no DEL events, content just gone)
-     #[serde(default = "default_max_dependency_retries")]
      pub max_dependency_retries: u32,
      /// Base for exponential backoff on transient retries (seconds)
-     #[serde(default = "default_initial_transient_backoff_secs")]
      pub initial_backoff_secs: u64,
      /// Backoff ceiling for transient retries (seconds)
-     #[serde(default = "default_max_transient_backoff_secs")]
      pub max_backoff_secs: u64,
      /// Base for MissingDependency polling backoff (seconds)
-     #[serde(default = "default_initial_missing_dep_backoff_secs")]
      pub initial_missing_dep_backoff_secs: u64,
      /// Backoff ceiling for MissingDependency (seconds)
-     #[serde(default = "default_max_missing_dep_backoff_secs")]
      pub max_missing_dep_backoff_secs: u64,
  }
```

The `default_*` functions can also be removed.

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -114,13 +127,42 @@ impl HsEventProcessor {
...
+                        hs.put_to_index().await?;
+                    }
                     Err(e) => warn!("{e}"),
                 }
-            } else {
-                self.process_event_line(line).await?;
+                continue;
+            }
+
+            // Process the event line
```
ok300: What do you think of #810 ? It re-uses the trait's default impl, which is currently duplicated in this segment.

ok300: Since you approved it, I merged the PR into this branch.

nexus-watcher/src/events/retry/processor.rs
aintnostressin: done [ce4074a](https://github.com/pubky/pubky-nexus/pull/809/commits/ce4074a5c0b89734076ffbb8de85bd52cd7579be)

nexus-common/src/config/watcher.rs
aintnostressin: done [82450c7](https://github.com/pubky/pubky-nexus/pull/809/commits/82450c7b64d882a4566fbd606bfe98bfb64e7ffb)

nexus-watcher/src/events/handlers/universal_tag.rs
aintnostressin: done [d5a87a3](https://github.com/pubky/pubky-nexus/pull/809/commits/d5a87a3ee47167f6428057941ecffcb83defdb6d)

nexus-common/default.config.toml
aintnostressin: done [c17d5de](https://github.com/pubky/pubky-nexus/pull/809/commits/c17d5de6881f9d37f2412b062bd293ba049f93ec)

nexus-common/src/config/watcher.rs
aintnostressin: @ok300 per field marcos were removed - do you still think it's too dense ?

nexus-common/default.config.toml
ok300: Done in c17d5de6881f9d37f2412b062bd293ba049f93ec

nexus-watcher/src/events/handlers/universal_tag.rs
ok300: Done in d5a87a3ee47167f6428057941ecffcb83defdb6d

nexus-common/src/config/watcher.rs
ok300: It would still help to split them, but it's better now. Let's leave them as they are.

nexus-common/src/db/connectors/pubky.rs
```diff
@@ -7,13 +7,90 @@ use tracing::debug;
...
+    #[error("Build failed: {message}")]
+    BuildFailed { message: String },
+
+    #[error("Parse failed: {message}")]
+    ParseFailed { message: String },
+}
+
+impl PubkyClientErrorKind {
+    /// Maps a pubky::Error to PubkyClientErrorKind
+    pub fn from_pubky_error(err: pubky::Error) -> Self {
```
ok300: This should be extracted as a

```rust
impl From<pubky::Error> for PubkyClientErrorKind {
  //
}
```

nexus-watcher/src/events/retry/scheduler.rs
```diff
@@ -0,0 +1,102 @@
...
+
+        let next_retry_at = Utc::now().timestamp_millis() + backoff_ms;
+        let retry_event = RetryEvent::new(event_type.clone(), uri.to_string(), next_retry_at);
+
+        self.store.put(&key, &retry_event).await?;
+        warn!("Queued raw event for retry: {uri}");
+        Ok(())
+    }
+
+    async fn enqueue(
```
ok300, line 58: If the watcher restarts and replays an event that already has an in-progress retry entry in Redis, `enqueue` constructs a fresh `RetryEvent::new(..)` with `retry_count = 0` and stores it with `store.put`, overwriting the existing entry.

A simple solution would be to `store.get(&key)` first and, if the entry exists, to preserve `retry_count` in the new `RetryEvent`.

nexus-watcher/src/events/retry/scheduler.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /// Returns the missing dependency initial backoff in milliseconds.
+    pub fn missing_dep_backoff_ms(&self) -> i64 {
+        self.initial.missing_dep_ms
+    }
+
+    /// Queues a retry event for an arbitrary URI with a custom backoff.
+    /// This is a generic method that does not require a full Event object.
+    pub async fn enqueue_raw(
```
ok300: This method appears unused. If so, it should be removed.

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -129,54 +139,55 @@ pub trait TEventProcessor: Send + Sync + 'static {
...
         error: EventProcessorError,
     ) -> Result<(), EventProcessorError> {
-        if matches!(error, EventProcessorError::MissingDependency { .. }) {
-            // TODO save event in retry manager
-            Ok(())
-        } else if error.is_infrastructure() {
-            return Err(error);
-        } else {
-            Ok(())
+        match &error {
```
ok300: Many branches say "log and continue" but don't log.

For troubleshooting later on, it would make sense to log these things on `debug` and the important ones on `warn`.

nexus-common/src/db/connectors/pubky.rs
aintnostressin: done [ce8958f](https://github.com/pubky/pubky-nexus/pull/809/commits/ce8958fe665f044707a1205ccd973a7f58414859)

nexus-watcher/src/events/retry/scheduler.rs
aintnostressin: removed [79d4676](https://github.com/pubky/pubky-nexus/pull/809/commits/79d4676d1b62940cf50af36ef12a0aa4566d6268)

nexus-watcher/src/service/indexer/mod.rs
aintnostressin: added [499baaa](https://github.com/pubky/pubky-nexus/pull/809/commits/499baaa0e8675e2e775b8bb68ff64acbb8cedce6)

nexus-watcher/src/events/retry/scheduler.rs
aintnostressin: That's true but ... the design assumes there can be one RetryEvent per Resource (POST, TAG, etc.).
I decided to overwrite it because if new event for a given resource comes and fails we should give it a full chance (reset counter) to succeed. I think this scenario is more likely to happen.

nexus-common/src/models/event/errors.rs
```diff
@@ -70,7 +70,9 @@ impl From<ModelError> for EventProcessorError {
 
 impl From<pubky::Error> for EventProcessorError {
     fn from(e: pubky::Error) -> Self {
-        EventProcessorError::client_error(e.to_string())
+        EventProcessorError::PubkyClientError(PubkyClientError::ClientError(
+            PubkyClientErrorKind::from(e),
+        ))
```
ok300:
```suggestion
        EventProcessorError::PubkyClientError(PubkyClientError::from(PubkyClientErrorKind::from(e)))
```

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -129,54 +139,55 @@ pub trait TEventProcessor: Send + Sync + 'static {
...
+            EventProcessorError::PubkyClientError(_) => {
+                warn!("Client error, queuing event for retry: {error}");
+                if let Some(s) = self.retry_scheduler() {
+                    s.queue_transient(event).await
+                } else {
+                    Ok(())
+                }
+            }
+            _ if error.is_infrastructure() => {
+                // Infrastructure error - stop the batch
```
ok300:
```suggestion
                warn!("Infrastructure error, stopping batch: {error}");
```

nexus-watcher/src/events/handlers/universal_tag.rs
ok300: Looks like it's removed in #816

nexus-watcher/src/events/handlers/universal_tag.rs
ok300: Addressed in #816

nexus-common/src/universal_tag/homeserver_parsed_uri.rs
```diff
@@ -0,0 +1,246 @@
...
+
+/// Parsed URI representation that can handle both:
+/// 1. Standard pubky-app-specs URIs (pubky://<user_id>/pub/pubky.app/...)
+/// 2. Universal tag URIs from other apps (pubky://<user_id>/pub/<app>/tags/<tag_id>)
+///
+/// This is needed because homeservers may return events from applications other than
+/// pubky.app (e.g., eventky.app, mapky), and ParsedUri from pubky-app-specs strictly
+/// requires the app path to be "pubky.app".
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub enum HomeserverParsedUri {
```
ok300: Addressed in #816

nexus-watcher/src/events/retry/event.rs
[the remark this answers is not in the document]
aintnostressin: reverted [9b49217](https://github.com/pubky/pubky-nexus/pull/809/commits/9b492170cf60dfe4f1526148f223434930aa6004)

nexus-watcher/src/events/retry/scheduler.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Returns the transient error initial backoff in milliseconds.
+    pub fn transient_backoff_ms(&self) -> i64 {
+        self.initial.transient_ms
+    }
+
+    /// Returns the missing dependency initial backoff in milliseconds.
+    pub fn missing_dep_backoff_ms(&self) -> i64 {
+        self.initial.missing_dep_ms
+    }
+
```
ok300:
```suggestion
```

The two methods aren't used anywhere.

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,283 @@
...
+                // Schedule retry with backoff (increments retry_count)
+                self.schedule_retry(&retry_event, index_key, &e).await?;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Schedule an event for retry with exponential backoff
+    async fn schedule_retry(
```
ok300: This can be simplified: #821

nexus-watcher/src/events/retry/scheduler.rs
```diff
@@ -0,0 +1,74 @@
...
+        initial_backoff_ms: i64,
+        reason: &str,
+    ) -> Result<(), EventProcessorError> {
+        let key: RetryEventIndexKey = event.uri.clone();
+
+        let next_retry_at = Utc::now().timestamp_millis() + initial_backoff_ms;
+        let retry_event =
+            RetryEvent::new(event.event_type.clone(), event.uri.clone(), next_retry_at);
+
+        self.store.put(&key, &retry_event).await?;
```
ok300:
```suggestion
        // New EventRetries for the same URI will reset the retry_count
        // The HS state changed since the earlier event, so we disregard previous retry attempts
        self.store.put(&key, &retry_event).await?;
```

Would make sense to document the `retry_count` reset topic, as the intention is not clear from the code.

nexus-common/src/db/connectors/pubky.rs
```diff
@@ -7,13 +7,91 @@ use tracing::debug;
 
 static PUBKY_SINGLETON: OnceCell<Arc<Pubky>> = OnceCell::const_new();
 
+#[derive(Debug, Error, Clone, Serialize, Deserialize)]
+pub enum PubkyClientErrorKind {
```
ok300: These new types belong into the existing `PubkyClientError`: #823

nexus-common/src/universal_tag/homeserver_parsed_uri.rs
```diff
@@ -0,0 +1,246 @@
...
+    #[test]
+    fn test_uppercase_scheme() {
+        let user_id = "operrr8wsbpr3ue9d4qj41ge1kcc6r7fdiy6o3ugjrrhi4y77rdo";
+        let tag_id = "ABC123";
+        let uri = format!("PUBKY://{user_id}/pub/mapky/tags/{tag_id}");
+        let result = HomeserverParsedUri::try_from(uri.as_str());
+        // This should fail because url::Url::parse doesn't allow uppercase schemes in the same way
+        // For case-insensitive support, callers may need to normalize the scheme first
+        // The universal_tag variant handles this separately with eq_ignore_ascii_case
+        assert!(result.is_err() || result.is_ok()); // Accept either behavior for now
```
ok300: This test is a no-op, the assertion will always succeed. Please fix.

nexus-watcher/src/events/retry/store.rs
```diff
@@ -0,0 +1,194 @@
...
+        Self {
+            inner: Mutex::new(HashMap::new()),
+        }
+    }
+
+    /// Convenience constructor returning an `Arc` already erased to `dyn RetryStore`,
+    /// the form the [`super::RetryProcessor`] expects.
+    pub fn arc() -> Arc<dyn RetryStore> {
+        Arc::new(Self::new())
+    }
```
ok300:
```suggestion
```

This is not used anywhere, should be removed.

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,255 @@
...
+        index_key: &RetryEventIndexKey,
+        retry_event: RetryEvent,
+    ) -> Result<(), EventProcessorError> {
+        // Reconstruct the event line and parse the event
+        // Event format is "METHOD URI" (e.g., "PUT pubky://...")
+        let event_type_str = match retry_event.event_type {
+            EventType::Put => "PUT",
+            EventType::Del => "DEL",
+        };
+        let event_line = format!("{} {}", event_type_str, retry_event.event_uri);
```
ok300:
```suggestion
        let event_line = format!("{} {}", retry_event.event_type, retry_event.event_uri);
```

`EventType` already implements `Display`.

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        // Parse the event from the line - if corrupted, remove and continue
+        let event = match Event::parse_event(&event_line, self.files_path().clone()) {
+            Ok(ParseResult::Parsed(event)) => event,
+            Ok(ParseResult::Skipped) | Err(_) => {
+                warn!(
+                    "Corrupted retry entry for key {}, removing: {}",
+                    index_key, event_line
+                );
```
ok300:
```suggestion
                warn!("Corrupted retry entry for key {index_key}, removing: '{event_line}'");
```

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,255 @@
...
+                self.store.remove(index_key).await?;
+                return Ok(());
+            }
+            Ok(ParseResult::UnrecognizedUri { reason, .. }) => {
+                warn!(
+                    "Unrecognized URI in retry entry for key {}: {}",
+                    index_key, reason
+                );
```
ok300:
```suggestion
                warn!("Unrecognized URI in retry entry for key {index_key}, removing: {reason}");
```

nexus-common/src/universal_tag/homeserver_parsed_uri.rs
aintnostressin: fixed: [6b7fc63](https://github.com/pubky/pubky-nexus/pull/809/commits/6b7fc638e4af2b1eebb83f288884e993649da329)

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,244 @@
...
+    async fn run_internal(self: Arc<Self>) -> Result<(), EventProcessorError> {
+        let now = Utc::now().timestamp_millis();
+
+        loop {
+            if *self.shutdown_rx.borrow() {
+                debug!("Shutdown detected; exiting retry processing loop");
+                return Ok(());
+            }
+
+            let events = self.fetch_ready_events(now).await?;
```
ok300, line 54:
```suggestion
        loop {
            if *self.shutdown_rx.borrow() {
                debug!("Shutdown detected; exiting retry processing loop");
                return Ok(());
            }

            let now = Utc::now().timestamp_millis();
            let events = self.fetch_ready_events(now).await?;
```

Since this is a `loop`, shouldn't `now` be refreshed on each loop iteration, e.g. inside the loop instead of outside?

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+    fn retry_scheduler(&self) -> Option<&Arc<RetryScheduler>> {
+        None
+    }
+
+    async fn run_internal(self: Arc<Self>) -> Result<(), EventProcessorError> {
+        let now = Utc::now().timestamp_millis();
+
+        loop {
```
ok300, line 53: Why is this a `loop`? Does it need to be one?

From my understanding, the retry processor is triggered periodically by the `nexusd` service:

https://github.com/pubky/pubky-nexus/blob/7b3bdceb077c08e01295d801fa895d7df16e523f/nexus-watcher/src/service/mod.rs#L124-L127

nexus-watcher/src/service/mod.rs
```diff
@@ -101,6 +106,9 @@ impl NexusWatcher {
             shutdown_rx.clone(),
         ));
 
+        // Create retry processor
+        let retry_processor = Arc::new(RetryProcessor::new(&config, shutdown_rx.clone()));
```
ok300, line 110: For the other two event processors, created just above this, the struct is called `...ProcessorRunner`.

Maybe `RetryProcessor` could be renamed to `RetryProcessorRunner` for consistency?

ok300, line 110: For the other two event processors, created just above this, the constructor pattern is

```rust
HsEventProcessorRunner::from_config(&config, shutdown_rx.clone())
KeyBasedEventProcessorRunner::from_config(&config, shutdown_rx.clone())
```

whereas this one is using `new`:

```rust
RetryProcessor::new(&config, shutdown_rx.clone())
```

For consistency, maybe rename `RetryProcessor::new` -> `RetryProcessor::from_config`?

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -129,54 +139,55 @@ pub trait TEventProcessor: Send + Sync + 'static {
...
     /// Called in the event processing loop.
     ///
-    /// It re-throws the erorr, which signals the caller should break the processing loop
-    /// and persist the cursor at the point where this error occurred. This is because, since
-    /// it's an infrastructure error, it is likely that trying to process the next event line
-    /// would result in the same issue.
+    /// Returns:
+    /// - `Ok(())` - Continue processing the batch (for MissingDependency, InvalidEventLine, SkipIndexing, 404)
+    /// - `Err(e)` - Stop processing and return error (for infrastructure errors)
     async fn handle_error(
```
ok300, line 149: Shouldn't `handle_error` include an explicit check whether an error `is_retryable` and only enqueue it if so?

nexus-watcher/src/events/mod.rs
```diff
@@ -10,14 +11,47 @@ pub mod retry;
...
+            EventType::Put => handle_put_event(event, self.moderation.clone()).await,
+            EventType::Del => handle_del_event(event).await,
+        }?;
+
+        event.store_event().await?;
+        Ok(())
+    }
 }
 
 pub async fn handle_put_event(
```
ok300, line 47: When an event is first processed, there is a class of errors which may result in retry events, but where retrying would not help, like for example:
- permission errors on the local Nexus filesystem, encountered when trying to save attachments
- attachment file fails to be saved because file with same name already exists

They are described in #735.

It sounds like these are "infrastructure errors" but I'm not sure they're currently treated as such.

The main concern is that they don't cause the retry manager to endlessly try such ops that would never succeed.

What do you think? Is there such a risk? And if so, are such cases already covered?

nexus-watcher/src/events/retry/processor.rs
aintnostressin: To avoid loading all events into memory. 
This inner loops is used to process all retry events in batches.

nexus-watcher/src/events/retry/processor.rs
aintnostressin: No, for this `RetryProcessor` run we want to process only events that were ready relatively to `now` created at the beginning of `run_internal` fn execution.

nexus-watcher/src/service/mod.rs
aintnostressin: I prefer to keep it as it is but it's just naming so feel free to change it.

aintnostressin: There is no other variant of creating this type so it make sense to keep plain `new`.
Aditionally i prefer to keep `from_*` for cases where there is only one source of creation.

If you really think this is something that has to be changed - i'm open for it.

nexus-watcher/src/service/indexer/mod.rs
aintnostressin: That's a good idea, it's changed in #825

nexus-watcher/src/events/mod.rs
aintnostressin: There is a 'class' of errors that, if no action is taken, may lead to endless retry loop.
They should eventually succeed, and they will if nexus administrator takes action.

We can dead letter them after some time but in my opinion that's worse than keeping them and naively retrying and self healing.
We may add alerts on top of that so administrator knows that some action has to be taken.

It's good that you raised this topic, what is your opinion on that ?

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,244 @@
...
+            if *self.shutdown_rx.borrow() {
+                debug!("Shutdown detected; exiting retry processing loop");
+                return Ok(());
+            }
+
```
ok300:
```suggestion
```

This shutdown signal check is done again a few lines below, when retry events are processed.

The only benefit of checking this here is to catch signals that came before `fetch_ready_events`, as the next check would catch any signal that came after.

However the `fetch_ready_events` consists of a few fast Redis lookups. So checking for the signal before this is overkill, as it makes little difference.

I would therefore remove this and rely on the signal check below.

nexus-watcher/src/events/retry/store.rs
```diff
@@ -0,0 +1,187 @@
...
+        for (index_key, _score) in key_score_pairs {
+            match RetryEvent::get_from_index(&index_key).await? {
+                Some(event) => events.push((index_key.clone(), event)),
+                None => {
+                    // Sorted-set entry with no JSON state — tombstone, clean up and skip.
+                    debug!(
+                        "Stale retry entry detected for key {}, cleaning up",
+                        index_key
+                    );
+                    RetryEvent::remove_from_index(&index_key).await?;
```
ok300: These calls can be batched: #826

nexus-watcher/src/events/retry/store.rs
```diff
@@ -0,0 +1,191 @@
...
+    ) -> Result<Vec<(RetryEventIndexKey, RetryEvent)>, EventProcessorError> {
+        let key_score_pairs = match RetryEvent::fetch_ready(now, limit).await? {
+            Some(pairs) => pairs,
+            None => return Ok(Vec::new()),
+        };
+
+        let mut events = Vec::with_capacity(key_score_pairs.len());
+        let mut stale: Vec<RetryEventIndexKey> = Vec::new();
+        for (index_key, _score) in key_score_pairs {
+            match RetryEvent::get_from_index(&index_key).await? {
```
ok300: These calls can also be batched: #827

nexus-watcher/src/events/retry/event.rs
```diff
@@ -31,86 +36,132 @@ impl RedisOps for RetryEvent {
...
+
+    /// Fetches events from the retry queue that are ready to be retried (next_retry_at <= now)
+    /// Returns Vec<(index_key, score)> pairs for events ready for retry
+    /// # Arguments
+    /// * `now` - Current time in milliseconds since epoch
+    /// * `limit` - Maximum number of events to fetch per batch
+    /// # Returns
+    /// A vector of (index_key, score) pairs, or None if no events found
+    #[tracing::instrument(name = "retry.index.fetch_ready", skip_all)]
+    pub async fn fetch_ready(
```
ok300, line 142: The return type can be simplified, as no caller makes a distinction between `None` vs empty vector: #828

nexus-watcher/src/events/mod.rs
ok300: I think the current approach of retrying and logging them is good enough for now.

nexus-watcher/tests/utils/mod.rs
```diff
@@ -0,0 +1,38 @@
+use nexus_common::models::event::{Event, EventProcessorError};
+use nexus_watcher::events::{EventHandler, Moderation};
+use pubky_app_specs::PubkyId;
+use std::sync::Arc;
+
+/// Mock implementation of EventHandler for testing.
+///
+/// If `target_uri_substring` is set, `result` only applies to events whose URI contains
+/// the substring; all other events return `Ok(())`.
+pub struct MockEventHandler {
```
ok300, line 11: This can be simplified: #829

The `moderation()` fn was not used by any callers, so it was removed.

nexus-watcher/src/events/retry/event.rs
```diff
@@ -31,86 +36,132 @@ impl RedisOps for RetryEvent {
...
 
         Ok(())
     }
 
     /// Checks if a specific event exists in the Redis sorted set
     /// # Arguments
-    /// * `event_index` - A `&str` representing the event index to check
-    pub async fn check_uri(event_index: &str) -> Result<Option<isize>, EventProcessorError> {
+    /// * `index_key` - A `&str` representing the index key to check
+    pub async fn check_uri(index_key: &str) -> Result<Option<isize>, EventProcessorError> {
```
ok300: This is only used in tests and can be simplified: #830

nexus-watcher/src/events/handlers/universal_tag.rs
ok300, line 1: The entire handler is already bypassed, so it can be removed: #831

nexus-common/src/universal_tag/homeserver_parsed_uri.rs
```diff
@@ -0,0 +1,243 @@
...
+            return Ok(HomeserverParsedUri::AppSpec {
+                user_id: parsed_uri.user_id,
+                resource: parsed_uri.resource,
+            });
+        }
+
+        // If ParsedUri::try_from failed, the URI might be from a different app.
+        // Try parsing as a universal tag URI: pubky://<user_id>/pub/<app>/tags/<tag_id>
+        // We need to manually parse the URI for non-pubky.app apps.
+        let parsed_url = url::Url::parse(uri).map_err(|e| format!("Invalid URL: {}", e))?;
```
ok300: A lot of this parsing is already done in `try_parse_app_tag_path`, so we can re-use it: #832

nexus-watcher/src/events/mod.rs
```diff
@@ -10,14 +11,40 @@ pub mod retry;
...
-    }?;
+/// Trait for handling events.
+///
+/// This trait abstracts event handling logic to allow for flexible implementations,
+/// including mocked versions for testing.
+#[async_trait::async_trait]
+pub trait EventHandler: Send + Sync {
+    /// Handle an event.
+    ///
+    /// Returns `Ok(())` on success, or `Err(EventProcessorError)` on failure.
```
ok300:
```suggestion
```

Generic comment, brings no more useful info than can be deduced from the method signature.

Should be removed.

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -0,0 +1,235 @@
...
+                    retry_event.event_uri, e
+                );
+                self.store.remove(index_key).await?;
+            }
+            Err(e) if e.is_infrastructure() => {
+                // Infrastructure errors (Neo4j/Redis failures) must NOT count against the
+                // application-level max_retries limit.  Reschedule with backoff but do NOT
+                // increment retry_count, then propagate to stop the current batch.
+                self.reschedule(&retry_event, index_key, &e, false).await?;
+                return Err(e);
```
ok300, line 153: Infrastructure errors from reschedule mask the original error

At line 150, when an infrastructure error occurs, `reschedule` is called first, and then `return Err(e)` propagates the original error. However, `reschedule` itself uses `?`, so if the Redis write in `reschedule` fails, that error propagates instead of the original infrastructure error. The original error context is lost.

A simple solution is proposed here: #835

nexus-watcher/src/events/mod.rs
```diff
@@ -72,9 +85,18 @@ pub async fn handle_put_event(
         }
         (PubkyAppObject::Tag(tag), Resource::Tag(tag_id)) => {
```
ok300, line 83: Moderation applied to universal tags may fail for non-pubky.app tagged URIs

Here moderation is checked before routing to the universal tag handler. If a moderator tags an universal tag with a moderated label, `apply_moderation` (2 lines below) calls `ParsedUri::try_from(moderator_tag.uri.as_str())` which would fail for non-pubky.app URIs. This would produce a `Generic` error that gets retried and eventually dead-lettered.

Looks like this bug existed before, added PR to fix it: #837

ok300 (approved):

PR #803, opened 2026-04-09
feat: post events handlers partial recovery

aintnostressin:
Prepares POST handlers for partial recovery.

  - PUT (sync_put): When a previous attempt wrote the graph node but failed before persisting Redis index entries, the retry now detects the missing PostDetails in Redis and runs recover_post_index_state to rebuild index state from graph truth     
  (details, relationships, counts, mention edges). Notifications are intentionally skipped on recovery (0 > N duplicates).
  - DEL (sync_del): Reordered to graph-last: PostRelationships is deleted first as a gate, so retries observe post_in_index = false and skip non-idempotent ops (counter decrements, notifications, score adjustments). On retry, parent URIs are       
  fetched from the still-present graph node. The graph node is deleted only after all Redis cleanup succeeds.                                                                                                                                           
  - Refactors: Extracted find_mentioned_ids and merge_mention_edges helpers for reuse in recovery. Renamed PostDetails::delete to delete_from_index and removed its graph deletion side-effect to support the graph-last invariant. Changed
  parent_post_key_wrapper from [String; 2] to (String, String).

PR #811, opened 2026-04-21
Simplify parallel ops

ok300:
This PR consolidates and simplifies some of the parallel ops introduced in #803.


aintnostressin (approved):
I'm approving it as this is functionally the same but I prefer current way:

1. this leaks a bit of details about redis internals and i would prefer to not have this kind of information at this level of abstraction (handler)
2. these two operations have no logical dependency on each other so from the caller perspective it is fine to call them concurrently
3. if both are executed at the same time then two requests are in flight (small perf gains)

ok300:
Alright then let's leave this as it is.

PR #816, opened 2026-04-21
refactor: create universal tag module

aintnostressin:
Extracts Universal Tags related logic to a nexus-common module.


nexus-common/src/models/universal_tags/homeserver_parsed_uri.rs
```diff
@@ -20,9 +21,8 @@ pub enum HomeserverParsedUri {
     /// Format: pubky://<user_id>/pub/<app>/tags/<tag_id>
     UniversalTag {
         user_id: PubkyId,
-        app: String,
         resource: Resource,
-        tag_id: String,
+        tag: UniversalTag,
```
ok300, line 24: Doesn't this add confusion, introducing a `UniversalTag` struct when there is already a `UniversalTag` enum type? And one contains the other?

Why not keep the `UniversalTag` enum are it was before?

nexus-common/src/models/event/homeserver_parsed_uri.rs
ok300, line 14: Technically this would also belong in the new module, as it's strictly related to parsing the new URI.

Essentially this is a `ParsedUriV2` of `pubky-app-spec`'s `ParsedUri`.

I'd say the entire `homeserver_parsed_uri.rs` belongs in the new module.

nexus-watcher/src/events/mod.rs
```diff
@@ -97,9 +97,17 @@ pub async fn handle_put_event(
             } else {
                 // Route universal tag events (non-pubky.app apps) to sync_put_resource
                 // which handles Resource nodes for InternalUnknown/InternalUnknown URIs.
-                if let HomeserverParsedUri::UniversalTag { app, .. } = &event.parsed_uri {
-                    handlers::tag::sync_put_resource(tag, user_id, tag_id.to_string(), app.clone())
-                        .await?
+                if let HomeserverParsedUri::UniversalTag {
```
ok300, line 101: The nested branching here in `handle_put_event` is already pretty hard to follow. This change seems to have made it even more verbose for no obvious benefit. I'd keep this unchanged if possible. (A)

nexus-watcher/src/events/mod.rs
```diff
@@ -136,19 +144,8 @@ pub async fn handle_del_event(event: &Event) -> Result<(), EventProcessorError>
...
-            } = &event.parsed_uri
-            {
-                let info = handlers::universal_tag::AppTagInfo {
-                    user_id,
-                    app: app.clone(),
-                    tag_id: utag_id.clone(),
-                    uri: event.uri.clone(),
-                };
-                handlers::universal_tag::handle_del(info).await?
+            if let HomeserverParsedUri::UniversalTag { tag, .. } = &event.parsed_uri {
```
ok300: Same as (A)

nexus-common/src/models/event/homeserver_parsed_uri.rs
aintnostressin: this is  a wider concept than universal tags thats why i decided to leave it as it is and extract only things related to universal tags.

In this PR i'm extracting `universal_tag` module so my goal is to keep it focused on that part of the domain.

nexus-common/src/models/universal_tags/homeserver_parsed_uri.rs
```diff
@@ -20,9 +21,8 @@ pub enum HomeserverParsedUri {
...
     },
 }
```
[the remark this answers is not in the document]
aintnostressin: I've removed redundant user_id field.
I decided to keep tag so reference to it can be returned by resource fn.

nexus-watcher/src/events/mod.rs
aintnostressin: This one is actually more compact

nexus-common/src/models/universal_tags/homeserver_parsed_uri.rs
aintnostressin: I gave it a second look and it looks reasonable to me.

nexus-common/src/models/event/homeserver_parsed_uri.rs
ok300: The entire file's purpose is parsing the URI. This logic was not needed before Universal Tags and is therefore linked to it.

As mentioned above
> Essentially this is a `ParsedUriV2` of `pubky-app-spec`'s `ParsedUri`.

So if the goal of this PR is to extract Universal Tag -related code in a separate module, such that a possible migration of that code to `pubky-app-specs` is made easier, it follows that this is part of that.

nexus-common/src/models/universal_tags/homeserver_parsed_uri.rs
ok300: I don't see how introducing a 2nd struct with the same name is reasonable.

Especially in a PR that is meant to move code to a separate module.

Please leave this unchanged and lets focus on cleanly separating the Universal Tags -related code.

nexus-common/src/models/universal_tags/homeserver_parsed_uri.rs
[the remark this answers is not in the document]
ok300: Such optimizations and refactorings should be done in a separate PR, if at all.

When done here, they only cause: more verbose diff, higher chance of conflicts when rebasing, higher complexity review, higher chance of bugs slipping through because of unnecessarily increased delta to `main`.

See comment above, please keep this focused on cleanly moving the Universal Tag code to its own module.

nexus-watcher/src/events/mod.rs
ok300: Lets keep the changes here strictly related to extracting a Universal Tag module, which can potentially be moved to `pubky-app-specs`. Doing this is already quite tricky, adding unrelated changes only makes things unnecessarily harder.

ok300 (changes requested):
Other structures that were added with Universal Tags, are conceptually bound to it, are not Nexus-specific, and therefore belong to the new module:

- `UriCategory` enum + `classify_uri()` fn
- `normalize_uri()` + `normalize_parsed_uri()`
- `resource_id()`
- `HomeserverParsedUri`, which is a newer iteration of the existing `pubky-app-spec` 's `ParsedUri` , duplicating much of its internal parsing logic

ok300:
Superseded by #820

PR #819, opened 2026-04-23
fix: handle attachment-only post edits

### LINKED ISSUE #818: Detect attachment change in updated post
**Problem:**
We have work in https://github.com/pubky/pubky-app/pull/1311 to edit attachments in posts and the cover image for articles.
Testing has exposed that when pubky-app does the following: deletes attachment file and blob, put new file and blob, overwrite put the post with correct file attachment reference, then nexus still serves up the original post referencing the blob that is now deleted so it cannot be retrieved.

**Proof**
Here nexus is only considering posts as 'edited' if the text content has changed.
https://github.com/pubky/pubky-nexus/blob/45e2c51e730c88f40a0d20bf4d76db4b2876effd/nexus-watcher/src/events/handlers/post.rs#L69-L71

**Solution**
Also check whether the attachment has changed to run the sync edit.

ok300:
This PR fixes the way a post edit was detected. Now the logic checks the post attachments, as well.

Fixes #818

### DIFF
--- a/nexus-common/src/models/post/details.rs
+++ b/nexus-common/src/models/post/details.rs
@@ -10,7 +10,8 @@ use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
 /// Represents post data with content, bio, image, links, and status.
-#[derive(Serialize, Deserialize, ToSchema, Default, Debug, PartialEq, Clone)]
+#[derive(Serialize, Deserialize, ToSchema, Default, Debug, PartialEq)]
+#[cfg_attr(test, derive(Clone))]
 // NOTE: Might not be necessary the default values for serde because before PUT a PostDetails node
 // we do sanity check
 pub struct PostDetails {
--- a/nexus-common/src/models/post/details.rs
+++ b/nexus-common/src/models/post/details.rs
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
 /// Represents post data with content, bio, image, links, and status.
-#[derive(Serialize, Deserialize, ToSchema, Default, Debug, PartialEq)]
+#[derive(Serialize, Deserialize, ToSchema, Default, Debug, PartialEq, Clone)]
 // NOTE: Might not be necessary the default values for serde because before PUT a PostDetails node
 // we do sanity check
 pub struct PostDetails {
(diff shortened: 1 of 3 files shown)


nexus-watcher/src/events/handlers/post.rs
```diff
@@ -245,6 +245,11 @@ pub async fn sync_put(
     Ok(())
 }
 
+fn post_details_changed(existing_details: &PostDetails, post_details: &PostDetails) -> bool {
```
aintnostressin: This name may be misleading - Have you though about something more general ?

aintnostressin: I suggest moving this function to `PostDetails` impl , it can take a reference to other PostDetails and compare .

The callsite could look like this `let was_changed = existing_details.was_changed(&post_details)`

nexus-watcher/src/events/handlers/post.rs
```diff
@@ -245,6 +245,11 @@ pub async fn sync_put(
...
+    existing_details.content != post_details.content
+        || existing_details.attachments != post_details.attachments
```
aintnostressin: Should we treat attachments order change as a change ?

nexus-watcher/tests/event_processor/posts/attachments.rs
```diff
@@ -67,3 +67,121 @@ async fn test_homeserver_post_attachments() -> Result<()> {
 
     Ok(())
 }
+
+#[tokio_shared_rt::test(shared)]
+async fn test_homeserver_post_attachment_only_edits() -> Result<()> {
```
aintnostressin, line 72: I suggest creating a separate file containing two tests cases that verify event handling detects:

1. details change
2. attachments change

nexus-watcher/src/events/handlers/post.rs
ok300: Good idea, done: 66921431880d8c96a56575fb502557614c6acac3

ok300: Good point.

Combined with the move to `PostDetails`, I think this is a good name: `PostDetails::is_different_than()`

Done: 66921431880d8c96a56575fb502557614c6acac3

nexus-watcher/tests/event_processor/posts/attachments.rs
ok300: Done: d60979a5bf555c38b7015586640df804aa389019

nexus-watcher/src/events/handlers/post.rs
ok300: Good question.

The most obvious answer seems to me that yes, changing the order of attachments should register as an edit. For example a post that has a few images, changing their order could very well change the meaning of the post.

Covered by the `PostDetails` tests: d60979a5bf555c38b7015586640df804aa389019

nexus-common/src/models/post/details.rs
```diff
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 use utoipa::ToSchema;
 
 /// Represents post data with content, bio, image, links, and status.
-#[derive(Serialize, Deserialize, ToSchema, Default, Debug, PartialEq)]
+#[derive(Serialize, Deserialize, ToSchema, Default, Debug, PartialEq, Clone)]
```
aintnostressin: If `clone` is needed only for tests could you please keep it for tests only or add a helper fn in tests ?

ok300: Done: 6d601f17da8fff032633433c65852d5a31454e0b

aintnostressin (approved):

PR #820, opened 2026-04-23
Move Universal Tags structs to separate module

ok300:
This PR moves structs and utils related to Universal Tag functionality in a separate module.

At a later point, we can decide if they should be extracted to `pubky-app-specs`.

Related to #806


aintnostressin (approved):
looks ok!

PR #834, opened 2026-04-24
test: add more test cases for multi event handling

aintnostressin:
This PR  adds new tests cases and assertions to both `HsEventProcessor` and `RetryProcessor` tests.

* assert how many event handler was called
* verify continue / stop processing in case of various event handler errors


ok300 (approved):
Other than the merge conflicts, LGTM.

PR #835, opened 2026-04-25
fix: infra errors from `reschedule` mask the original error

ok300:
The handling of infrastructure errors requires

```rust
self.reschedule(&retry_event, index_key, &e, false).await?;
```

which itself can fail due to the same infrastructure error that caused it. This can mask the error being handled, which is supposed to be re-thrown on the line below.

This PR pre-emptively logs errors before they are being classified and handled.

### DIFF
--- a/nexus-watcher/src/events/retry/processor.rs
+++ b/nexus-watcher/src/events/retry/processor.rs
@@ -120,10 +120,17 @@ impl RetryProcessor {
             }
         };
 
-        // Call event_handler directly to get the actual error (bypassing handle_event/handle_error)
         let ev_uri = &retry_event.event_uri;
         let ev_retry_count = retry_event.retry_count;
-        match self.event_handler().handle(&event).await {
+
+        // Call event_handler directly to get the actual error (bypassing handle_event/handle_error)
+        let event_handle_res = self.event_handler().handle(&event).await.inspect_err(|e| {
+            // In case of error, log it before the error is itself is classified and handled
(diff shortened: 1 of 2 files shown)


nexus-watcher/src/events/retry/processor.rs
```diff
@@ -120,27 +120,30 @@ impl RetryProcessor {
...
+                debug!("Retry successful for event: {ev_uri}");
                 self.store.remove(index_key).await?;
             }
             Err(e) if e.is_404() => {
                 // Content gone - remove from retry queue
-                warn!(
-                    "Content no longer exists (404) for retry: {}",
-                    retry_event.event_uri
-                );
+                warn!("Content no longer exists (404) for retry: {ev_uri}");
```
aintnostressin: If we log every error then this line makes a little sense now - can you please remove it ?

nexus-watcher/src/events/retry/processor.rs
aintnostressin: could you extract same helper methods for these ?

nexus-watcher/src/events/retry/processor.rs
ok300: Done: 3372e538fcc7810a963c9af01265e32d6cf13725

nexus-watcher/src/events/retry/processor.rs
ok300: Done: 13022149bc75dc6e7b6dd2e7d528709e3f861f7a

aintnostressin (approved):

PR #837, opened 2026-04-25
fix: add moderation of Universal Tags + integration test

ok300:
This PR fixes moderation of Universal Tags.

---

`apply_moderation` silently failed when the moderation tag's `uri` pointed to a Universal Tag (e.g. `pubky://<user_id>/pub/mapky/tags/<tag_id>`). `ParsedUri::try_from` rejects any URI where the app segment isn't `pubky.app`, so moderating Universal Tags was broken with no test coverage catching it.

## Changes

- **`universal_tag.rs`** — make `try_parse_app_tag_path` `pub` so moderation can reuse it.
- **`moderation.rs`** — in `apply_moderation`, check for a Universal Tag URI *before* calling `ParsedUri::try_from`. When matched, delegate directly to `handlers::tag::del`:

  ```rust
  if let Some(info) = try_parse_app_tag_path(moderator_tag.uri.as_str()) {
      info!("Moderation tag '{}' detected. Deleting universal tag {}:{}", ...);
      return handlers::tag::del(info.user_id, info.tag_id).await;
  }
  // existing ParsedUri::try_from path follows
  ```

- **`tags/moderated.rs`** *(new)* — integration test `test_moderated_universal_tag_lifecycle`: creates a Universal Tag at `/pub/mapky/tags/<tag_id>`, verifies it exists, moderator tags its own URI with the moderation label, asserts the Universal Tag is deleted.

### DIFF
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -128,10 +128,15 @@ pub fn get_post_replies(author_id: &str, post_id: &str) -> Query {
 
 // Read the target details for a tag without deleting the TAGGED edge.
 // Used in tag del to read before graph-last deletion.
-pub fn get_tag_target(user_id: &str, tag_id: &str) -> Query {
-    Query::new(
-        "get_tag_target",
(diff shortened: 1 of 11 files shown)


ok300:
Superseded by #850

PR #838, opened 2026-04-25
Remove `files_path` from `Event` struct

ok300:
This PR removes the `Event.files_path` field, which conceptually doesn't belong to an `Event`.

### DIFF
--- a/nexus-common/src/config/stack.rs
+++ b/nexus-common/src/config/stack.rs
@@ -1,4 +1,4 @@
-use crate::{db::DatabaseConfig, get_files_dir_pathbuf};
+use crate::{db::DatabaseConfig, get_files_dir_pathbuf, get_files_dir_test_pathbuf};
 use serde::{Deserialize, Deserializer, Serialize};
 use std::{fmt::Debug, path::PathBuf};
 
@@ -52,6 +52,15 @@ impl Default for StackConfig {
             files_path: get_files_dir_pathbuf(),
             otlp: OtlpConfig::default(),
             db: DatabaseConfig::default(),
+        }
+    }
+}
+
+impl StackConfig {
+    pub fn test_default() -> Self {
+        Self {
+            files_path: get_files_dir_test_pathbuf(),
+            ..Default::default()
         }
     }
 }
--- a/nexus-watcher/src/events/context.rs
+++ b/nexus-watcher/src/events/context.rs
@@ -1,0 +1,26 @@
+use std::path::PathBuf;
+
+use nexus_common::WatcherConfig;
+
+use super::Moderation;
+
+pub struct EventContext {
+    pub moderation: Moderation,
+
+    /// Nexus-local directory path where certain events may store data on Nexus.
+    ///
+    /// Moderation may trigger the deletion of such data for specific events.
+    pub files_path: PathBuf,
+}
+
+impl EventContext {
+    pub(crate) fn from_config(config: &WatcherConfig) -> Self {
+        EventContext {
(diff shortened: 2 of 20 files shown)


ok300:
@greptile please review again

nexus-watcher/tests/event_processor/files/delete.rs
```diff
@@ -122,7 +122,7 @@ async fn test_delete_pubkyapp_file_is_idempotent() -> Result<()> {
 
     // Simulate a replay: call the DEL handler again directly (e.g. crash between FS delete and store_event)
     let user_pubky_id = PubkyId::try_from(user_id.as_str()).map_err(anyhow::Error::msg)?;
-    let result = handlers::file::del(&user_pubky_id, file_id, get_files_dir_test_pathbuf()).await;
+    let result = handlers::file::del(&user_pubky_id, file_id, &get_files_dir_test_pathbuf()).await;
```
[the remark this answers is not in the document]
ok300: It compiles just fine.

nexus-common/src/config/stack.rs
```diff
@@ -55,3 +55,12 @@ impl Default for StackConfig {
         }
     }
 }
+
+impl StackConfig {
+    pub fn test_default() -> Self {
```
aintnostressin, line 60: If it's `test_default` can we add `#[cfg(test)]` ?

ok300: I tried, but unfortunately it's not valid across crates.

`nexus-common` (where `StackConfig` is located) is a dependency of `nexus-watcher`. The `#[cfg(test)]` annotation would only be visible inside `nexus-watcher`. Such test-only blocks in dependencies are not visible to (test) callers from other crates.

SHAcollision (approved):
LGTM!

2 small NITS: Moderation is now misnamed / mixed-purpose. It now carries `files_path` purely so DEL handlers can find it — but file deletion happens for non-moderation DEL events too (handle_del_event → handlers::file::del). Storing a generic filesystem path on a struct called "Moderation" is a leaky abstraction.
Consider either:
    - renaming to something like EventContext / HandlerContext, or
    - keeping Moderation pure and introducing a small FilesContext (also Arc-shared) passed alongside it.

Second NIT: Mixed `PathBuf` / `&Path`. `handlers::file::sync_put` still takes an owned `PathBuf`. If we're already changing `del` to `&Path`, doing the same to `sync_put` would eliminate that clone and keep the codebase consistent. Worth a follow-up if not worth changing here.

ok300:
> Second NIT: Mixed `PathBuf` / `&Path`.

Done: dfda91c8d803e2388a6e077921e0bb7a1dbaca19

ok300:
> Moderation is now misnamed / mixed-purpose.
> Consider (..) `EventContext`

Done: a63048665ee34be04d063477aa04475532c366a4

nexus-watcher/src/events/context.rs
```diff
@@ -0,0 +1,26 @@
+use std::path::PathBuf;
+
+use nexus_common::WatcherConfig;
+
+use super::Moderation;
+
+pub struct EventContext {
```
aintnostressin, line 7: There is already such context https://github.com/pubky/pubky-nexus/blob/feat/dx-events-by-user/nexus-watcher/src/events/mod.rs#L24-L26

I suggest applying this fix to that branch and reusing new things.

SHAcollision (approved):

ok300:
Superseded by #990

PR #839, opened 2026-04-25
fix: scope tag deletion to app namespace to prevent cross-app interference

ok300:
When a user has TAGGED relationships with the same `tag_id` across different app namespaces (possible when two apps store identical tag content, producing the same hash-derived ID), `get_tag_target` returned an arbitrary match and deletion could operate on the wrong relationship.

## Root cause

`get_tag_target` matched solely on `(user_id, tag_id)` with no `app` filter, while `delete_tag` already had the correct `WHERE tag.app = $app` / `WHERE tag.app IS NULL` scoping. `universal_tag::handle_del` also never forwarded the known `app` name down to `tag::del`.

## Changes

- **`nexus-common` — `get_tag_target`**: add `app: Option<&str>` parameter; inject `WHERE tag.app = $app` (universal) or `WHERE tag.app IS NULL` (standard pubky.app), consistent with `delete_tag`.
- **`nexus-watcher` — `tag::del`**: accept `app: Option<String>`, thread it into `get_tag_target`.
- **`universal_tag::handle_del`**: pass `Some(info.app)` instead of dropping the known app.
- **`handle_del_event` / `moderation.rs`**: pass `None` — standard pubky.app tags carry no app property.
- **`del_idempotent.rs`**: update direct `tag::del` call sites to pass `None`.

## Test — `resource_multi_app_del.rs`

Adversarial scenario: one user creates 3 tags sharing the same label —
- 1 standard pubky.app user tag
- 2 universal tags (`mapky`, `eventky.app`) with **identical content → same `tag_id`**

Verifies that deleting each tag removes only its own `TAGGED` relationship and leaves the other two intact.

### DIFF
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -128,10 +128,21 @@ pub fn get_post_replies(author_id: &str, post_id: &str) -> Query {
 
 // Read the target details for a tag without deleting the TAGGED edge.
 // Used in tag del to read before graph-last deletion.
-pub fn get_tag_target(user_id: &str, tag_id: &str) -> Query {
-    Query::new(
-        "get_tag_target",
-        "MATCH (user:User {id: $user_id})-[tag:TAGGED {id: $tag_id}]->(target)
+//
+// `app` is used to scope the lookup to a specific app namespace.  When `app`
+// is `Some`, only the TAGGED relationship whose `app` property equals that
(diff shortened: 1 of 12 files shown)


ok300:
@greptile please review again

ok300:
@greptile please review again

ok300:
Needs further clarification if tag deletion is app-scoped and/or user-scoped.

Will be covered in #837

PR #843, opened 2026-04-27
feat: check event user id

aintnostressin:
Ensure events returned from homeserver point to requested user's resources.
If there is a mismatch, stop processing events for that user.

Without this check, a misbehaving homeserver could trigger PUT/DEL operations against users it doesn't host.

### DIFF
--- a/nexus-watcher/src/service/indexer/key_based.rs
+++ b/nexus-watcher/src/service/indexer/key_based.rs
@@ -1,19 +1,40 @@
 use std::path::PathBuf;
-use std::sync::Arc;
-
-use futures::StreamExt;
-use nexus_common::db::{PubkyConnector, RedisOps};
-use nexus_common::models::event::{Event, EventProcessorError};
-use nexus_common::models::homeserver::Homeserver;
-use nexus_common::models::user::{user_hs_cursor_key, UserDetails};
-use pubky::{EventCursor, PublicKey};
-use tokio::sync::watch::Receiver;
-use tracing::{debug, error, info};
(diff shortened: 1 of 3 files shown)


nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -165,7 +169,15 @@ impl KeyBasedEventProcessor {
 
                 match Event::from_stream_event(&stream_event, self.files_path.clone()) {
```
ok300, line 170: This returns `Ok(None)` for skipped resources (`LastRead`, `Feed`, etc). In this case, the user check is skipped. However, such events with an incorrect user would still count towards the cursor advancement, so a malicious HS could still return such events to confuse the cursor calculation.

nexus-watcher/src/service/indexer/key_based.rs
ok300, line 161: We also don't need the fancy stream. We could iterate over the stream to get a vector of the events, then first iterate over them and validate users (if any incorrect, throw error), then iterate again to process them.

This would bypass the cursor advancement topic above.

I think that would be the MVP of this check. Everything else around seems to me over-engineered. Custom error-type, metric, etc.

nexus-watcher/src/service/indexer/key_based.rs
aintnostressin: I think we cannot do much about it - if HS behaves bad user should change it. 
The goal of this PR is to prevent malicious HS from altering other users state.
Do you think we can achieve more?

PR #845, opened 2026-04-28
fix: validate user ID before handling event in key-based processor

ok300:
This PR offers a simpler alternative to #843 , to validate user IDs before processing events from 3rd party HSs.

### DIFF
--- a/nexus-watcher/src/service/indexer/key_based.rs
+++ b/nexus-watcher/src/service/indexer/key_based.rs
@@ -189,6 +189,7 @@ impl KeyBasedEventProcessor {
             .await?;
 
         let user_id = user_pk.z32();
+        let expected_user_id = PubkyId::from(user_pk.clone());
         let mut latest_cursor: Option<u64> = None;
 
         let result: Result<(), EventProcessorError> = async {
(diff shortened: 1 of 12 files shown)


nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -204,6 +208,25 @@ impl KeyBasedEventProcessor {
         result
     }
 
+    fn validate_user_id(
+        hs_id: &str,
+        event: &Event,
+        expected_user_pk: &PublicKey,
```
aintnostressin: Have you considering passing `PubkyId` here so you can compare it directly with `event.parsed_uri.user_id()` ?

aintnostressin (commented):
It would be good to have test asserting that:

* processor doesn't process any further events for this user if there is user id mismatch
* processor still tries to process events for other users from that homeserver

nexus-watcher/src/service/indexer/key_based.rs
ok300: Done: 5a69c20b3ec72ec2d0e67a5472eaca5bacd0dc07

aintnostressin (approved):

ok300:
@greptile please review again

ok300:
@greptile please review again

ok300:
> It would be good to have test asserting that:
> 
> * processor doesn't process any further events for this user if there is user id mismatch
> * processor still tries to process events for other users from that homeserver

Done: ee27da8d2b3973600ead7b96deabde8d41cc3186

aintnostressin (approved):

PR #848, opened 2026-04-29
refactor: consolidate event parsing

ok300:
This PR refactors event parsing to share URI/resource classification between homeserver event lines and stream events.

`Event::build` now returns `ParseResult`, allowing `parse_event` to delegate common parsing logic directly while `from_stream_event` preserves its existing `Option<Event>` API by mapping skipped or unrecognized events to `None`. 

This removes duplicated match logic and keeps the public method signatures unchanged.

### DIFF
--- a/nexus-common/src/models/event/mod.rs
+++ b/nexus-common/src/models/event/mod.rs
@@ -137,7 +137,7 @@ impl Event {
         // standard pubky-app-specs URIs and universal tag URIs from other apps.
         let parsed_uri = match HomeserverParsedUri::try_from(uri.as_str()) {
             Ok(parsed) => parsed,
-            Err(e) => return Ok(ParseResult::unrecognized_uri(event_type, uri, e.into())),
+            Err(e) => return Ok(ParseResult::unrecognized_uri(event_type, uri, e)),
         };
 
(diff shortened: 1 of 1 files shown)


nexus-common/src/models/event/mod.rs
```diff
@@ -154,18 +121,30 @@ impl Event {
...
     fn build(
         event_type: EventType,
         uri: String,
         event_line: String,
         files_path: PathBuf,
-    ) -> Result<Option<Self>, EventProcessorError> {
-        let parsed_uri = HomeserverParsedUri::try_from(uri.as_str()).map_err(|e| {
-            EventProcessorError::InvalidEventLine(format!("Cannot parse event URI: {e}"))
-        })?;
+    ) -> Result<ParseResult, EventProcessorError> {
```
aintnostressin, line 135: This is really unusual. Build functions are meant to return Self. In this case this is misleading. Have you considered changing fn name ?

ok300: Done: fcf76842bdc4b0144d104b2e52432297c4a43a1f

aintnostressin (approved):

Issue #579, opened 2025-09-20
feat: validators


tipogi:
Before handling a request, it is important to validate at least some received arguments. A common example of this is arguments related to the `PubkyId`. By performing validation at this early stage, we can prevent unnecessary database queries. This approach also allows us to provide more meaningful and scoped error messages to the client. For instance, instead of returning the generic _User not found_, we can explicitly report issues such as _Wrong PubkyId format in [argument_name]_. This not only improves the accuracy of error reporting but also helps clients correct their requests more easily and reduces wasted processing on the server side

aintnostressin:
Closed by related issue.

PR #849, opened 2026-05-01
feat: validate api inputs

aintnostressin:
Replace free-floating String parameters across HTTP handlers with validated newtypes, pushing input validation to deserialization.                                                                                                                    
   
  Changes                                                                                                                                                                                                                                               
                            
  - New input types in nexus-webapi/src/models/: PubkyId, PostId, ResourceId, TagLabel, Tags (CSV), UserIds/PostIds/GlobalPostIds (JSON arrays + CSV), UserIdPrefix, UsernamePrefix. Each enforces format/length rules at deserialize time.             
  - bounded_vec helpers (deserialize_csv, deserialize_json_array) for reusable MIN..=MAX length validation on collection inputs.
  - Path<T> / Json<T> wrappers in routes/mod.rs map extractor rejections to Error::InvalidInput (400) with a JSON body, instead of axum's default plain-text rejection.                                                                                 
  - StreamSourceKind split from nexus-common::StreamSource for posts: cross-field requirements (post_replies needs author_id + post_id, etc.) now return 400 instead of silently falling back to All.                                                   
  - Removed dead Error::BookmarksNotFound and the inline validate_resource_id / sanitize_validate helpers superseded by the newtypes.                                                                                                                   
                                                                                                                                                                                                                                                        
  Behavior changes worth noting                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                        
  - ?tags= (empty value) now returns 400 instead of being treated as None.                                                                                                                                                                              
  - Stream endpoints with missing required parameters return 400 instead of falling back to the global feed.
  - Username/ID prefix validation no longer trims input - what you send is what gets validated and queried.  

related: #579


docker/test-graph/mocks/posts.cypher
```diff
@@ -35,91 +35,91 @@ MATCH (u1:User {id: $eixample}), (u2:User {id: $detroit}) MERGE (u1)-[:FOLLOWS {
 // ###############################
 // ##### Posts related tags ######
 // ###############################
-MERGE (p:Post {id: "V8N1P3L9J4X0"}) SET p.content = "API security is crucial for ensuring user privacy and data protection", p.kind = "short", p.indexed_at = 1709308315917;
```
ok300, line 38: Looks like this might require a migration.

If so, it would be helpful to document this in the PR description (incl. related infos like: what DB fields have to be migrated, which migrations have to be run, etc)

nexus-webapi/src/routes/v0/stream/posts.rs
```diff
@@ -11,12 +11,10 @@ use nexus_common::{
...
 use pubky_app_specs::PubkyAppPostKind;
-use serde::{de, Deserialize, Deserializer};
+use serde::Deserialize;
 use tracing::debug;
 use utoipa::{OpenApi, ToSchema};
 
-const MAX_TAGS: usize = 5;
-
 #[derive(Deserialize, Debug, ToSchema)]
 pub struct PostStreamQuery {
```
ok300, line 125: Post stream source params are documented as validated types but still deserialize through `StreamSource` as raw `Strings`.

`PostStreamQuery` flattens `Option<StreamSource>`, and `nexus-common/src/models/post/stream.rs:29-49` keeps `author_id`, `observer_id`, and `post_id` as `String`. Invalid values like `source=post_replies&author_id=bad&post_id=bad` bypass the new `PubkyId`/`PostId` validators.

nexus-webapi/src/models/global_post_id.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Consumes self and returns a `Vec<String>` of the underlying ID strings.
+    pub fn into_string_vec(self) -> Vec<String> {
+        self.0.into_iter().map(|id| id.0).collect()
+    }
+}
+
+impl GlobalPostId {
+    fn validate(s: &str) -> Result<(), String> {
+        let (pubky_part, post_part) = s.split_once(':').ok_or_else(|| {
+            "Validation Error: GlobalPostId must be in the format '{PubkyId}:{PostId}'".to_string()
```
ok300: Since a lot of validation checks are added, it might make sense to add a `Validation` error enum type.

The error messages then don't need to have the prefix "Validation Error: ...".

nexus-webapi/src/models/post_id.rs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        let s: String = String::deserialize(deserializer)?;
+        Self::try_from(s).map_err(de::Error::custom)
+    }
+}
+
+impl PostId {
+    /// Validates that the provided ID is a valid Crockford Base32-encoded timestamp,
+    /// 13 characters long, and decodes to 8 bytes.
+    fn validate_id(id: &str) -> Result<(), String> {
```
ok300: Looks like a replica of `TimestampId::validate_id` from `pubky-app-specs`. This is used by `PubkyAppPost`.

Have you tried re-using that instead?

nexus-webapi/src/models/pubky_id.rs
```diff
@@ -0,0 +1,66 @@
+use std::fmt;
+use std::ops::Deref;
+
+use serde::{de, Deserialize};
+use utoipa::ToSchema;
+
+#[derive(ToSchema, Debug)]
+pub struct PubkyId(pub pubky_app_specs::PubkyId);
```
ok300: I would avoid using the same struct name as already existing structs, since some callers can easily confuse them (for example, utoipa swagger annotation types).

In fact why not use a naming pattern for these results of `path_extractor_impl`, like `PathExtractedPubkyId` here?

(A)

nexus-webapi/src/models/bounded_vec.rs
```diff
@@ -0,0 +1,298 @@
...
+/// define_bounded_vec!(
+///     name: UserIds,
+///     element_type: PubkyId,
+///     min: 1,
+///     max: 100,
+///     serialize_as: json_array,
+/// );
+/// ```
+#[macro_export]
+macro_rules! define_bounded_vec {
```
ok300: Similar to (A), why not use a naming scheme like `Validated...` for the generated types?

For example
```
/// define_bounded_vec!(
///     name: ValidatedPostIds,  <-- (based on element_type below)
///     element_type: PostId,
```

A scheme like this could probably be also enforced or even generated by the macro.

nexus-webapi/src/models/resource_id.rs
```diff
@@ -0,0 +1,162 @@
...
+        f.write_str(&self.0)
+    }
+}
+
+impl ResourceId {
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    fn validate_id(id: &str) -> Result<(), String> {
```
ok300: Looks like a candidate for #806 , as 3rd party apps might want to do similar validations.

Just FYI, no action needed here for now.

nexus-webapi/src/routes/v0/post/bookmark.rs
```diff
@@ -18,25 +20,24 @@ pub struct PostQuery {
...
+        ("post_id" = PostId, Path, description = "Post Crockford32 ID"),
+        ("viewer_id" = Option<PubkyId>, Query, description = "Viewer Pubky ID")
     ),
     responses(
         (status = 200, description = "Post Bookmark", body = Bookmark),
         (status = 404, description = "Post not found"),
         (status = 500, description = "Internal server error")
     )
 )]
+//todo: test it manually - no integration tests
```
ok300: Is this WIP or is there a major reason why a test can't be added?

nexus-webapi/tests/user/reach.rs
```diff
@@ -43,7 +43,10 @@ async fn test_get_followers() -> Result<()> {
 
     // Test non-existing user
     invalid_get_request(
-        &format!("/v0/user/{}/followers", "bad_user_id"),
+        &format!(
+            "/v0/user/{}/followers",
+            "4snwyct86m383rsduhw5xgcxpw7c63j3pq8x4ycqikxgik8y64rp"
```
ok300: If this is treated as "non-existing user" in several tests, it might make sense to extract it to a global `const` or similar, so further new tests can easily re-use it.

nexus-webapi/tests/utils/mod.rs
```diff
@@ -1,9 +1,33 @@
...
+}
+
+// #######################################
+// ##### Test data generators ########
+// #######################################
+
+/// Generates a valid PostId by encoding the given timestamp as Crockford Base32.
+/// The resulting ID is 13 characters long and decodes to 8 bytes, satisfying PostId validation.
+pub fn generate_post_id(timestamp: i64) -> String {
+    let id = encode(Alphabet::Crockford, &timestamp.to_be_bytes());
```
ok300: Why not generate a real `PubkyAppPost` and use its ID, instead of replicating the ID-generation logic?

nexus-webapi/src/models/mod.rs
```diff
@@ -1,5 +1,36 @@
...
+pub use global_post_id::{GlobalPostId, GlobalPostIds};
 pub use info::ServerInfo;
 pub use post::{PostStreamDetailed, PostViewDetailed};
+pub use post_id::{PostId, PostIds};
+pub use pubky_id::PubkyId;
+pub use resource_id::ResourceId;
+pub use tag_label::TagLabel;
+crate::define_bounded_vec!(
+    name: Tags,
+    element_type: String,
```
ok300: Switching the `element_type` to `TagLabel` would make the tag-label validation possible as well, in addition to count validation done here.

nexus-webapi/src/models/path_extractor.rs
```diff
@@ -0,0 +1,39 @@
...
+/// `crate::Error::InvalidInput`; server-error rejections (e.g. router misconfiguration)
+/// map to `crate::Error::InternalServerError`. `TryFrom` failures map to
+/// `crate::Error::invalid_input(&e.to_string())`.
+///
+/// # Usage
+/// ```ignore
+/// path_extractor_impl!(PostId);
+/// ```
+#[macro_export]
+macro_rules! path_extractor_impl {
```
ok300: This adds a custom single-path extractor macro, while multi-param paths already use `Path<T>` plus `Deserialize`. This creates two extraction styles.

A simpler solution would be to use `Path<PubkyId>`, `Path<TagLabel>`, etc. consistently, or centralize all validation in `serde`/`path` structs.

nexus-webapi/src/models/bounded_vec.rs
ok300: This is a large macro with duplicated implementations for CSV and JSON forms. This could be smaller as either normal `serde` helper functions (`bounded_csv`, `bounded_json_array`) or a const-generic `BoundedVec<T, MIN, MAX>` plus format-specific `serde` modules.

nexus-webapi/src/models/bounded_vec.rs
```diff
@@ -0,0 +1,298 @@
...
+                    if i > 0 {
+                        f.write_str(", ")?;
+                    }
+                    ::std::fmt::Display::fmt(item, f)?;
+                }
+                Ok(())
+            }
+        }
+
+        impl<'de> ::serde::Deserialize<'de> for $Name {
```
ok300: This deserializes JSON arrays through `serde_json::Value`, cloning the array before deserializing elements. This is heavier than necessary and couples the type to JSON. A `Vec<T>::deserialize` path with length validation after deserialization is simpler unless fail-fast length errors are a hard requirement.

nexus-webapi/tests/stream/user/list.rs
```diff
@@ -53,29 +53,38 @@ async fn test_stream_users_by_ids_valid_request() -> Result<()> {
...
+    assert!(
+        body.contains("100") || body.to_lowercase().contains("maximum"),
+        "Error message should mention the maximum items limit, got: {body}"
+    );
+
     Ok(())
 }
 
-#[tokio_shared_rt::test(shared)]
+// todo: this will no longer pass - id validation will fail
```
ok300: What's the best way to deal with this test?

Remove it? Can it instead be rewritten / updated?

nexus-webapi/tests/stream/user/list.rs
```diff
@@ -53,29 +53,38 @@ async fn test_stream_users_by_ids_valid_request() -> Result<()> {
...
+    let user_ids: Vec<String> = (0..101).map(|i| format!("user_id_{i}")).collect();
 
     let request_body = json!({
         "user_ids": user_ids,
         "viewer_id": null
     });
 
     // Send the POST request to the endpoint
-    invalid_post_request(
+    // Expect 422 Unprocessable Entity (JSON deserialization/validation error)
```
ok300: NIT: Right now the input validation error codes are inconsistent, based on whether they came from query/path or JSON body:
- Query/path invalid input: 400
- JSON body invalid input: currently 422

Axum treats extractor failures differently depending on where they happen.

For this PR, invalid query/path validation generally flows through `crate::Error::InvalidInput`, so it returns:
> 400 Bad Request

But invalid JSON body validation happens inside Axum’s `Json<T>` extractor before the handler runs. When `serde` fails while deserializing `UserIds`, `GlobalPostIds`, `PubkyId`, etc., Axum returns its default JSON rejection:
> 422 Unprocessable Entity

Question: should we keep the API behavior consistent (from a caller perspective) and use the same error code for both input validation error types?

---

If the API should consistently say “bad client input” as 400, the PR can wrap JSON extraction failures in the app error type. Example shape:

```rust
use axum::extract::{FromRequest, Json, Request};
use axum::response::IntoResponse;
pub struct ValidJson<T>(pub T);
impl<S, T> FromRequest<S> for ValidJson<T>
where
    S: Send + Sync,
    Json<T>: FromRequest<S>,
{
    type Rejection = crate::Error;
    async fn from_request(req: Request, state: &S) -> Result<Self, Self::Rejection> {
        Json::<T>::from_request(req, state)
            .await
            .map(|Json(value)| ValidJson(value))
            .map_err(|err| crate::Error::invalid_input(&err.into_response().status().to_string()))
    }
}
```

Then handlers that need consistent validation would use:
```rust
pub async fn stream_users_by_ids_handler(
    ValidJson(request): ValidJson<UserStreamByIdsRequest>,
) -> Result<Json<UserStream>> {
    ...
}
```

The actual error message mapping should preserve `JsonRejection::body_text()` rather than just status text, but the core idea is: convert Axum’s `JsonRejection` into `Error::InvalidInput`, so API validation failures return 400.

nexus-webapi/src/models/pubky_id.rs
```diff
@@ -0,0 +1,66 @@
+use std::fmt;
+use std::ops::Deref;
+
+use serde::{de, Deserialize};
+use utoipa::ToSchema;
+
+#[derive(ToSchema, Debug)]
```
ok300:
```suggestion
#[derive(ToSchema, Debug)]
#[schema(value_type = String)]
```

This helps utopipa render this in the Swagger UI as having string base type.

nexus-webapi/src/models/bounded_vec.rs
aintnostressin: It's redundant in my opinion.

If object is created it means it's valid so no need to explicitly state that it's validated.

nexus-webapi/src/models/bounded_vec.rs
aintnostressin: Yes, I wanted to check len before validating each item separately.

nexus-webapi/src/models/global_post_id.rs
aintnostressin: I see no benefit in adding another enum for representing errors that are immediately serialized.

nexus-webapi/src/models/path_extractor.rs
aintnostressin: I wanted to avoid single field structs. Thanks for feedback I'll consider `Path<X>`

nexus-webapi/src/models/post_id.rs
aintnostressin: Yes, hovewer I think time checks are unnecessary for reads https://github.com/pubky/pubky-app-specs/blob/main/src/traits.rs#L41-L57 - they make more sense for writes.

Additionaly - it may require `self` removal from signature

nexus-webapi/src/models/pubky_id.rs
aintnostressin: it also impl `Deserialize` so it's not only "path extracted". 

This type (and others) may be constructed in various way and coupling name to impl detail may result in more confusion in the future as the name will have to change more often.

I need to take a closer look at open-api generated specs to understand your concerns better.

nexus-webapi/src/models/resource_id.rs
aintnostressin: any actions that may require changes in pubky-app-specs i treat as  separate improvements that will be introduced in follow up PRs

nexus-webapi/src/routes/v0/post/bookmark.rs
aintnostressin: yes, draft = wip

nexus-webapi/tests/utils/mod.rs
aintnostressin: It's a helper fn to generate ids that can be used in tests. Please check usage.

docker/test-graph/mocks/posts.cypher
aintnostressin: This is only mock data, what kind of migration are you considering ?

nexus-webapi/src/routes/v0/stream/posts.rs
aintnostressin: I missed `StreamSource` completely

aintnostressin: done [b50449b](https://github.com/pubky/pubky-nexus/pull/849/commits/b50449bbb95ca91aa069233b2b2627090a49c6eb)

nexus-webapi/tests/stream/user/list.rs
aintnostressin: done [a8f9184](https://github.com/pubky/pubky-nexus/pull/849/commits/a8f91844312139d93ea66da7217864ae598f4c2d)

nexus-webapi/tests/stream/user/list.rs
aintnostressin: changed [890988c](https://github.com/pubky/pubky-nexus/pull/849/commits/890988c337cc08ca792a29553be94ac3008c5998)

nexus-webapi/src/models/bounded_vec.rs
aintnostressin: Good idea, done: [5bd2566](https://github.com/pubky/pubky-nexus/pull/849/commits/5bd25660ee9a8c9ad750009bab4268380f5c7db8)

nexus-webapi/tests/user/reach.rs
aintnostressin: done [eddcacf](https://github.com/pubky/pubky-nexus/pull/849/commits/eddcacfcfa97c1b1612d5969e75529601ee275d9)

nexus-webapi/src/models/pubky_id.rs
aintnostressin: We will switch to `pubky_app_specs::PubkyId` and remove this wrapper once https://github.com/pubky/pubky-app-specs/pull/110 is merged and applied to this repo.

nexus-webapi/src/models/pubky_id.rs
aintnostressin: This type is not being included in swagger docs - the `PubkyId` from specs takes precedence. Wrapper will be eventually removed.

nexus-webapi/src/models/mod.rs
aintnostressin: done [a2cd393](https://github.com/pubky/pubky-nexus/pull/849/commits/a2cd3936c26e0b789c55486efacf6f8b9eebc923)

docker/test-graph/mocks/posts.cypher
ok300: It seems like this field, which is reflected in the graph via the mock data import, (needs to?) change format from `V8N1P3L9J4X0` to `00000039YD99Y`.

Is this needed for non-mock-data too? If so, we need a migration.

nexus-webapi/src/routes/v0/stream/posts.rs
```diff
@@ -16,10 +16,80 @@ use serde::Deserialize;
...
+fn build_stream_source(
+    kind: &StreamSourceKind,
+    author_id: Option<&PubkyId>,
+    observer_id: Option<&PubkyId>,
+    post_id: Option<&PostId>,
+) -> StreamSource {
+    match kind {
+        StreamSourceKind::PostReplies => StreamSource::PostReplies {
+            post_id: post_id
+                .expect("post_id required for post_replies")
```
ok300: I wouldn't use `.expect()` as it would kill the service.

nexus-webapi/src/models/global_post_id.rs
ok300: It would allow for a more consistent handling of validation errors.

Now, some are returned as `String`, have the "Validation Error: " prefix and get converted to an error type in the caller

```rust
        temp_tag
            .validate(None)
            .map_err(|e| Error::invalid_input(&e.to_string()))?;
```

others are directly mapped to `invalid_input` without the prefix:

```rust
    pub(crate) fn validate(prefix: &str) -> Result<(), Error> {
        if prefix.chars().count() < USER_ID_SEARCH_MIN_PREFIX_LEN {
            return Err(Error::invalid_input(&format!(
                "ID prefix must be at least {USER_ID_SEARCH_MIN_PREFIX_LEN} chars"
            )));
        }
        Ok(())
    }
```

If a validation error type would be used, it could automatically be mapped to `Error::InvalidInput` via a `From` impl, for example. It wouldn't need manual mapping + it wouldn't even need the validation prefix.

A simpler solution is to just avoid the "validation error" prefix, as it's not applied consistently and it doesn't make the log statements more useful.

nexus-webapi/src/models/post_id.rs
ok300: The risk I see is that if the ID logic ever changes or is extended, even in a backward compatible way, it will be very hard to remember to update the logic here too (in addition to `pubky-app-specs`). I would generally avoid duplicating the ID generation logic.

Why not split the `TimestampId::validate_id` in `pubky-app-specs` in 2 parts: `validate_encoding` + `validate_encoded_timestamp`? The first would essentially be what you re-implemented here and it wouldn't need the `&self` arg.

nexus-webapi/src/models/pubky_id.rs
ok300: I think you can already do that in this PR.

First, use https://github.com/pubky/pubky-app-specs/pull/110 in `Cargo.toml` :

```diff
- pubky-app-specs = { git = "https://github.com/pubky/pubky-app-specs.git", rev = "bf78fe845ef55e96984a3bea954b69053fa7e8af", features = ["openapi"] }
+ pubky-app-specs = { git = "https://github.com/ok300/pubky-app-specs.git", branch = "codex/add-pubkyid-to-publickey-conversion", features = ["openapi"] }
```

Second, remove the wrapper and use `pubky_app_specs::PubkyId`.

nexus-webapi/tests/utils/mod.rs
ok300: `test_generate_post_id` is not really a test, it's more of a helper. But even so, it would belong more in `pubky-app-specs`.

It's a slippery slope if we start duplicating ID-generating logic.

I'd move `generate_post_id(ts)` (or equivalent) there, re-use it inside the normal ID generation, and remove `test_generate_post_id` here.

nexus-webapi/src/models/post_id.rs
aintnostressin: Can be done in follow up PR - wdyt ?

nexus-webapi/src/models/pubky_id.rs
aintnostressin: It's not merged yet so i'm not relying on it here. I'll switch once it's merged.

nexus-webapi/Cargo.toml
```diff
@@ -41,6 +41,7 @@ httpc-test = "0.1.10"
 pubky-testnet = { workspace = true }
 tempfile = { workspace = true }
 tokio-shared-rt = { workspace = true }
+tower = { version = "0.5", features = ["util"] }
```
ok300: Are you sure this is needed? Tests run fine without it.

aintnostressin: May be a leftover from other changes that were reverted in the meantime.

nexus-webapi/src/routes/v0/stream/posts.rs
aintnostressin: it was changed in https://github.com/pubky/pubky-nexus/pull/849/changes/6bf26b967a88ba403e48915d42f1eac0f9be2d4f

docker/test-graph/mocks/posts.cypher
aintnostressin: It just reveals the problem we had with mock data. Invalid identifiers were used.

nexus-webapi/src/routes/v0/stream/posts.rs
```diff
@@ -110,25 +189,26 @@ The `source` parameter determines the type of stream. Depending on the `source`,
 - *author*:  Requires  **author_id** to filter posts by a specific author.
 - *author_replies*:  Requires  **author_id** to filter replies by a specific author.
 
-Ensure that you provide the necessary parameters based on the selected `source`. If the required parameter is not provided, the provided `source` will be ignored and the stream type will default to *all*"#
+Ensure that you provide the necessary parameters based on the selected `source`. If a required parameter is missing, a 400 Bad Request error will be returned."#
 )]
 pub async fn stream_posts_handler(
     Query(mut query): Query<PostStreamQuery>,
```
ok300, line 226: Validated query params still use `axum::extract::Query` directly, so invalid query values like `viewer_id=bad`, `author_id=bad`, `post_id=short`, or invalid tags bypass `crate::Error` and return Axum’s default query rejection body instead of the JSON `{"error": ...}` shape used by the new `Path<T>`/`Json<T>` wrappers.

This conflicts with the consistency goal described in

https://github.com/pubky/pubky-nexus/blob/e7367391bafadeab26cf5e5d262ac9c7e1d37171/nexus-webapi/src/routes/mod.rs#L17-L20

Consider adding a `Query<T>` wrapper analogous to `Path<T>` and `Json<T>` and using it for the validated query structs.

nexus-webapi/src/models/user_id_prefix.rs
```diff
@@ -0,0 +1,157 @@
...
+    }
+}
+
+impl UserIdPrefix {
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    pub(crate) fn validate(prefix: &str) -> Result<(), Error> {
+        if prefix.chars().count() < USER_ID_SEARCH_MIN_PREFIX_LEN {
```
ok300, line 47: This count whitespace towards the minimum prefix length. The previous code used `prefix.trim().chars().count()`.

aintnostressin: that's true. I decided to accept value as it is, remove whitespaces before sending query.

nexus-webapi/src/models/post_id.rs
```diff
@@ -0,0 +1,136 @@
...
+use crate::models::bounded_vec;
+
+/// 13-character Crockford Base32-encoded post ID.
+#[derive(Debug, ToSchema)]
+#[schema(value_type = String, example = "00000039YD9DP")]
+pub struct PostId(pub String);
+
+/// Comma-separated list of `PostId` values (min=0, max=100).
+#[derive(Debug, ToSchema)]
+pub struct PostIds(pub Vec<PostId>);
```
ok300: This defines `PostIds`, but only `GlobalPostIds` is used in API args.

Looks like `PostIds` and its re-export can be removed.

nexus-webapi/tests/stream/post/post_replies.rs
ok300: These six Post IDs still have 12 chars (only remaining ones).

They're also reflected in `docker/test-graph/mocks/posts.cypher`.

nexus-webapi/src/models/error_response.rs
ok300, line 1: `error_response.rs` and its `ErrorResponse` aren't used in production code (only in tests).

Since `Error::into_response` still builds JSON inline at

https://github.com/pubky/pubky-nexus/blob/e7367391bafadeab26cf5e5d262ac9c7e1d37171/nexus-webapi/src/error.rs#L133-L137

the struct `ErrorResponse` is dead-weight and can be removed.

Tests that use it:

```rust
        assert!(
            error_response.error.contains("ID prefix must be at least"),
            "Error message should mention min size limit, got: {}",
            error_response.error
        );
```

can revert to the assertions made in `main`, so their diff in this PR can essentially be reverted:

```rust
        assert!(
            res["error"]
                .as_str()
                .unwrap_or("")
                .contains("ID prefix must be at least"),
            "Error message should mention min size limit"
        );
```

nexus-webapi/src/models/pubky_id.rs
ok300: Good point.

I merged it, so now you can reference it by bumping the `pubky-app-specs` dependency to its latest commit.

nexus-webapi/src/models/post_id.rs
ok300: I think it's a pretty small `pubky-app-specs` change, so it would be reviewed and merged quite quickly.

Worth doing it as part of this PR IMO.

nexus-webapi/src/models/pubky_id.rs
ok300: Now that #854 is merged, you can achieve this by simply merging `main` into this PR branch, to bring the latest `pubky-app-specs` + associated changes.

aintnostressin: done [5b9611e](https://github.com/pubky/pubky-nexus/pull/849/commits/5b9611ef0110e628f5e8102b0efc6f81e7a2e5db)

can be further simplified if https://github.com/pubky/pubky-app-specs/pull/114 is merged

nexus-webapi/tests/utils/mod.rs
aintnostressin: removed util [4ac3444](https://github.com/pubky/pubky-nexus/pull/849/commits/4ac34446bb5ad65a4156d77829d703f075d75ff4)

nexus-webapi/src/routes/v0/stream/posts.rs
aintnostressin: Thank you for pointing this out. fixed: [58ec29b](https://github.com/pubky/pubky-nexus/pull/849/commits/58ec29b3ac54e84c9d68ac441f3e420dbad995bc)

nexus-webapi/Cargo.toml
aintnostressin: removed: [636acf1](https://github.com/pubky/pubky-nexus/pull/849/commits/636acf1814e819e332428ca87a617e9054f9ba72)

nexus-webapi/src/models/post_id.rs
aintnostressin: done [4013b64](https://github.com/pubky/pubky-nexus/pull/849/commits/4013b64da392f6553e16e2349a91353d5bc996a6)

nexus-webapi/src/models/error_response.rs
aintnostressin: [39b1d41](https://github.com/pubky/pubky-nexus/pull/849/commits/39b1d41d2094f3adf7dc2e8f1016de9c62bc95ac)

renamed to `ErrorResponsePayload` - it's used in production code.
I prefer to keep it strongly typed. The code base and error handling should be consistent and rely on it.

nexus-webapi/src/models/post_id.rs
aintnostressin: waiting for upstream [PR](https://github.com/pubky/pubky-app-specs/pull/118)

nexus-webapi/tests/stream/post/post_replies.rs
aintnostressin: done [60b1f38](https://github.com/pubky/pubky-nexus/pull/849/commits/60b1f3895fa7c734ca758c8738ae9e3995c796c6)

nexus-webapi/src/models/global_post_id.rs
aintnostressin: We switched recently to `pubky_app_specs::PubkyId` and in some cases the error string will still contain "Validation error:" prefix. Fully eliminating it would need a change in `pubky-app-specs`, which I'd treat as a separate PR.

prefix removed: [1bf6c35](https://github.com/pubky/pubky-nexus/pull/849/commits/1bf6c35d6ea39137cea0c42f995caaae4abdd3dd)

aintnostressin: made consistent return types: [f115d4c](https://github.com/pubky/pubky-nexus/pull/849/commits/f115d4c2a12b73ca6c0e389443450512f77cbbc9)

nexus-webapi/src/models/pubky_id.rs
aintnostressin: simplified in [41d69f4](https://github.com/pubky/pubky-nexus/pull/849/commits/41d69f4acf86bc559265d7459eb6b642b9671ffd)

nexus-webapi/src/models/post_id.rs
aintnostressin: done: [a5b95c1](https://github.com/pubky/pubky-nexus/pull/849/commits/a5b95c15f706543af6126d603e8000d7f72c3246)

nexus-webapi/tests/utils/mod.rs
ok300: There is a related

https://github.com/pubky/pubky-nexus/blob/8bb8a73058cfc14096b5d0f61f232a04d76e87a2/nexus-watcher/tests/event_processor/utils/watcher.rs#L38

which can also be removed.

It's only used in

https://github.com/pubky/pubky-nexus/blob/8bb8a73058cfc14096b5d0f61f232a04d76e87a2/nexus-watcher/tests/event_processor/utils/watcher.rs#L262-L269

where its use can be replaced with

```diff
    pub async fn create_post(
        &mut self,
        user_kp: &Keypair,
        post: &PubkyAppPost,
    ) -> Result<(String, ResourcePath)> {
-       let post_id = generate_post_id();
+       let post_id = post.create_id();
```

Once `generate_post_id` is removed, `static COUNTER` can also go:

https://github.com/pubky/pubky-nexus/blob/8bb8a73058cfc14096b5d0f61f232a04d76e87a2/nexus-watcher/tests/event_processor/utils/watcher.rs#L33

ok300:
@greptile please review again

nexus-webapi/src/routes/v0/search/posts.rs
ok300: This still accepts the tag path parameter as raw String.

This endpoint now uses the custom `Path` extractor, but because the target type is `String`, malformed tag labels bypass the new `TagLabel` validation used by other tag endpoints. Inputs with commas, colons, internal whitespace, or labels over 20 chars reach `PostsByTagSearch::get_by_label` instead of returning `400`.

This leaves `/search/posts/by_tag/{tag}` inconsistent with `search/tags.rs`, post taggers, resource taggers, and Tags query validation.

nexus-webapi/src/models/user_id_prefix.rs
ok300: Hmm why? What benefit does this change bring, to justify the added fragility?

The previous logic had trimming in place, to catch this as early as possible.

> remove whitespaces before sending query

That acts then as last line of defense. If some future refactor accidentally removes it, the whole point of restricting the prefix to 3 chars (namely, to avoid overloading the DB with too generic user queries) is bypassed.

(Same applies to `UsernamePrefix::validate`, `UserIdPrefix::validate`)

nexus-webapi/src/error.rs
```diff
@@ -14,23 +15,27 @@ pub type Result<T> = core::result::Result<T, Error>;
 #[derive(Error, Debug)]
 pub enum Error {
```
ok300, line 16: The use of `Box<PubkyId>` in this enum's values is unnecessary.

This is an error struct, where the inner fields are essentially used for logging, so they can be plain `String`.

I'd therefore change
- the use of `Box<PubkyId>` -> `String`
- the use of `PostId` -> `String`

nexus-webapi/src/routes/static/files.rs
```diff
@@ -19,17 +22,15 @@ use nexus_common::{
...
     // dl is download parameter to set the content-disposition header to attachment
     dl: Option<String>,
 }
 
-#[derive(Deserialize, Serialize)]
+#[derive(Deserialize)]
 pub struct FilePath {
-    owner_id: String,
+    owner_id: PubkyId,
     file_id: String,
```
ok300: NIT: `file_id` remains an unvalidated `String` coming via

https://github.com/pubky/pubky-nexus/blob/8bb8a73058cfc14096b5d0f61f232a04d76e87a2/nexus-webapi/src/routes/static/files.rs#L50

I'm not sure if there are any rules we can enforce about the file ID.

nexus-webapi/src/models/tag_label.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+impl TagLabel {
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    fn validate_and_sanitize(label: String) -> Result<String, Error> {
+        let temp_tag = PubkyAppTag::new(
```
ok300, line 45: Why not directly use `sanitize_tag_label` / `validate_tag_label` from `pubky-app-specs`?

Looks like these two functions are already public in `pubky-app-specs`. If they have to be exported to be usable here, I think a small PR on `..specs` is good idea.

nexus-webapi/Cargo.toml
```diff
@@ -10,6 +10,7 @@ build = "build.rs"
 
 [dependencies]
 axum = "0.8.9"
+base32 = { workspace = true }
```
ok300: I wonder if we really need this.

It's only used in 1-2 test utils, which I think could re-use `pubky-app-specs` utils, to avoid adding this dependency.

nexus-webapi/src/models/bounded_vec.rs
```diff
@@ -0,0 +1,155 @@
...
+        return Err(de::Error::custom(format!(
+            "At least {MIN} item(s) required"
+        )));
+    }
+    if arr.len() > MAX {
+        return Err(de::Error::custom(format!("Maximum {MAX} items allowed")));
+    }
+    arr.into_iter()
+        .map(|v| serde_json::from_value(v).map_err(de::Error::custom))
+        .collect()
```
ok300, line 57: `deserialize_json_array` goes through `serde_json::Value`, checks length, then re-deserializes each item. The logic can be shorter and less JSON-coupled:

```suggestion
    let items = Vec::<T>::deserialize(deserializer)?;
    if items.len() < MIN {
        return Err(de::Error::custom(format!(
            "At least {MIN} item(s) required"
        )));
    }
    if items.len() > MAX {
        return Err(de::Error::custom(format!("Maximum {MAX} items allowed")));
    }
    Ok(items)
```

nexus-webapi/src/models/bounded_vec.rs
```diff
@@ -0,0 +1,155 @@
...
+        .map(|t| T::try_from(t.to_string()).map_err(de::Error::custom))
+        .collect::<Result<_, _>>()?;
+    if items.len() < MIN {
+        return Err(de::Error::custom(format!(
+            "At least {MIN} item(s) required"
+        )));
+    }
+    if items.len() > MAX {
+        return Err(de::Error::custom(format!("Maximum {MAX} items allowed")));
+    }
```
ok300: The length validation is duplicated in the CSV and JSON deserializers, so it can be extracted to smth like

```rust
fn validate_len<E, const MIN: usize, const MAX: usize>(len: usize) -> Result<(), E>
where
    E: serde::de::Error,
{
    if len < MIN {
        return Err(E::custom(format!("At least {MIN} item(s) required")));
    }
    if len > MAX {
        return Err(E::custom(format!("Maximum {MAX} items allowed")));
    }
    Ok(())
}
```

then it could be used in both deserializers as

```rust
validate_len::<D::Error, MIN, MAX>(items.len())?;
```

nexus-webapi/src/routes/mod.rs
```diff
@@ -9,6 +14,68 @@ pub mod v0;
 
 mod middlewares;
 
+/// A wrapper around Axum's Json extractor that maps deserialization errors
+/// to Error::InvalidInput (400 Bad Request) for consistent error handling
+/// across all extraction types (Query, Path, and Json).
```
ok300:
```suggestion
/// JSON extractor that maps Axum rejections to `Error::InvalidInput`.
```

Rustdoc can be made more concise.

nexus-webapi/src/routes/mod.rs
```diff
@@ -9,6 +14,68 @@ pub mod v0;
...
+            .map_err(|rejection| crate::Error::invalid_input(&rejection.to_string()))?;
+        Ok(Json(json.0))
+    }
+}
+
+/// A wrapper around Axum's Path extractor that maps deserialization/validation errors
+/// to Error::InvalidInput (400 Bad Request) for consistent JSON error responses.
+/// This ensures that path parameter validation failures return a proper JSON body
+/// instead of Axum's default plain-text rejection.
+pub struct Path<T>(pub T);
```
ok300:
```suggestion
/// Path extractor that maps Axum rejections to `Error::InvalidInput`.
pub struct Path<T>(pub T);
```


Rustdoc can be made more concise.

nexus-webapi/src/routes/mod.rs
```diff
@@ -9,6 +14,68 @@ pub mod v0;
...
+            .map_err(|rejection| crate::Error::invalid_input(&rejection.to_string()))?;
+        Ok(Path(path.0))
+    }
+}
+
+/// A wrapper around Axum's Query extractor that maps deserialization/validation errors
+/// to Error::InvalidInput (400 Bad Request) for consistent JSON error responses.
+/// This ensures that query parameter validation failures return a proper JSON body
+/// instead of Axum's default plain-text rejection.
+pub struct Query<T>(pub T);
```
ok300:
```suggestion
/// Query extractor that maps Axum rejections to `Error::InvalidInput`.
pub struct Query<T>(pub T);
```

Rustdoc can be made more concise.

nexus-webapi/src/error.rs
aintnostressin: done [73d918b](https://github.com/pubky/pubky-nexus/pull/849/commits/73d918b4b7324118dfd6d8e251e7858c845a7566)

nexus-webapi/src/routes/v0/search/posts.rs
ok300: Should probably be changed to `TagLabel`, together with L25 above.

nexus-webapi/Cargo.toml
aintnostressin: removed [3b3a123](https://github.com/pubky/pubky-nexus/pull/849/commits/3b3a12364591a67feac71b819d79617f7d4d7158)

nexus-webapi/tests/utils/mod.rs
aintnostressin: Please remove this in separate PR if needed.

nexus-webapi/src/models/error_response.rs
```diff
@@ -0,0 +1,22 @@
...
+
+impl ErrorResponsePayload {
+    pub fn new(error: impl Into<String>) -> Self {
+        Self {
+            error: error.into(),
+        }
+    }
+}
+
+impl IntoResponse for ErrorResponsePayload {
```
ok300: `ErrorResponsePayload::IntoResponse` is redundant.

`ErrorResponsePayload` is a useful struct, but its `IntoResponse` impl does exactly one thing:

```rust
impl IntoResponse for ErrorResponsePayload {
    fn into_response(self) -> axum::response::Response {
        axum::Json(self).into_response()  // just wraps in Json
    }
}
```

The entire purpose is to allow `(status_code, body).into_response()` in `error.rs` without writing `axum::Json(body)`. That saves one word at the callsite but introduces implicit behavior.

I would remove it. In the caller, writing  `(status_code, axum::Json(body)).into_response()` is clearer:

```diff
  let body = ErrorResponsePayload::new(self.to_string());

- (status_code, body).into_response()
+ (status_code, axum::Json(body)).into_response()
```

nexus-webapi/src/models/bounded_vec.rs
aintnostressin: The code snippet provided breaks intention described in https://github.com/pubky/pubky-nexus/pull/849#discussion_r3180256055

nexus-webapi/src/models/mod.rs
```diff
@@ -1,5 +1,62 @@
...
+pub use pubky_app_specs::PubkyId;
+pub use resource_id::ResourceId;
+pub use tag_label::TagLabel;
+pub use user_id_prefix::UserIdPrefix;
+pub use username_prefix::UsernamePrefix;
+
+/// Comma-separated list of tag labels (min=1, max=5). Each label is validated and sanitized.
+#[derive(Debug, ToSchema)]
+#[schema(value_type = String, example = "dev,free,opensource")]
+pub struct Tags(pub Vec<TagLabel>);
```
ok300, line 34: `UserIds` and `Tags` are inconsistent and leak raw `.0` access at callsites.

`GlobalPostIds` has `into_string_vec()`, but `UserIds` and `Tags` don't, so every callsite is forced to reach inside the wrapper:

```rust
  // stream/users.rs — UserIds callsite:
  let user_ids: Vec<String> = request.user_ids.iter().map(|id| id.to_string()).collect();

  // stream/resources.rs — Tags (duplicated twice, once per handler):
  let tags: Option<Vec<String>> = query.tags.as_ref()
      .map(|t| t.0.iter().map(|tl| tl.0.clone()).collect());
```

The `t.0.iter()` also ignores the `Deref<Target=Vec<TagLabel>>` that `Tags` provides — it should at least be `t.iter()`. And `PostStreamQuery::tags_as_strings()` does the same conversion but as a method on the query struct rather than on `Tags` itself.

If `Tags` had `into_string_vec()` (mirroring `GlobalPostIds`), all three sites would collapse to a one-liner and the asymmetry would disappear.

nexus-webapi/src/models/bounded_vec.rs
aintnostressin: done: [3606d7a](https://github.com/pubky/pubky-nexus/pull/849/commits/3606d7a157ee9ff8c9235e687bdef6549b629c12)


However i think JSON deserializer should be dropped, but I don't know why such inconsistency was introduced.

nexus-webapi/src/models/error_response.rs
aintnostressin: done [b6e0b5c](https://github.com/pubky/pubky-nexus/pull/849/commits/b6e0b5cfe2b87942b87e2867151f43b15c76a07b)

nexus-webapi/src/models/mod.rs
aintnostressin: done [13ffdac](https://github.com/pubky/pubky-nexus/pull/849/commits/13ffdac9b4a06b7d48a684e52b9241b8dd899cb8)

nexus-webapi/src/routes/v0/search/posts.rs
aintnostressin: done [e7abac0](https://github.com/pubky/pubky-nexus/pull/849/commits/e7abac0e92cb3f0d6d2cabfe7682ecabbba1333b)

nexus-webapi/src/models/tag_label.rs
aintnostressin: This is just a move of existing code from https://github.com/pubky/pubky-nexus/blob/main/nexus-webapi/src/routes/v0/search/tags.rs#L56-L69

nexus-webapi/src/models/user_id_prefix.rs
aintnostressin: Ok I see, you are right.

I though it counts toward MAX not MIN.


done: [7515f66](https://github.com/pubky/pubky-nexus/pull/849/commits/7515f665c8b368378f5b4a13a7f08ca1da67d3a8)

nexus-webapi/src/models/username_prefix.rs
```diff
@@ -0,0 +1,127 @@
...
+impl UsernamePrefix {
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+
+    fn validate(prefix: &str) -> Result<(), Error> {
+        if prefix.is_empty() {
+            return Err(Error::invalid_input("Username cannot be empty"));
+        }
+        Ok(())
```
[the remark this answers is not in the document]
aintnostressin: this should be fixed

nexus-webapi/src/routes/static/files.rs
aintnostressin: done [e64914a](https://github.com/pubky/pubky-nexus/pull/849/commits/e64914ab6806fd4107dba56f9a61a2d82589b37b)

Added `FileId` it's almost identical to `PostId`. Both are basically `TimestampId` . I decided to keep two in order to have them render in swagger with more friendly names. 

We can reduce to one `TimestampId` type or have a macro to generate duplicated code.

nexus-webapi/src/models/tag_label.rs
aintnostressin: @ok300 let's keep it for a follow up PR.

nexus-webapi/src/routes/v0/tag/global.rs
```diff
@@ -87,7 +89,7 @@ pub async fn tag_taggers_handler(
     description = "Global Tags by reach",
     tag = "Tags",
     params(
-        ("user_id" = Option<String>, Query, description = "User Pubky ID"),
+        ("user_id" = Option<PubkyId>, Query, description = "User Pubky ID"),
         ("reach" = Option<StreamReach>, Query, description = "Reach type: `follower` | `following` | `friends` | `wot`. To apply that, user_id is required"),
```
ok300, line 93: One issue I noticed on this endpoint (`/v0/tags/hot`) when I tried to set the `user_id` and `reach`, is the query failed with

> Failed to deserialize query string: reach: unknown variant Followers, expected one of followers, following, friends, wot, wot_1, wot_2, wot_3

The GET request was triggered from the Swagger UI available at `/swagger-ui`  then choosing `Followers` from the reach dropdown.

Looks like it's related to the validation of `StreamReach` and / or what Swagger UI considers as valid values that it proposes in the dropdown (in this case `Followers` capitalized, when the error msg mentioned `followers` lowercase).

In fact, I think `StreamReach` is maybe just missing some annotations, like:

```diff
+ #[serde(rename_all = "snake_case")]
+ #[schema(rename_all = "snake_case")]
  pub enum StreamReach {
```

I initially wanted to propose a fix, but since it's related to input validation, I thought it's better to abandon my PR and instead raise the point here. You can see my initial draft PR here: https://github.com/ok300/pubky-nexus/pull/124 (it also covers some `wot` field renaming).

What do you think? Would those annotations be enough?

aintnostressin: Honestly - I don't think this is related to validation and this issue is preexisting on main branch.
To me it seems like schema - serde misalignment.

Happy to review https://github.com/ok300/pubky-nexus/pull/124 once ready.

ok300:
@greptile please review again

nexus-webapi/src/models/tag_label.rs
ok300: Tracking in #886

ok300 (approved):

PR #851, opened 2026-05-06
[`univ-tag-del` B] fix: tag deletion for tags pointing to known Pubky resources

ok300:
This PR addresses a bug in tag deletion, whereby Universal Tags that point to known Pubky resources (User / Post) would be skipped during deletion or moderation.

This is because when such tags are indexed, `sync_put_resource` delegates `InternalKnown` targets to `sync_put` (`nexus-watcher/src/events/handlers/tag.rs:76`), which creates post/user `TAGGED` edges without `tag.app` (`nexus-common/src/db/graph/queries/put.rs:239`). The new delete path parses `/pub/<app>/tags/<id>` and filters with `tag.app = $app` (`nexus-watcher/src/events/handlers/tag.rs:412`, `nexus-common/src/db/graph/queries/get.rs:132`), so the matching edge is not found and delete/moderation becomes an idempotent no-op. This affects normal DELs for app-specific internal-known tags and moderation tags pointing at those tag URIs.

---

Detailed list of changes made:
- `nexus-watcher/src/events/handlers/tag.rs`: app-specific tag deletes now fall back to the app-less Post/User tag edge used by the existing internal-known PUT flow.
- `nexus-watcher/tests/event_processor/tags/resource_internal_known.rs`: asserts app-specific internal-known DEL removes the Post tag.
- `nexus-watcher/tests/event_processor/tags/moderated.rs`: adds moderation coverage for app-specific known Post tags.

### DIFF
--- a/nexus-watcher/src/events/handlers/tag.rs
+++ b/nexus-watcher/src/events/handlers/tag.rs
@@ -402,26 +402,32 @@ async fn put_sync_user(
 #[tracing::instrument(name = "tag.del", skip_all, fields(tag_uri = %tag_uri))]
 pub async fn del(tag_uri: &str) -> Result<(), EventProcessorError> {
     let tag_storage_uri = parse_tag_storage_uri(tag_uri)?;
-    let user_id = tag_storage_uri.user_id;
-    let tag_id = tag_storage_uri.tag_id;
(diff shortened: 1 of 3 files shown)


SHAcollision (approved):
LGTM

PR #857, opened 2026-05-09
feat: index Collections via kind=collection

SHAcollision:
> [DRAFT] Stacked on the unmerged `pubky-app-specs` v0.5.0 branch — see "Spec dep wiring" below.

## Summary

Adds support for the new `kind=collection` post type (introduced in pubky-app-specs v0.5.0) to the Nexus indexer + web API. Collections flow through the existing post pipeline (graph node + cached PostDetails) so they remain queryable, but they are **suppressed from default Home / Hot / By-Tag streams** and surface only when callers explicitly pass `?kind=collection`.

**Zero new endpoints, zero new graph constraints, zero new homeserver paths.** All collection surfaces ride the existing `/v0/stream/posts` route.

## Spec dep wiring (temporary, will be flipped before merge)

`Cargo.toml` currently pins `pubky-app-specs` to the unmerged v0.5.0 branch:

```toml
pubky-app-specs = { git = "https://github.com/pubky/pubky-app-specs.git", branch = "feat/spec-v0.5.0-collection-variant", features = ["openapi"] }
```

Once https://github.com/pubky/pubky-app-specs/pull/116 (and its parent #115) merge and v0.5.0 is tagged, a follow-up commit on this branch flips the dep to a `rev = <v0.5.0-tag-hash>` pin. Until then this PR's CI runs against the spec branch.

## Indexing changes (one PR, two commits)

**Commit 1 — integration code:**

- Three Redis-writer gates so collections never enter the index sorted sets that power default streams:
  - `add_to_timeline_sorted_set` (gated in `nexus-common/.../post/details.rs`)
  - `add_to_per_user_sorted_set` (same)
  - `PostCounts::put_to_index` engagement set (gated inline at the watcher call site, single call site, no signature change)
- Cypher fallback at `get.rs:836-838`: when no kind filter is set, append `p.kind <> 'collection'` so the Cypher path doesn't re-leak collections.
- Tag-handler gate: look up the target post's kind via `PostDetails::get_by_id` and skip `PostsByTagSearch::put_to_index` for collection-kind targets at both call sites.
- New `EventProcessorError::SpecValidation(String)` non-retryable variant. `PubkyAppObject::from_resource` validation failures are deterministic — re-running produces the same error, so we drop them rather than poisoning the retry queue. The classifier in `extract_retry_event_info` matches it next to `InvalidEventLine`.
- OpenAPI doc-comment for the `kind` query parameter now lists `collection` and notes the default-exclude behavior.

**Commit 2 — test suite:**

- `collection_post(name, description, attachments)` test builder.
- 1 nexus-common unit test (`can_use_index` short-circuit for kind=Collection).
- 3 watcher integration tests (create / edit / delete a collection; assert graph is populated AND sorted sets stay empty).
- 1 forwards-compat regression test that asserts `Unknown`-kind events classify as non-retryable `SpecValidation` errors — the load-bearing assertion that justifies the staged v0.4.5 → v0.5.0 spec rollout.
- 3 webapi integration tests (default stream excludes collections / `?kind=collection` returns only collections / profile-collections-tab returns the author's collections).
- Seed data: 3 collection posts (`COLW1TGL5BKG1-3`) added to `docker/test-graph/mocks/posts.cypher`.

## Test plan

- [ ] `docker compose up -d` then `cargo nextest run --workspace --no-fail-fast` green
- [ ] CI green
- [ ] Spec dep flipped to `rev = <v0.5.0-tag-hash>` once https://github.com/pubky/pubky-app-specs/pull/116 merges
- [ ] Move out of draft after the spec dep is flipped


nexus-watcher/src/service/processor.rs
```diff
@@ -244,3 +250,59 @@ fn extract_retry_event_info(
...
+    use super::*;
+    use nexus_common::models::event::ParseResult;
+    use std::path::PathBuf;
+
+    /// Build a syntactically-valid `Event` for classifier tests. The body of
+    /// `extract_retry_event_info` only consults `event.uri` for the retry-eligible
+    /// branch; the skip branches we test below short-circuit before reaching it.
+    fn fixture_event() -> Event {
+        let line = "PUT pubky://4snwyct86m383rsduhw5xgcxpw7c63j3pq8x4ycqikxgik8y64ro/pub/pubky.app/posts/0034A0X7NJ52A";
+        match Event::parse_event(line, PathBuf::from("/tmp/nexus-test-files")).unwrap() {
```
aintnostressin: use `tempfile` crate for producing cross os valid temp path

nexus-watcher/tests/event_processor/posts/forwards_compat.rs
```diff
@@ -0,0 +1,75 @@
...
+        "kind": "collection",
+        "parent": null,
+        "embed": null,
+        "attachments": [
+            "pubky://userA/pub/pubky.app/posts/0034A0X7NJ52A"
+        ]
+    }"#;
+
+    // Deserialization succeeds because of `#[serde(other)] Unknown`.
+    let post: PubkyAppPost =
```
aintnostressin, line 39: this succeeds because `Collection` variant is present in newer specs. I think blob should use a not used yet post kind so we can check it falls back to Unknown variant here.

nexus-watcher/tests/event_processor/posts/validation_errors.rs
```diff
@@ -0,0 +1,100 @@
...
+        .validate(Some(&id))
+        .expect_err("Collection post with parent must fail validation");
+    let lower = err.to_lowercase();
+    assert!(
+        lower.contains("parent") || lower.contains("embed"),
+        "validation error must mention 'parent' (or the combined 'parent or embed'), got: {err}"
+    );
+
+    // The watcher wraps spec errors as SpecValidation (non-retryable).
+    let wrapped = EventProcessorError::SpecValidation(err.clone());
```
aintnostressin: are we testing specs or nexus-watcher logic here ? 

It seems like this test should not belong to nexus at all.

nexus-webapi/tests/stream/post/kind/collection.rs
```diff
@@ -0,0 +1,165 @@
...
+        .as_array()
+        .expect("Post stream should be an array")
+        .iter()
+        .map(|p| p["details"]["id"].as_str().unwrap_or_default().to_string())
+        .collect()
+}
+
+#[tokio_shared_rt::test(shared)]
+async fn test_kind_collection_returns_only_collections() -> Result<()> {
+    let path = format!("{ROOT_PATH}?source=all&kind=collection&sorting=timeline");
```
aintnostressin, line 57: Would be good to have here result for non collection kind as a pre check so we can see that `kind=collection` is effective.

nexus-webapi/tests/stream/post/kind/collection.rs
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    Ok(())
+}
+
+#[tokio_shared_rt::test(shared)]
+async fn test_default_stream_excludes_collections() -> Result<()> {
+    // No `kind` filter → default streams must exclude collections via the
+    // Cypher `p.kind <> 'collection'` clause introduced in Phase 3.
+    let path = format!("{ROOT_PATH}?source=all&sorting=timeline&limit=200");
```
aintnostressin, line 103: Same as above - would be good to see the results with `kind=collection` so test can really assert that results are different and param is effective. Then test could simply check missing posts ids and assert they are of collection type.

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -832,9 +833,19 @@ pub fn post_stream(
...
 
-    // If post kind is provided, add the corresponding condition
-    if kind.is_some() {
-        append_condition(&mut cypher, "p.kind = $kind", &mut where_clause_applied);
+    // If post kind is provided, add the corresponding condition.
+    // When no kind filter is set, exclude collections from the default stream
+    // (defense-in-depth: the Redis writers in models/post/details.rs and the
+    // watcher's PostCounts call site already gate collections out of the index
+    // sorted sets, but the Cypher fallback path bypasses those, so we filter
+    // here too).
```
ok300:
```suggestion
```

I would remove these details, as it will be extremely hard to keep them accurate as the code evolves (refactors, renames, etc).

nexus-watcher/src/events/handlers/post.rs
```diff
@@ -28,6 +28,10 @@ pub async fn sync_put(
     let post_details = PostDetails::from_homeserver(post.clone(), &author_id, &post_id);
     // We avoid indexing replies into global feed sorted sets
     let is_reply = post.parent.is_some();
+    // Collections are excluded from the global engagement sorted set so they
+    // do not pollute Home / Hot streams; they surface only via explicit
+    // `?kind=collection` queries (Cypher fallback).
```
ok300:
```suggestion
```

I would remove these, as this is a generic detail about the feature, not specific to why this variable is needed here. The right place to document why the var is needed is where it's used in a function (already done, lower below around `PostCounts::put_to_index`)

nexus-watcher/src/events/handlers/tag.rs
```diff
@@ -200,6 +202,24 @@ async fn put_sync_resource(
...
 /// - `indexed_at` - A 64-bit integer representing the timestamp when the post was indexed.
 ///
+/// Returns `true` if the post at `(author_id, post_id)` is a Collection.
+///
+/// Only called from the `Updated` and `CreatedOrDeleted` arms of
+/// `put_sync_post` — never from `MissingDependency`, which returns early
+/// before any index writes — so the lookup cost never lands on the retry
+/// path. Returns `false` if the post isn't indexed yet (e.g. a tag-on-
+/// non-existent-post race), which is the safe default: indexing the tag
+/// in the by-tag stream is the standard behavior for unknown targets.
```
ok300:
```suggestion
/// Returns `true` if the post at `(author_id, post_id)` is indexed and is a Collection.
```

I'd remove the lower section of the rustdoc, as it doesn't help understand why the method exists or how it works. Instead its descriptive of the way this is used now. May or may not be true as the code evolves + will be very hard to keep accurate as time goes on.

nexus-watcher/tests/event_processor/posts/utils.rs
```diff
@@ -182,6 +183,32 @@ pub fn short_repost(content: impl Into<String>, parent_uri: String) -> PubkyAppP
...
+        name: name.into(),
+        description: description.map(|s| s.to_string()),
+    };
+    let content = serde_json::to_string(&envelope).expect("envelope serialization");
+    PubkyAppPost {
+        content,
+        kind: PubkyAppPostKind::Collection,
+        parent: None,
+        embed: None,
+        attachments: if attachments.is_empty() {
```
ok300: Can a collection have attachments?

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -832,9 +833,19 @@ pub fn post_stream(
...
+    match &kind {
+        Some(_) => append_condition(&mut cypher, "p.kind = $kind", &mut where_clause_applied),
+        None => append_condition(
+            &mut cypher,
+            "p.kind <> 'collection'",
+            &mut where_clause_applied,
+        ),
```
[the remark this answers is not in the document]
ok300: Greptile made a good point, the addition to the API field description says

> Note: `source=bookmarks` is exempt — a bookmarked collection appears in the bookmarks tab regardless of the kind filter (a user explicitly bookmarking a collection should see it among their bookmarks).

nexus-webapi/tests/stream/post/kind/collection.rs
```diff
@@ -0,0 +1,165 @@
+//! Tests for `kind=collection` filtering on `/v0/stream/posts`.
+//!
+//! Seed data (in `docker/test-graph/mocks/posts.cypher`):
+//!   - 3 collection posts: COLW1TGL5BKG1, COLW1TGL5BKG2 (author: $bogota),
+//!     COLW1TGL5BKG3 (author: $cairo).
+//!
+//! Phase 3 stream-suppression invariants we assert here:
```
ok300: "Phase 3" is mentioned several times in the rustdocs, but not sure what it means.

If its a Collections internal-doc reference, I'd remove it.

If it's relevant for the codebase, I'd keep it and add an `md` doc file where people can cross-reference to see what Phase 3 is.

nexus-watcher/src/events/handlers/post.rs
ok300: Collections can still enter the global Hot/engagement stream after receiving a reply, repost, or tag.

The following paths only check `!post_relationship_is_reply`:
- `handlers/post.rs:170` // here, for replies
- `handlers/post.rs:220` // reposts
- `handlers/tag.rs:317` // tags

Looks like those places need a collection check for the target / parent post.

nexus-common/src/db/graph/queries/get.rs
ok300, line 213: `PostsByTagSearch::reindex()` rebuilds both by-tag sorted sets from graph queries that include collection posts:

https://github.com/pubky/pubky-nexus/blob/ef0878dfecf08f37168093b7eded91767cd41a19/nexus-common/src/models/post/search.rs#L36-L40

After a Redis reindex, tagged collections will appear in `?tags=..` timeline and total-engagement streams even if the watcher write path skips them.

To fix this, the `global_tags_by_post*` queries:
- `global_tags_by_post` (here)
- `global_tags_by_post_engagement`

need a `post.kind <> 'collection'` exclusion.

nexus-watcher/src/events/handlers/tag.rs
```diff
@@ -291,8 +325,14 @@ async fn put_sync_post(
...
                     Ok::<(), EventProcessorError>(())
                 },
-                // Add post to global label timeline
-                PostsByTagSearch::put_to_index(&author_id, post_id, tag_label),
+                // Add post to global label timeline (skip for collections so they
+                // don't pollute by-tag streams; reachable only via ?kind=collection)
+                async {
+                    if !target_is_collection {
+                        PostsByTagSearch::put_to_index(&author_id, post_id, tag_label).await?;
+                    }
```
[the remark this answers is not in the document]
SHAcollision: Adressed de85b7f5

nexus-common/src/db/graph/queries/get.rs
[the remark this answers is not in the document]
SHAcollision: Adressed de85b7f5

nexus-common/src/models/post/counts.rs
```diff
@@ -27,9 +27,9 @@ impl PostCounts {
             Some(counts) => Ok(Some(counts)),
             None => {
                 let graph_response = Self::get_from_graph(author_id, post_id).await?;
-                if let Some((post_counts, is_reply)) = graph_response {
+                if let Some((post_counts, is_reply, is_collection)) = graph_response {
                     post_counts
-                        .put_to_index(author_id, post_id, !is_reply)
+                        .put_to_index(author_id, post_id, !is_reply, is_collection)
                         .await?;
```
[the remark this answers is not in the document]
SHAcollision: Adressed de85b7f5

nexus-watcher/src/events/handlers/tag.rs
```diff
@@ -200,6 +202,24 @@ async fn put_sync_resource(
...
+/// in the by-tag stream is the standard behavior for unknown targets.
+async fn target_post_is_collection(
+    author_id: &PubkyId,
+    post_id: &str,
+) -> Result<bool, EventProcessorError> {
+    Ok(PostDetails::get_by_id(author_id, post_id)
+        .await?
+        .map(|d| matches!(d.kind, PubkyAppPostKind::Collection))
+        .unwrap_or(false))
+}
```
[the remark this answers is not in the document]
SHAcollision: Adressed de85b7f5

nexus-watcher/tests/event_processor/posts/forwards_compat.rs
SHAcollision: fixed in de85b7f5

nexus-watcher/tests/event_processor/posts/validation_errors.rs
SHAcollision: you are right; deleted the file entirely in de85b7f5

nexus-webapi/tests/stream/post/kind/collection.rs
SHAcollision: added baseline assertions in de85b7f5; test now proves the kind filter changes the result set.

nexus-webapi/tests/stream/post/kind/collection.rs
SHAcollision: same fix; baseline added in de85b7f5

nexus-watcher/tests/event_processor/posts/utils.rs
ok300: Just now saw the `pubky-app-specs` PR, indeed attachments are supported so I'll mark this as resolved.

nexus-watcher/src/service/processor.rs
SHAcollision: I just took a deeper look, it seems this path literal goes nowhere importat: the path is accessed `parse_event` stores `files_path` on the resulting Event, but the three tests using `fixture_event()` all call  `extract_retry_event_info(&event, err)`, which only inspects `event.uri` and the passed error, never `event.files_path`. So the literal is dead state in this test path. Still, the portability nit stands for Windows. So will fix in a follow-up commit: need to add `tempfile = { workspace = true }` to `nexus-watcher` dev-deps too and switch the fixture to `tempfile::tempdir()`.

nexus-webapi/tests/stream/post/kind/collection.rs
SHAcollision: Removed in bd8fe938

nexus-watcher/src/events/handlers/post.rs
SHAcollision: Fixed in 31120cc2

Gated all 4 affected sites in `post.rs` . The DEL mirror matters: `decrement_score_index_sorted_set` is `ZINCRBY` -1, which creates the member on absence, so fixing only the INC side would have leaked the Collection into `POST_TOTAL_ENGAGEMENT` with score -1 on reply/repost deletion. Two new regression tests (`reply_to_collection.rs`, `repost_collection.rs`) assert both PUT and DEL phases stay clean.

Re: `tag.rs:317`, that path is already gated at :325 (the target_is_collection check wraps the engagement increment in that `traced_join!` arm), but thanks for the systematic enumeration.

nexus-common/src/db/graph/queries/get.rs
SHAcollision: Good catch. Fixed in e44da263. Added `WHERE post.kind <> 'collection'` to both `global_tags_by_post` and `global_tags_by_post_engagement`. New regression test (nexus-watcher/tests/event_processor/posts/reindex_collection.rs) tags a Collection, runs `PostsByTagSearch::reindex()`, and asserts both sorted sets stay free of the Collection.

nexus-watcher/src/events/handlers/tag.rs
SHAcollision: Applied in 31120cc2. The helper was moved to `handlers/utils.rs` alongside `post_relationships_is_reply` . Kept just the one-liner you suggested.

aintnostressin (approved):

ok300:
@greptile please review again

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -832,9 +835,24 @@ pub fn post_stream(
...
+    // If post kind is provided, add the corresponding condition.
+    // When no kind filter is set, exclude collections from the default stream
+    //
+    // Exception: `source=bookmarks` is exempt — a user who bookmarked a
+    // collection should see it in their bookmarks stream regardless of kind.
+    // The Redis bookmark sorted set is already kind-agnostic, so under
+    // `sorting=timeline` bookmarked collections surface naturally. Under
+    // `sorting=total_engagement` `can_use_index` returns false and execution
+    // falls through to this Cypher path; without this exemption we would
+    // silently drop bookmarked collections when sorting changes.
```
ok300:
```suggestion
```

I'd remove these.

They are too verbose and describe behavior somewhere else (iif / when logic changes there, this will be a stale / confusing / misleading 5 lines of docs).

The lines just above are clear and concise, and refer to the logic in this (graph) method.

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -212,6 +213,7 @@ pub fn global_tags_by_post() -> Query {
         "global_tags_by_post",
         "
         MATCH (tagger:User)-[t:TAGGED]->(post:Post)<-[:AUTHORED]-(author:User)
+        WHERE post.kind <> 'collection'
```
ok300, line 216: > WHERE post.kind <> 'collection'

This filter is used in queries at `nexus-common/src/db/graph/queries/get.rs:216`, `:234`, `:853`.

Unfortunately it excludes legacy posts where `kind` is missing.

In Cypher, `NULL <> 'collection'` evaluates to `null/false` in `WHERE`, while this codebase already treats missing kind as `short` via `COALESCE(p.kind, 'short')` in `get_post_by_id`. This can drop older posts from Cypher fallback streams and by-tag reindex results after Redis rebuilds.

A possible fix is to use `COALESCE(post.kind, 'short') <> 'collection'`.

SHAcollision: > Unfortunately it excludes legacy posts where kind is missing. 

I don't exactly remember if there ever was a period of posts without kind so I checked. The concern seems to be theoretical, there never were posts without `kind`. I could still apply the `COALESCE(post.kind, 'short') <> 'collection'` fix if you think this defense is depth is needed, thoughts?

Prod:
image
Staging:
image

ok300: Nice, in that case it's good as it is!

ok300 (approved):

Issue #860, opened 2026-05-11
Homeserver cursor regression on a Redis read hiccup


SHAcollision:
## TL;DR

On 2026-05-11 a ~600 ms network blip convinced Nexus it had never seen the watched homeserver before, so we obligingly bootstrapped its cursor back to `"0000000000000"` and spent the next ~30 minutes re-indexing a year's worth of events at 1000 per batch. That is how a year-old PUT showed up in users' notification feeds.

The bad write came out of `Homeserver::get_by_id`'s graph fallback at [`nexus-common/src/models/homeserver.rs:84-95`](https://github.com/pubky/pubky-nexus/blob/3f84f6aef63cd3e7546e5c0cc7a618fb0dbfdbc0/nexus-common/src/models/homeserver.rs#L84-L95). On what looked like a cache miss, we rebuilt the homeserver from Neo4j (which doesn't store the cursor), and wrote the result back to Redis without checking what was already there.

**PR #662 would have stopped this on its own.** It is the primary fix. A ~10-line adjacent change to `json::get` closes the theoretical residual race that #662 doesn't.

## What we saw in the logs (UTC)

Nine minutes of steady state at 500 ms ticks (`watcher_sleep: 500`), polls returning `[]`, "No new events", each tick ~5 ms. Then:

```
2026-05-11T11:11:55.463641Z DEBUG nexus_watcher::service:94: Indexing homeservers…
2026-05-11T11:11:55.625739Z  WARN nexus_common::db::graph::instrumented:173:
    Slow Neo4j query total_ms=161 execute_ms=76 fetch_ms=85 rows=1
    query=get_all_homeservers
2026-05-11T11:11:56.220471Z DEBUG nexus_common::db::kv::index::json:62:
    Set key: Homeserver:8um71us3… with optional expiration: None
2026-05-11T11:11:56.221918Z DEBUG …service::processor:61: Polling new events from homeserver
2026-05-11T11:11:56.247313Z DEBUG …service::processor:84: Homeserver response lines ["PUT pubky://…", …]
2026-05-11T11:11:56.248642Z  INFO …service::processor:43: Processing 1001 event lines
```

Three things pin the failure to the cache-miss fallback:

[...]

PR #866, opened 2026-05-15
fix: stop json::get swallowing errors; re-seed homeserver index from graph

SHAcollision:
## Summary

Two complementary fixes for the post-#861 homeserver indexing path, motivated by [#860](https://github.com/pubky/pubky-nexus/issues/860) and by the docker-testnet failure mode James hit in #nexus this week:

```
ERROR nexus_watcher::service::traits::tevent_processor_runner:127:
  Failed to build event processor for homeserver: 8pinx…: Homeserver not found
…
Skipping homeserver 8pinx… (in backoff)
```

…on every subsequent docker testnet stack run, with no way to recover except wiping local storage.

### 1. `nexus-common/src/db/kv/index/json.rs` — stop `json::get` collapsing Redis errors into `Ok(None)`

The `if let Ok(_)` shape on `json_get` discarded every failure mode — server errors, WRONGTYPE, I/O timeouts, deserialization quirks from the redis crate — into `Ok(None)`, making a transient hiccup indistinguishable from a clean cache miss. That's the silent-swallow #860 called out as the residual fix after #861 removed the `Homeserver::get_by_id` graph fallback.

Binds `json_get` to `Option<String>` so the redis crate's `FromRedisValue` decodes a genuine `Nil` to `None` and propagates every other error via `?`. Same shape as the existing `json_mget` call site a few lines down. No public API change.

The sibling `_get_bool` is `pub` and had the same silent-swallow shape (no callers today, but the divergent contract was a footgun for any future user); it gets the same `Option<i32>` + `?` rewrite for consistency.

This is also a precondition for fix 2: with errors propagating, an `Ok(None)` from `get_from_index` is finally a reliable signal that Redis is *genuinely* empty.

### 2. `nexus-common/src/models/homeserver.rs` — `persist_if_unknown` re-seeds the index when only the graph has the homeserver

`persist_if_unknown` only wrote when the graph was missing the homeserver. Since #861 stopped `get_by_id` from rebuilding the index from the graph, the asymmetric storage state — Neo4j volume persisted, Redis volume wiped — leaves the watcher stuck logging `Homeserver not found` on every tick and parking the homeserver in 60s/3600s backoff with no path back. That state happens every time a docker testnet operator tears one volume down without the other (the common "down -v on one stack, keep the other" workflow), and it's exactly what James reported:

> I'm seeing this on subsequent docker testnet stack runs and nexus doesn't seem to index homeserver. A quick fix is to wipe local storage and start fresh.

This PR restores the self-heal-on-restart behaviour the system was designed around — but only at the bootstrap call site (`NexusWatcher::start` and `maybe_ingest_for_user`), only as a one-shot write, never as a mid-flight `get_by_id` fallback. The watcher then re-indexes from `cursor=0`, which is the same recovery the system already performs after a Redis crash by design. The asymmetric branch logs a warning so operators can see when the heal fires.

Safety analysis:

- **Mid-flight cursor regression** (the bug #861 fixed): not possible. `persist_if_unknown` runs once per homeserver entry — at startup for the config HS, on first sighting for an external HS. It is not on the per-tick `get_by_id` path.
- **Transient Redis hiccup at startup**: with fix 1 in place, a Redis error propagates as `Err` and the re-seed never fires. Without fix 1, a hiccup could trigger a spurious cursor=0 re-seed — which is why the two changes ship together.

## Behaviour change & blast radius

`json::get` (and `_get_bool`):
- Genuine missing key → `Ok(None)` (unchanged).
- Redis error / WRONGTYPE / I/O / deserialization-from-redis → `Err(RedisError::*)` (was `Ok(None)`).

Caller audit walked every direct and indirect caller of `try_from_index_json` / `try_from_index_multiple_json` / `mget`. No caller has a "treat `Ok(None)` as fresh-install, write a default" pattern except `Homeserver::get_by_id` (already guarded by #861) and the bootstrap writes in `UserCounts::sync_put` / `PostCounts::sync_put`, which gate the default write behind a separate `is_none()` check after `?` propagation — a Redis error now aborts the put rather than writing a stale default. Route handlers using read-through caches will surface 5xx on Redis trouble instead of a misleading 404.

`persist_if_unknown`:
- New: graph has the row, index doesn't → re-seed the index with `cursor=0` and `warn!` (the testnet self-heal path).
- Unchanged: graph missing → write to both (first-time bootstrap).
- Unchanged: both present → no-op. Existing cursor is never overwritten.

## Tests

Six new tests, all passing locally (`cargo nextest run -p nexus-common` — 63/63):

`db::kv::index::json::tests`:
- `test_get_roundtrip_returns_some` — put + get round-trip returns `Ok(Some(_))`.
- `test_get_missing_key_returns_none` — unset key still returns `Ok(None)`.
- `test_get_on_wrong_type_propagates_err` — `SET` a plain string at the key, then `JSON.GET` it; asserts `Err(RedisError::CommandFailed(_))`. Direct regression test for #860; before fix 1, this returned `Ok(None)`.

`models::homeserver::tests`:
- `test_persist_if_unknown_first_time_install_writes_both` — preserved behaviour.
- `test_persist_if_unknown_reseeds_index_when_only_graph_has_it` — direct regression test for James's docker-testnet scenario. Before fix 2, the index stayed empty.
- `test_persist_if_unknown_is_noop_when_both_present` — seeds index with a non-default cursor (`1234567890123`), runs `persist_if_unknown`, asserts the cursor is preserved. Confirms steady-state runs don't overwrite real state.

### Verification

- [x] `cargo check --workspace`
- [x] `cargo nextest run -p nexus-common --no-fail-fast` — 63/63 passing including the 6 new tests
- [x] Manual repro of James's scenario passes via `test_persist_if_unknown_reseeds_index_when_only_graph_has_it`

## Out of scope

- PR [#662](https://github.com/pubky/pubky-nexus/pull/662) (`ok300-prevent-hs-reindexing`) adds the monotonic-cursor invariant; #860 calls for it as a separate primary fix and it should be rebased and merged on its own.
- Redis durability in `docker/docker-compose.yml` (default RDB, no AOF). Not changed here; the heal path makes the asymmetry recoverable instead of changing the persistence configuration.

### DIFF
--- a/nexus-common/src/db/kv/index/json.rs
+++ b/nexus-common/src/db/kv/index/json.rs
@@ -3,7 +3,7 @@ use crate::db::kv::error::{RedisError, RedisResult};
 use deadpool_redis::redis::Script;
 use deadpool_redis::redis::{AsyncCommands, JsonAsyncCommands};
 use serde::{de::DeserializeOwned, Serialize};
-use tracing::{debug, trace};
+use tracing::debug;
 
 #[derive(Clone, Debug)]
 pub enum JsonAction {
@@ -405,21 +405,17 @@ pub async fn _get_bool(prefix: &str, key: &str) -> RedisResult<Option<bool>> {
     let mut redis_conn = get_redis_conn().await?;
     let index_key = format!("{prefix}:{key}");
 
-    if let Ok(indexed_value) = redis_conn.get::<_, i32>(&index_key).await {
-        trace!(
-            "Restored boolean key: {} with value: {}",
-            index_key,
-            indexed_value
-        );
-        let value = match indexed_value {
-            1 => true,
-            0 => false,
-            _ => return Ok(None), // Invalid value in Redis
-        };
-        return Ok(Some(value));
-    }
+    // Match `get<T>`'s contract: a genuine miss decodes to None, every other
+    // failure propagates via `?`. The previous `if let Ok(_)` shape collapsed
+    // every error into `Ok(None)` (see #860).
+    let indexed_value: Option<i32> = redis_conn.get(&index_key).await?;
 
-    Ok(None)
+    match indexed_value {
+        Some(1) => Ok(Some(true)),
+        Some(0) => Ok(Some(false)),
+        Some(_) => Ok(None), // Invalid value in Redis
+        None => Ok(None),
+    }
 }
 
 /// Deletes multiple keys from Redis.
--- a/nexus-common/src/db/kv/index/json.rs
+++ b/nexus-common/src/db/kv/index/json.rs
@@ -305,19 +305,22 @@ pub async fn get<T: DeserializeOwned + Send + Sync>(
 ) -> RedisResult<Option<T>> {
     let mut redis_conn = get_redis_conn().await?;
     let index_key = format!("{prefix}:{key}");
-    let json_path = path.unwrap_or("$").to_string(); // Ensure path is a String
+    let json_path = path.unwrap_or("$");
 
-    // Use RedisJSON commands to get the value from the specified path
-    if let Ok(indexed_value) = redis_conn
-        .json_get::<String, String, String>(index_key.clone(), json_path)
-        .await
-    {
-        //debug!("Restored key: {} with value: {}", index_key, indexed_value);
-        let value: Vec<T> = from_json_str(&indexed_value)?;
-        return Ok(value.into_iter().next()); // Extract the first element from the Vec
+    // Bind to Option<String>: a genuine Nil reply (key absent) decodes to
+    // None; any other redis-side failure propagates via `?`. The previous
+    // `if let Ok(_)` shape collapsed every failure mode into `Ok(None)`,
+    // which let read-through caches mistake a transient error for a clean
+    // miss and bootstrap a default over real state (see #860).
+    let indexed_value: Option<String> = redis_conn.json_get(index_key, json_path).await?;
+
+    match indexed_value {
+        Some(s) => {
+            let value: Vec<T> = from_json_str(&s)?;
+            Ok(value.into_iter().next())
+        }
+        None => Ok(None),
     }
-
-    Ok(None)
 }
 
 /// Retrieves a list of JSON values from Redis based on a list of keys.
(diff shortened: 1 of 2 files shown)


nexus-common/src/db/kv/index/json.rs
```diff
@@ -305,19 +305,22 @@ pub async fn get<T: DeserializeOwned + Send + Sync>(
...
-        .await
-    {
-        //debug!("Restored key: {} with value: {}", index_key, indexed_value);
-        let value: Vec<T> = from_json_str(&indexed_value)?;
-        return Ok(value.into_iter().next()); // Extract the first element from the Vec
-    }
+    let json_path = path.unwrap_or("$");
 
-    Ok(None)
+    // Bind to Option<String>: a genuine Nil reply (key absent) decodes to
```
aintnostressin: I think this comment can be removed. 
The code is self explanatory and the commit will leave trace back to PR and github issue.

nexus-common/src/db/kv/index/json.rs
```diff
@@ -402,21 +405,17 @@ pub async fn _get_bool(prefix: &str, key: &str) -> RedisResult<Option<bool>> {
...
-            indexed_value
-        );
-        let value = match indexed_value {
-            1 => true,
-            0 => false,
-            _ => return Ok(None), // Invalid value in Redis
-        };
-        return Ok(Some(value));
-    }
+    // Match `get<T>`'s contract: a genuine miss decodes to None, every other
```
aintnostressin: I think this comment can also be removed.

nexus-common/src/db/kv/index/json.rs
```diff
@@ -402,21 +405,17 @@ pub async fn _get_bool(prefix: &str, key: &str) -> RedisResult<Option<bool>> {
...
+    // failure propagates via `?`. The previous `if let Ok(_)` shape collapsed
+    // every error into `Ok(None)` (see #860).
+    let indexed_value: Option<i32> = redis_conn.get(&index_key).await?;
 
-    Ok(None)
+    match indexed_value {
+        Some(1) => Ok(Some(true)),
+        Some(0) => Ok(Some(false)),
+        Some(_) => Ok(None), // Invalid value in Redis
```
aintnostressin: It may make sense to `trace::warn` here so we can detect invalid values in redis

nexus-common/src/models/homeserver.rs
aintnostressin: Should we change this comment ? 
Redis has a persist delay so it doesn't ensure we may get the same value as we previously stored.

aintnostressin (approved):

SHAcollision:
@aintnostressin Addressed your comments on the last commit

aintnostressin:
@SHAcollision thank you. Looks good to me :raised_hands:

catch-21:
Verified that if either the entire pubky-docker stack or just redis is taken down and brought back up again that nexus recovers and continues indexing.

PR #869, opened 2026-05-20
fix: bump RetryEvent redis key prefix to V2 for schema break

aintnostressin:
Due to `RetryEvent` schema change previously stored objects cannot be deserialized anymore. 
There is also no possibility of migrating data because previous model contained only serialized error.

### DIFF
--- a/nexus-watcher/src/events/retry/event.rs
+++ b/nexus-watcher/src/events/retry/event.rs
@@ -7,9 +7,11 @@ use nexus_common::db::RedisOps;
 
 use crate::events::EventProcessorError;
 
-const RETRY_MANAGER_PREFIX: &str = "RetryManager";
-const RETRY_MANAGER_EVENTS_INDEX: [&str; 1] = ["events"];
-const RETRY_MANAGER_STATE_INDEX: [&str; 1] = ["state"];
+// v2: RetryEvent schema changed incompatibly; bumped prefix so old keys under
+// "RetryManager:*" stay orphaned rather than failing to deserialize.
+pub const RETRY_MANAGER_PREFIX: &str = "RetryManagerV2";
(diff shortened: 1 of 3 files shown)


ok300 (approved):

ok300:
> There is also no possibility of migrating data because previous model contained only serialized error.

Not entirely true.

The project has a ready-made migration framework with a `delete_keys_by_pattern utility`, as seen in `nexusd/src/migrations/migrations_list/remove_muted_1771718400.rs`. A one-liner migration would be sufficient:

```rust
delete_keys_by_pattern("RetryManager:*").await?;
```

IMO not entirely necessary, but if you want to add a migration to prune the old data, that would be the way.

aintnostressin:
> > There is also no possibility of migrating data because previous model contained only serialized error.
> 
> Not entirely true.
> 
> The project has a ready-made migration framework with a `delete_keys_by_pattern utility`, as seen in `nexusd/src/migrations/migrations_list/remove_muted_1771718400.rs`. A one-liner migration would be sufficient:
> 
> ```rust
> delete_keys_by_pattern("RetryManager:*").await?;
> ```
> 
> IMO not entirely necessary, but if you want to add a migration to prune the old data, that would be the way.

I wouldn't call removing objects entirely as a migration.
What I meant here is that currently nexus stores only `error_type` and `retry_count`. The rest of the information required for retrying is not there and I see no way of bringing it back even through migration script.

ok300:
I understand, but the meaning of migration in Nexus also covers such pruning of unused data, as you can see by the existence of `delete_keys_by_pattern`. It's a migration step, even though no data is migrated _per se_.

I mention this to avoid confusion in the future, as such steps will often be referred to as "migrations".

PR #872, opened 2026-05-20
feat(stream): `?source=collection` on `/v0/stream/posts`

SHAcollision:
Adds `?source=collection&author_id=X&post_id=Y` to `/v0/stream/posts`, returning the posts curated inside a Collection's envelope in curator order, using the existing `PostStreamDetailed` response shape.

## Dispatch

1. Fetch the Collection's `PostDetails`, verify `kind=Collection` (else empty stream).
2. Parse `envelope.items`, slice by `skip`/`limit`.
3. Parse each URI; keep only `Resource::Post(_)`. Non-post URIs (legacy v0.5.0 envelopes) drop here at parse.
4. `from_listed_post_ids` resolves the keys; missing posts drop silently (returns `Option<PostView>`).

## Rejected params for `source=collection` (400 Bad Request)

`tags`, `kind`, `sorting`, `order`, `start`, `end` are all incompatible with curator order and explicit-reject so the FE catches misuse during development. `validate_source_compat()` runs **before** `initialize_defaults()` so we see the user's raw input (not the defaulted form).

Accepted: `skip`, `limit`, `viewer_id`, `include_attachment_metadata`.

## FE pattern for the curator's broken-item UX

1. FE fetches `GET /v0/post/{author_id}/{post_id}` (existing) for Collection metadata + `envelope.items[]`.
2. FE fetches `GET /v0/stream/posts?source=collection&author_id={author_id}&post_id={post_id}` for resolved `PostView` list.
3. (parallel, same wall-time as one combined call).
4. To surface dead refs to the curator: client-side diff `envelope.items[]` against the returned post URIs, render tombstones inline, curator clicks "remove broken", FE re-PUTs the Collection envelope with those URIs stripped from `items[]`.

## Test coverage

12 integration tests on `source=collection`:
- Curator order preserved + skip/limit pagination
- Missing posts + non-post URIs silently dropped (load-bearing)
- Empty Collection returns empty stream
- Skip past end returns empty
- Non-Collection source returns empty; unknown post returns empty
- Malformed envelope JSON returns empty (defensive parse, no 500)
- `viewer_id` flows through to bookmark hydration
- Collection-of-Collections: nested Collection URI in items resolves as kind=collection entry
- All 6 rejected params each verified with 400

## Files

- `nexus-common/src/models/post/stream.rs`: new `StreamSource::Collection` variant + `can_use_index` short-circuit + `get_collection_items_post_keys` helper
- `nexus-common/src/db/graph/queries/get.rs`: `unreachable!()` on the impossible Cypher arm
- `nexus-webapi/src/routes/v0/stream/posts.rs`: `validate_source_compat()` + `SOURCE_PARAM_DESC` const for OpenAPI dedup
- `nexus-webapi/tests/stream/post/kind/collection.rs`: 12 new tests
- `docker/test-graph/mocks/posts.cypher`: seed, `COLW1TGL5BKG1.items` populated, plus `MALF1TGL5BKG7` (malformed envelope) and `NEST1TGL5BKG8` (Collection-of-Collections)

~+380 LOC, no breaking changes to existing stream callers.

### DIFF
--- a/docker/test-graph/mocks/posts.cypher
+++ b/docker/test-graph/mocks/posts.cypher
@@ -235,7 +235,7 @@ MATCH (reply:Post {id: "SIJW1TGL5BKG9" }), (parent:Post {id: "SIJW1TGL5BKG8" }) 
 // Collections are kind="collection" with a JSON envelope ({name, description?, items})
 // in `content`. Timestamps placed below other kinds so they don't disturb existing
 // timeline/bootstrap test fixtures.
-MERGE (p:Post {id: "COLW1TGL5BKG1"}) SET p.content = "{\"name\":\"AI papers\",\"description\":\"Best stuff\",\"items\":[]}", p.kind = "collection", p.indexed_at = 1980477299101;
+MERGE (p:Post {id: "COLW1TGL5BKG1"}) SET p.content = "{\"name\":\"AI papers\",\"description\":\"Best stuff\",\"items\":[\"pubky://ep441mndnsjeesenwz78r9paepm6e4kqm4ggiyy9uzpoe43eu9ny/pub/pubky.app/posts/A5D6P9V3Q0T\",\"pubky://ep441mndnsjeesenwz78r9paepm6e4kqm4ggiyy9uzpoe43eu9ny/pub/pubky.app/posts/C3L7W0F9Q4K8\",\"pubky://ep441mndnsjeesenwz78r9paepm6e4kqm4ggiyy9uzpoe43eu9ny/pub/pubky.app/posts/K1P6Q9M2X4J8\",\"pubky://ep441mndnsjeesenwz78r9paepm6e4kqm4ggiyy9uzpoe43eu9ny/pub/pubky.app/posts/ZZZZZZZZZZZZZ\",\"pubky://ep441mndnsjeesenwz78r9paepm6e4kqm4ggiyy9uzpoe43eu9ny/pub/pubky.app/profile.json\"]}", p.kind = "collection", p.indexed_at = 1980477299101;
 MATCH (u:User {id: $bogota}), (p:Post {id: "COLW1TGL5BKG1"}) MERGE (u)-[:AUTHORED]->(p);
 MERGE (p:Post {id: "COLW1TGL5BKG2"}) SET p.content = "{\"name\":\"Privacy reads\",\"items\":[]}", p.kind = "collection", p.indexed_at = 1980477299111;
 MATCH (u:User {id: $bogota}), (p:Post {id: "COLW1TGL5BKG2"}) MERGE (u)-[:AUTHORED]->(p);
@@ -245,6 +245,15 @@ MATCH (u:User {id: $cairo}), (p:Post {id: "COLW1TGL5BKG3"}) MERGE (u)-[:AUTHORED
 // Tag on a Collection (used by `?tags` stream tests). Bogota tags Cairo's
 // Collection with the `api` label.
 MATCH (u:User {id: $bogota}), (p:Post {id: "COLW1TGL5BKG3"}) MERGE (u)-[:TAGGED {label: $api_tag, id: "7APICOLBKG3Z", indexed_at: 1980477299130}]->(p);
+
+// Defensive seed: malformed envelope JSON. The resolver must `warn!` and
+// return an empty stream rather than 500 on corrupt graph state.
+MERGE (p:Post {id: "MALF1TGL5BKG7"}) SET p.content = "this is not JSON", p.kind = "collection", p.indexed_at = 1980477299430;
+MATCH (u:User {id: $bogota}), (p:Post {id: "MALF1TGL5BKG7"}) MERGE (u)-[:AUTHORED]->(p);
+
+// Collection-of-Collections: items[] references another Collection's post URI.
+MERGE (p:Post {id: "NEST1TGL5BKG8"}) SET p.content = "{\"name\":\"Meta\",\"items\":[\"pubky://ep441mndnsjeesenwz78r9paepm6e4kqm4ggiyy9uzpoe43eu9ny/pub/pubky.app/posts/COLW1TGL5BKG1\"]}", p.kind = "collection", p.indexed_at = 1980477299440;
+MATCH (u:User {id: $bogota}), (p:Post {id: "NEST1TGL5BKG8"}) MERGE (u)-[:AUTHORED]->(p);
 
 // Bookmarked collection (used by `?source=bookmarks` stream tests). Eixample
 // bookmarks Bogota's COLW1TGL5BKG1.
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -1,3 +1,4 @@
+use crate::db::graph::error::{GraphError, GraphResult};
 use crate::db::graph::Query;
 use crate::db::kv::SortOrder;
 use crate::models::post::StreamSource;
@@ -783,7 +784,7 @@ pub fn post_stream(
     tags: &Option<Vec<String>>,
     pagination: Pagination,
     kind: Option<PubkyAppPostKind>,
-) -> Query {
+) -> GraphResult<Query> {
     // Initialize the cypher query
     let mut cypher = String::new();
 
@@ -929,20 +930,30 @@ pub fn post_stream(
     }
 
     // Build the query and apply parameters using `param` method
-    let query = Query::new(
-        match &source {
-            StreamSource::Following { .. } => "post_stream_following",
-            StreamSource::Followers { .. } => "post_stream_followers",
-            StreamSource::Friends { .. } => "post_stream_friends",
-            StreamSource::Bookmarks { .. } => "post_stream_bookmarks",
-            StreamSource::Author { .. } => "post_stream_author",
-            StreamSource::AuthorReplies { .. } => "post_stream_author_replies",
-            StreamSource::PostReplies { .. } => "post_stream_post_replies",
-            StreamSource::All => "post_stream_all",
-        },
-        &cypher,
-    );
-    build_query_with_params(query, &source, tags, kind, &pagination)
+    let query_name = match &source {
+        StreamSource::Following { .. } => "post_stream_following",
+        StreamSource::Followers { .. } => "post_stream_followers",
+        StreamSource::Friends { .. } => "post_stream_friends",
+        StreamSource::Bookmarks { .. } => "post_stream_bookmarks",
+        StreamSource::Author { .. } => "post_stream_author",
+        StreamSource::AuthorReplies { .. } => "post_stream_author_replies",
+        StreamSource::PostReplies { .. } => "post_stream_post_replies",
+        // Short-circuited upstream in collect_post_keys.
+        StreamSource::Collection { .. } => {
+            return Err(GraphError::QueryBuildError(
+                "StreamSource::Collection must be served by collect_post_keys".to_string(),
+            ));
+        }
+        StreamSource::All => "post_stream_all",
+    };
+    let query = Query::new(query_name, &cypher);
+    Ok(build_query_with_params(
+        query,
+        &source,
+        tags,
+        kind,
+        &pagination,
+    ))
 }
 
 /// Appends a condition to the Cypher query, using `WHERE` if no `WHERE` clause
--- a/nexus-common/src/models/post/stream.rs
+++ b/nexus-common/src/models/post/stream.rs
@@ -9,7 +9,7 @@ use crate::models::{
 };
 use crate::types::{Pagination, StreamSorting};
 use futures::TryStreamExt;
-use pubky_app_specs::PubkyAppPostKind;
+use pubky_app_specs::{ParsedUri, PubkyAppCollectionContent, PubkyAppPostKind, Resource};
 use serde::{Deserialize, Serialize};
 use tokio::task::spawn;
 use tokio::time::{timeout, Duration};
@@ -48,6 +48,10 @@ pub enum StreamSource {
     AuthorReplies {
         author_id: String,
     },
+    Collection {
+        author_id: String,
+        post_id: String,
+    },
     #[default]
     All,
 }
@@ -63,6 +67,8 @@ impl StreamSource {
         }
     }
 
+    /// Author whose posts are streamed. Collection returns `None`: its
+    /// `author_id` is the curator, not the items' authors.
     pub fn get_author(&self) -> Option<&str> {
         match self {
             StreamSource::PostReplies {
@@ -157,6 +163,18 @@ impl PostStream {
         tags: Option<Vec<String>>,
         kind: Option<PubkyAppPostKind>,
     ) -> ModelResult<PostKeyStream> {
+        // Collection has its own envelope-driven resolution path (neither
+        // sorted-set index nor Cypher).
+        if let StreamSource::Collection { author_id, post_id } = &source {
+            return Self::get_collection_items_post_keys(
+                author_id,
+                post_id,
+                pagination.skip,
+                pagination.limit,
+            )
+            .await;
+        }
+
         // Decide whether to use index or fallback to graph query
         let use_index = Self::can_use_index(&sorting, &source, &tags, &kind);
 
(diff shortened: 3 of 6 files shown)


nexus-webapi/src/routes/v0/stream/posts.rs
```diff
@@ -82,7 +105,7 @@ where
     path = STREAM_POSTS_ROUTE,
     tag = "Stream",
     params(
-        ("source" = Option<StreamSource>, Query, description = "Source of posts for streams with viewer (following, followers, friends, bookmarks, post_replies, author, author_replies, all)"),
+        ("source" = Option<StreamSource>, Query, description = SOURCE_PARAM_DESC),
```
ok300: I would not extract this to a separate `const`, but keep it here (inline / multiline).

If it's extracted, it "breaks immersion": you'd have to constantly jump up and down to read the `const` of various fields, if you're trying to get an overview of the API endpoint decscription.

nexus-common/src/models/post/stream.rs
```diff
@@ -247,6 +259,49 @@ impl PostStream {
...
+            .filter_map(|uri| {
+                ParsedUri::try_from(uri.as_str())
+                    .ok()
+                    .and_then(|p| match p.resource {
+                        Resource::Post(item_post_id) => {
+                            Some(format!("{}:{}", p.user_id, item_post_id))
+                        }
+                        _ => None,
+                    })
+            })
```
ok300:
```suggestion
            .filter_map(|uri| match ParsedUri::try_from(uri.as_str()) {
                Ok(p) => match p.resource {
                    Resource::Post(item_post_id) => Some(format!("{}:{}", p.user_id, item_post_id)),
                    _ => None,
                },
                Err(_) => None,
            })
```

The result handling can be simplified.

nexus-common/src/models/post/stream.rs
```diff
@@ -175,6 +179,10 @@ impl PostStream {
         tags: &Option<Vec<String>>,
         kind: &Option<PubkyAppPostKind>,
     ) -> bool {
+        // Collection resolves items from the envelope JSON, not a sorted set.
+        if matches!(source, StreamSource::Collection { .. }) {
+            return true;
+        }
```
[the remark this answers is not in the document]
SHAcollision: Good catch. Fixed in 0b189ed0 by moving Collection out of the index/graph dispatch entirely: `collect_post_keys` now short-circuits to `get_collection_items_post_keys` BEFORE `can_use_index` runs, so `can_use_index` returns to its honest "Redis sorted-set" semantics. Collection special-cases are gone from `can_use_index` and `get_from_index`. The `unreachable!()` arm in the Cypher `Query::new` match stays as documentation that Collection truly never reaches that code path.

nexus-common/src/models/post/stream.rs
```diff
@@ -247,6 +259,49 @@ impl PostStream {
...
+                ParsedUri::try_from(uri.as_str())
+                    .ok()
+                    .and_then(|p| match p.resource {
+                        Resource::Post(item_post_id) => {
+                            Some(format!("{}:{}", p.user_id, item_post_id))
+                        }
+                        _ => None,
+                    })
+            })
+            .collect();
```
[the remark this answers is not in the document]
SHAcollision: Fair. Reordered to `filter_map().skip().take()` in 0b189ed0, so dead refs interspersed in the items array don't shorten pages. The FE diffs by URI (not by index) so positional alignment with raw `items[]` isn't required.

One source of short pages remains and is unavoidable: a post can be deleted between when `from_listed_post_ids` builds the key list and when it fetches `PostView`s. Documented that in `SOURCE_PARAM_DESC` ("pages may be shorter than `limit` when posts have been deleted since the envelope was last edited").

nexus-common/src/models/post/stream.rs
```diff
@@ -247,6 +263,52 @@ impl PostStream {
         Ok(result)
     }
 
+    async fn get_collection_items_post_keys(
```
ok300, line 268: Doesn't this presume the collection necessarily contains Post URIs?

Several inner variables also hint to that (`post_keys`, etc). In fact the logic to build `post_keys` specifically excludes non-posts.

As per the PR description:

> Parse each URI; keep only Resource::Post(_). Non-post URIs (legacy v0.5.0 envelopes) drop here at parse.

I think the intention here was to drop non-post Collection entities, but what the code actually does is it drops the non-post collection items (inside the collection).

SHAcollision: > Doesn't this presume the collection necessarily contains Post URIs?

Indeed, it does. In fact, it must be only Post URIs. I met FE team yesterday and the only way collections work is if all items are Pubky App Posts, see sibling PR here: https://github.com/pubky/pubky-app-specs/pull/122

nexus-webapi/src/routes/v0/stream/posts.rs
SHAcollision: Agree, applied in 5cfe0e29

nexus-common/src/models/post/stream.rs
SHAcollision: Applied in 5cfe0e29.

SHAcollision:
@greptile review again

aintnostressin:
Why can't we keep collections flexible and if posts are needed fetch them by `/posts/by_ids` ?

This question takes https://github.com/pubky/pubky-app-specs/pull/122/ into consideration.

docker/test-graph/mocks/posts.cypher
```diff
@@ -236,13 +236,24 @@ MATCH (reply:Post {id: "SIJW1TGL5BKG9" }), (parent:Post {id: "SIJW1TGL5BKG8" })
...
 MATCH (u:User {id: $bogota}), (p:Post {id: "COLW1TGL5BKG2"}) MERGE (u)-[:AUTHORED]->(p);
 MERGE (p:Post {id: "COLW1TGL5BKG3"}) SET p.content = "{\"name\":\"Cryptography classics\",\"items\":[]}", p.kind = "collection", p.indexed_at = 1980477299420;
 MATCH (u:User {id: $cairo}), (p:Post {id: "COLW1TGL5BKG3"}) MERGE (u)-[:AUTHORED]->(p);
 
+// Defensive seed: malformed envelope JSON. The resolver must `warn!` and
+// return an empty stream rather than 500 on corrupt graph state.
+MERGE (p:Post {id: "MALF1TGL5BKG7"}) SET p.content = "this is not JSON", p.kind = "collection", p.indexed_at = 1980477299430;
+MATCH (u:User {id: $bogota}), (p:Post {id: "MALF1TGL5BKG7"}) MERGE (u)-[:AUTHORED]->(p);
+
+// Collection-of-Collections: items[] references another Collection's post URI.
```
ok300, line 254: Are nested Collections allowed?

If so, can there be loops / graph cycles?
- simple: Collection C1 includes C1 as item
- nested: C1 incl. C2 incl. C1

I imagine this can mess up FE rendering and / or notifications.

Regardless if they're allowed or not, IMO we should make this explicit (tests, docs) to avoid surprises.

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -951,6 +951,11 @@ pub fn post_stream(
             StreamSource::Author { .. } => "post_stream_author",
             StreamSource::AuthorReplies { .. } => "post_stream_author_replies",
             StreamSource::PostReplies { .. } => "post_stream_post_replies",
+            // Collection never reaches Cypher fallback (can_use_index short-circuits to true).
+            // Panic if the invariant breaks rather than emitting a fake query name.
```
ok300: Wouldn't it be better to return a `GraphResult` instead?

Instead of panic it could throw a `GraphError::QueryBuildError`.

nexus-common/src/models/post/stream.rs
```diff
@@ -247,6 +263,49 @@ impl PostStream {
...
+            }
+        };
+
+        let skip = skip.unwrap_or(0);
+        let limit = limit.unwrap_or(usize::MAX);
+
+        // Filter before slicing: dead-ref shorts from non-post URIs don't
+        // bleed into pagination (the FE diffs by URI, not by index, so
+        // positional alignment with raw items[] isn't required).
+        let post_keys: Vec<String> = envelope
```
ok300, line 292: THis filters collection `items[]` only by parseable `Resource::Post`, then returns keys directly at line 306. It does not verify that each referenced post still exists.

This method is called via `PostStream::collect_post_keys` in:
- `PostStream::get_posts`
- `PostStream::get_post_keys`

The first one silently drops missing posts with `from_listed_post_ids`, while the second doesn't.

This breaks the existing endpoint invariant that post keys align with the posts stream, and contradicts the new OpenAPI text in `nexus-webapi/src/routes/v0/stream/posts.rs:170` saying missing/deleted items are silently dropped.

Suggested fix: either resolve/existence-check collection keys before returning `PostKeyStream`, or explicitly document/test that `/posts/keys` route may include unresolved collection refs. The former seems consistent with the PR description and existing `post_keys.rs` alignment tests.

nexus-common/src/models/post/stream.rs
ok300, line 72: `get_author()` / `get_observer()` don't handle `Collection`. Both fall through to `_ => None`.

For `Collection { author_id, .. }`, `get_author` returning `None` may be surprising if any downstream call-site uses it for logging or routing.

Is this intentional? If so, a short comment on `get_author` / `get_observer` would be helpful.

nexus-webapi/tests/stream/post/kind/collection.rs
```diff
@@ -235,3 +236,256 @@ async fn test_bookmarks_engagement_sort_includes_bookmarked_collection() -> Resu
     );
     Ok(())
 }
+
+// ---------------------------------------------------------------------------
+// ?source=collection: items inside a Collection
```
ok300: There's no test for `?source=collection` without `author_id` or `post_id`.

nexus-webapi/src/routes/v0/stream/posts.rs
```diff
@@ -58,6 +58,27 @@ impl PostStreamQuery {
         }
         Ok(())
     }
+
+    /// Must run before `initialize_defaults()` to see the user's raw input.
+    pub fn validate_source_compat(&self) -> AppResult<()> {
```
aintnostressin: This seems to be inconsistent with current validation logic.
Tags are also ignored for `StreamSource::PostReplies` but error is not returned.
So let's return error in case of junk params in all `StreamSource` variants or don't do it at all.

Can we add this logic in a separate PR if needed ?

nexus-webapi/src/routes/v0/stream/posts.rs
aintnostressin, line 154: Should this list be extended ?

SHAcollision:
> Why can't we keep collections flexible and if posts are needed fetch them by `/posts/by_ids` ?
> 
> This question takes [pubky/pubky-app-specs#122](https://github.com/pubky/pubky-app-specs/pull/122) into consideration.

The FE consumers explicitly asked for this endpoint shape, treating Collections as a stream source (like Home/Hot/Search) lets Nexus drive a "New Posts" notification on a Collection feed, which `/posts/by_ids` can't do because it's stateless on the BE side. The FE also gets parallel fetch (Collection metadata + items) instead of sequential, and avoids reimplementing pagination/cursor logic. Without this endpoint, we set back Collection feature release significantly.

docker/test-graph/mocks/posts.cypher
SHAcollision: The BE returns a flat, single-level list of post keys regardless of whether items are themselves Collections, `get_collection_items_post_keys` does not recurse. Cycles (C1 → C2 → C1) are therefore safe at this layer: the response is always one hop, no infinite loop.

 `test_source_collection_accepts_nested_collection_post_uris` documents the contract. Cycle handling during rendering is a FE concern, only one collection is expanded at a time according to design.

nexus-common/src/db/graph/queries/get.rs
SHAcollision: Done in 6ecdd285. `post_stream` now returns `GraphResult<Query>`; the Collection arm returns `GraphError::QueryBuildError` instead of panicking. Only one caller (`PostStream::get_from_graph`) needed `?`.

nexus-common/src/models/post/stream.rs
SHAcollision: Addressed in 6ecdd285 via OpenAPI clarification rather than an EXISTS sweep. Reasoning: every other `source` on `/stream/post_keys` already returns best-effort keys, ZSET / graph snapshots lag
  deletions in the same way the Collection envelope does, so adding an existence sweep only for Collection would make it the odd-one-out and force callers to write source-specific handling.

  The split OpenAPI text now says:
  - `/stream/posts?source=collection`: missing items dropped during hydration; pages may be shorter than `limit`.
  - `/stream/post_keys?source=collection`: like every other source, returned keys may reference deleted/non-indexed posts; callers should hydrate.

Also added a note that `skip`/`limit` pagination isn't stable across deletions (true for all sources, but worth calling out since curator order makes the window shift more visible to clients).

nexus-common/src/models/post/stream.rs
SHAcollision: Yes, intentional. `author_id` in `StreamSource::Collection` identifies the *curator* of the Collection, not the authors of the streamed items (which can be heterogeneous). Added a docstring on `get_author` and a 1-line comment on `get_observer` in 6ecdd285 explaining this.

nexus-webapi/tests/stream/post/kind/collection.rs
SHAcollision: Added two tests in 6ecdd285:  
  - `test_source_collection_without_author_id_rejected_400`
  - `test_source_collection_without_post_id_rejected_400`
  
  Serde rejects the tagged-enum variant before the handler runs -> 400.

nexus-webapi/src/routes/v0/stream/posts.rs
SHAcollision: Agree, extending this strictness to every other variant (e.g. rejecting `tags` on `PostReplies`) is a separate refactor and probably needs design alignment first (do we 400 silently-ignored params everywhere, or document them as ignored?). Should be filed as a follow-up; keeping Collection-only strict here for now.

nexus-webapi/src/routes/v0/stream/posts.rs
SHAcollision: Done in 6ecdd285. Added `*collection*: Requires **author_id** and **post_id** of the Collection post; items are returned in curator order.` to both `stream_posts_handler` and `stream_post_keys_handler` docstrings.

nexus-webapi/src/routes/v0/stream/posts.rs
aintnostressin: I would still suggest not adding it at all or cover all variants.

ok300:
@greptile please review again

ok300 (approved):
LGTM, with 2 remarks:

- The question about cycles: IMO its less about "who has to deal with the cycles (Nexus or FE)", but more about "does the spec allow or deny cycles". An answer to that would clarify it for all parties involved. I'm still not sure if the spec mentions this, meaning it will probably be a surprise and bring unknown / unexpected side-effects (for example: if there is a cycle and a user follows the root Collection, will a Notification caused by a change in an inner Collection cause endless loop / notification storm?)
- There are merge conflicts vs `main`

SHAcollision:
> * The question about cycles: IMO its less about "who has to deal with the cycles (Nexus or FE)", but more about "does the spec allow or deny cycles". An answer to that would clarify it for all parties involved. I'm still not sure if the spec mentions this, meaning it will probably be a surprise and bring unknown / unexpected side-effects (for example: if there is a cycle and a user follows the root Collection, will a Notification caused by a change in an inner Collection cause endless loop / notification storm?)

Aah, I get it now. The spec currently doesn't forbid cycles: `PubkyAppCollectionContent` validation is per-post, and cross-post cycle detection would require global graph traversal which isn't in the validator's scope.... So self-reference and multi-step cycles are both technically valid today.

On the notification-storm scenario: no risk in current Nexus, notifications fire per-post (follow C1 -> notified when C1 changes, not when C1's items change). The "follow a Collection -> notified about descendants" semantics doesn't exist (and will never?).

Picking a policy, and following the "tighten now, loosen later via soft-fork" principle, I'd propose: **items[] SHOULD NOT include Collection posts**. That rule would eliminate any cycle related problem. The spec validator can't enforce it (no network access in the pure validator), but the FE write path can, and we can "soft-fork: to allow nested Collections later if the use case appears (lmao "Best Collections of 2026"). I guess we should file this as a spec issue against on `pubky-app-specs`; happy to push back if you see a use case for nested Collections we'd lose by tightening.

PR #877, opened 2026-05-22
feat: fetch resource from event origin HS

aintnostressin:
Watcher fetches now hit the event-origin homeserver directly. Previously every  `public_storage().get(pubky_uri)` resolved the *user's* pkarr record to find their homeserver — but the watcher already knows which homeserver emitted the event, so this was redundant indirection and racy: if a user re-pointed their pkarr record between event emission and our fetch, we'd silently read from a different homeserver than the one that produced the event.


aintnostressin:
@greptile review

nexus-watcher/src/events/handlers/file.rs
```diff
@@ -51,9 +51,10 @@ async fn ingest(
     file_id: &str,
     pubkyapp_file: &PubkyAppFile,
     files_path: PathBuf,
+    homeserver_id: &PubkyId,
 ) -> Result<FileMeta, EventProcessorError> {
-    let pubky = PubkyConnector::get()?;
-    let response = pubky.public_storage().get(&pubkyapp_file.src).await?;
+    let client = HomeserverClient::new(homeserver_id.clone())?;
```
SHAcollision: nit (optional): for file PUT events we construct `HomeserverClient` twice, once in `handle_put_event` for the metadata, again here for the blob. Both calls do `PubkyConnector::get()?` + `client.clone()`, so it's cheap (Arc), but you could thread the client down from `handle_put_event` -> `sync_put` -> `ingest` and drop the second construction. Happy to defer to a follow-up.

SHAcollision (approved):
LGTM

nexus-watcher/src/events/handlers/file.rs
aintnostressin: True - I'll improve the code so `HomeserverClient` is constructed once per event and passed down through call stack.

This is just a draft - if the idea is fine I'll invest more time and improve code.

aintnostressin:
@greptile review again

ok300:
I'd wait with this until DX, this code-path is refactored there and it brings logic similar to this PR.

aintnostressin:
@ok300 which logic is similar?

If it's duplicate I'll close this one, if not let's simply merge it - DX branch is open since February.

ok300:
It's not duplicate, it is refactored because here it's assumed there is 1 HS, on DX there are 2 kinds of processors, both fetching events in different ways (main HS, fetching from `/events` + externals fetching from `/event-stream`). Merging this into `main` will only cause more rebase conflicts, so I would suggest
- either wait until DX is ready, then see if it still makes sense, or
- point this PR onto DX, at the right place (this change looks relevant for the `HsEventProcessor`, maybe not so much the `KeyBasedEventProcessor`)

aintnostressin:
@ok300 let's keep it open and wait for DX to be merged first.

nexus-watcher/src/events/homeserver_client.rs
```diff
@@ -0,0 +1,44 @@
...
+        Ok(Self {
+            homeserver_id,
+            client: pubky.client().clone(),
+        })
+    }
+
+    /// GET a `pubky://<user>/<path>` URI from the bound homeserver.
+    ///
+    /// Errors on a non-success HTTP status, mirroring `public_storage().get()`.
+    pub async fn get(&self, pubky_uri: &str) -> pubky::Result<Response> {
```
ok300, line 32: Every caller uses the returned `reqwest::Response` to then call

```rust
response.bytes().await
```

Wouldn't it then make more sense to include that in this method `HomeserverClient::get` and directly reteurn the bytes, such that we don't need to import `reqwest::Response` for this return type, such that we therefore don't need to add a dependency on `reqwest`?

nexus-watcher/src/events/handlers/file.rs
```diff
@@ -51,9 +59,9 @@ async fn ingest(
     file_id: &str,
     pubkyapp_file: &PubkyAppFile,
     files_path: PathBuf,
+    homeserver_client: Arc<HomeserverClient>,
 ) -> Result<FileMeta, EventProcessorError> {
-    let pubky = PubkyConnector::get()?;
-    let response = pubky.public_storage().get(&pubkyapp_file.src).await?;
+    let response = homeserver_client.get(&pubkyapp_file.src).await?;
```
ok300, line 64: AFAIK, the `blob` behind a `file` , referenced here via `pubkyapp_file.src`, doesn't necessarily have to be on the same HS and belong to the same user.

If user A on HS-1 references a blob from user B on HS-2, the request hits HS-1 instead of HS-2 and gets a 404 — which becomes a retryable `ClientError` in the current error mapping, causing unnecessary retries on a resource that will never exist on that server.

(It was also flagged as a P1 finding by Greptile, which I guess accidentally got marked as resolved; I un-resolved it for reference)

nexus-watcher/src/events/handlers/universal_tag.rs
```diff
@@ -31,27 +33,16 @@ pub async fn try_handle(
...
-        let body = response
-            .text()
-            .await
-            .unwrap_or_else(|_| "<unable to read body>".to_string());
-        return Err(EventProcessorError::client_error(format!(
-            "Fetch universal tag failed {}: HTTP {status} - {body}",
-            info.uri
-        )));
-    }
+async fn handle_put(
```
ok300, line 41: Error observability regression:

The old code in `handle_put_event()` and `universal_tag.rs` logged the full HTTP response body on failure:
```rust
let body = response.text().await.unwrap_or_else(|_| "<unable to read body>".to_string());
return Err(EventProcessorError::client_error(format!(
    "Fetch resource failed {}: HTTP {status} - {body}", event.uri
)));
```

Now `HomeserverClient::get()` uses `reqwest::Response::error_for_status()`, which only includes the status code and a generic message — no response body. For production debugging (especially 403s, 429s, etc.), the body context is valuable.

Consider preserving the body text in the error variant.

nexus-watcher/src/events/handlers/file.rs
ok300, line 64: Loss of ICANN-endpoint fallback:

The old `public_storage().get()` routes through `client.cross_request()` in `pubky-core` ([`actors/storage/core.rs:60`](https://github.com/pubky/pubky-core/blob/06f242ce21075dd0a6e7d784acaf1d90f9be21dd/pubky-sdk/src/actors/storage/core.rs#L60)), which does endpoint resolution plus TCP-probing the direct endpoint and falling back to the ICANN domain when a homeserver publishes one.

`HomeserverClient::get()` instead calls the lower-level `client.request(...)` ([`http_targets/native.rs:258`](https://github.com/pubky/pubky-core/blob/06f242ce21075dd0a6e7d784acaf1d90f9be21dd/pubky-sdk/src/client/http_targets/native.rs#L258)), which only classifies the host (`HostKind::Pubky` → `PubkyTLS/self.http`) and does not perform the ICANN fallback.

For a HS reachable only via an ICANN domain endpoint, the old path worked and the new one would likely fail TLS/connect.

ok300 (changes requested):
Found some regressions, see below.

I appreciate the effort, but I'm not a big fan of this PR. For a very small gain, it introduces brittleness and fragility in how internally requests are routed and resolved. It assumes a narrow use-case of how we fetch events and optimizes for that. Yes it might work now (assuming the regressions above are fixed) but we don't know what kind of `Resources` we'll be resolving soon. Maybe for some collection types, we have to resolve and fetch the inner URIs (which might live on a different HS? Maybe for some types of Universal Tag, we have to fetch the tagged resource (possibly on different HS)? etc

I would generally not go ahead with this, but if you insist, I'd suggest to restrict this optimization to only the kinds of `Resources` where we are absolutely sure it can't touch another HS. (Another scenario: posts with attachments that belong to different user, on different HS.. Endless such possibilities.)

nexus-watcher/src/events/handlers/file.rs
aintnostressin: Thanks, that's a very good feedback.

@SHAcollision I see you created `PubkyHttpClient` in sdk.
What do you think about adding a method so we can specify homeserver public key and user public key. It would build a request to a resource on a known homeserver, on behalf of  `user`, without resolving the user's own pkarr record.

aintnostressin:
> Found some regressions, see below.
> 
> I appreciate the effort, but I'm not a big fan of this PR. For a very small gain, it introduces brittleness and fragility in how internally requests are routed and resolved. It assumes a narrow use-case of how we fetch events and optimizes for that. Yes it might work now (assuming the regressions above are fixed) but we don't know what kind of `Resources` we'll be resolving soon. Maybe for some collection types, we have to resolve and fetch the inner URIs (which might live on a different HS? Maybe for some types of Universal Tag, we have to fetch the tagged resource (possibly on different HS)? etc
> 
> I would generally not go ahead with this, but if you insist, I'd suggest to restrict this optimization to only the kinds of `Resources` where we are absolutely sure it can't touch another HS. (Another scenario: posts with attachments that belong to different user, on different HS.. Endless such possibilities.)

  Thanks for the thorough review.
  
  I see it the other way on routing: fetching from the homeserver that emitted the event is the logically correct model that HS is the source of truth for what it served. Resolving the author's pkarr instead is convenient but can't always be
  accurate, since the user can re-point their record between emission and our fetch. It's invisible today because we index a single homeserver, but third-party HS ingestion and the retry queue (deferred fetches) are exactly where pkarr resolution starts reading the wrong source.

That said - if you still think resolving via the user's homeserver is the correct approach, I'm fine keeping it as it is.

PR #726, opened 2026-02-15
feat: Call user events endpoint when indexing 3rd party HSs

ok300:
This PR updates HS indexing logic to rely on the new user events endpoint on the HS.

This PR acts as an "umbrella PR" for this DX feature. It is meant to contain all relevant changes and sub-PRs. It will remain a draft until the full feature is ready for final review.

Steps:

- [x] Bump `pubky` to relevant feature branch (done in 6a06e9cbabc14b932fbf2ec9279fd298327bba61)
  - https://github.com/pubky/pubky-core/pull/311
- [x] Bump `pubky` to `v0.7.x` (done in c6513d7a15508dc22c61242f70cf4340bcb7a99d)
- [x] Restructure `NexusWatcher::start` to start 3 threads, each with own `tick`: index default HS, index other HSs, resolve user-to-HS mappings
- [x] Implement "index default HS" logic
  - [x] Only process events of an user if that user's HS is explicitly set to the default HS (e.g. do not blindly ingest all events)
    - https://github.com/pubky/pubky-nexus/pull/889
    - https://github.com/pubky/pubky-nexus/pull/902
    - [x] Double-check how `remove_user_homeserver` (introduced in [PR 791](https://github.com/pubky/pubky-nexus/pull/791)) works with the above + disallowing HS switching
  - [x] If a user originally bound to the default HS later changes or unpublishes their HS, we should “pause indexing” for this user's default-HS events. See:
    - #884
    - https://github.com/pubky/pubky-nexus/pull/889
- [x] Implement "index other HSs" logic
  - [x] Switch to using new HS SDK user events endpoint
  - [x] Store user cursors per HS (one idea: when user leaves HS, persist cursor in sorted set)
  - [x] #800
  - [x] Periodically skip processing of users for whom we get 404 (backoff approach, in-memory `HashMap`)
  - [x] Add support for HS PK blacklist (config list of HS PKs that should never be indexed)
    - https://github.com/pubky/pubky-nexus/pull/906
- [x] Implement "resolve user-to-HS mappings"
  - [x] https://github.com/pubky/pubky-nexus/pull/765
  - [x] Add backoff mechanism to periodically skip event ingestion for HSs found to be offline
 - ~Avoid duplicated events through event hash (e.g. when user migrates across HSs)~
   - migrating blocked
 - [x] https://github.com/pubky/pubky-nexus/pull/809
 - [ ] TBD: Move Universal Tag related utils to `pubky-app-specs`?
[...]

PR #884, opened 2026-05-27
feat: block switching from 3rd party HS

aintnostressin:
Disallow changing Homeserver for the third party homeserver's users that were already indexed by Nexus.

### DIFF
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -393,8 +393,8 @@ pub fn get_homeserver_by_id(id: &str) -> Query {
 pub fn get_all_homeservers_with_active_users() -> Query {
     Query::new(
         "get_all_homeservers_with_active_users",
-        "MATCH (u:User)-[:HOSTED_BY]->(hs:Homeserver)
-        WHERE u.name <> '[DELETED]'
+        "MATCH (u:User)-[r:HOSTED_BY]->(hs:Homeserver)
+        WHERE u.name <> '[DELETED]' AND NOT coalesce(r.stale, false)
         WITH hs.id AS id, count(u) AS active_users
         ORDER BY active_users DESC
         RETURN collect(id) AS homeservers_list",
@@ -418,12 +418,26 @@ pub fn get_users_needing_hs_resolution(ttl_ms: u64) -> Query {
     .param("ttl_ms", ttl_ms as i64)
 }
 
-/// Retrieves all user IDs hosted on a given homeserver.
+/// Retrieves the homeserver ID a user is currently hosted on, if any.
+pub fn get_user_homeserver(user_id: &str) -> Query {
+    Query::new(
+        "get_user_homeserver",
+        "MATCH (u:User {id: $user_id})-[:HOSTED_BY]->(hs:Homeserver)
+         RETURN hs.id AS homeserver_id",
+    )
+    .param("user_id", user_id.to_string())
+}
+
+/// Retrieves all user IDs actively hosted on a given homeserver.
+///
+/// Excludes users whose mapping is marked `stale` — i.e. whose published
+/// homeserver has diverged from the stored one — so the watcher stops
+/// indexing them until the mapping realigns.
 pub fn get_users_by_homeserver(hs_id: &str) -> Query {
     Query::new(
         "get_users_by_homeserver",
-        "MATCH (u:User)-[:HOSTED_BY]->(:Homeserver {id: $hs_id})
-         WHERE u.name <> '[DELETED]'
+        "MATCH (u:User)-[r:HOSTED_BY]->(:Homeserver {id: $hs_id})
+         WHERE u.name <> '[DELETED]' AND NOT coalesce(r.stale, false)
          RETURN collect(u.id) AS user_ids",
     )
     .param("hs_id", hs_id.to_string())
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -433,9 +433,9 @@ pub fn get_user_homeserver(user_id: &str) -> Query {
 /// Excludes users whose mapping is marked `stale` — i.e. whose published
 /// homeserver has diverged from the stored one — so the watcher stops
 /// indexing them until the mapping realigns.
-pub fn get_users_by_homeserver(hs_id: &str) -> Query {
+pub fn get_active_users_by_homeserver(hs_id: &str) -> Query {
     Query::new(
-        "get_users_by_homeserver",
+        "get_active_users_by_homeserver",
         "MATCH (u:User)-[r:HOSTED_BY]->(:Homeserver {id: $hs_id})
          WHERE u.name <> '[DELETED]' AND NOT coalesce(r.stale, false)
          RETURN collect(u.id) AS user_ids",
--- a/nexus-common/src/db/graph/queries/put.rs
+++ b/nexus-common/src/db/graph/queries/put.rs
@@ -392,3 +392,20 @@ pub fn set_user_homeserver(user_id: &str, homeserver_id: &str) -> Query {
     .param("user_id", user_id.to_string())
     .param("hs_id", homeserver_id.to_string())
 }
+
+/// Toggles the `stale` flag on a user's existing `HOSTED_BY` relationship and
+/// refreshes `resolved_at`.
+///
+/// Homeserver switching is not fully implemented, so the bound homeserver is
+/// never changed. Instead, when the published homeserver diverges (changed or
+/// removed) the mapping is marked stale to pause indexing; when it realigns the
+/// flag is cleared to resume.
+pub fn set_user_homeserver_stale(user_id: &str, stale: bool) -> Query {
+    Query::new(
+        "set_user_homeserver_stale",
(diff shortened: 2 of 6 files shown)


aintnostressin:
@greptile review again

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -418,12 +418,26 @@ pub fn get_users_needing_hs_resolution(ttl_ms: u64) -> Query {
...
+    )
+    .param("user_id", user_id.to_string())
+}
+
+/// Retrieves all user IDs actively hosted on a given homeserver.
+///
+/// Excludes users whose mapping is marked `stale` — i.e. whose published
+/// homeserver has diverged from the stored one — so the watcher stops
+/// indexing them until the mapping realigns.
 pub fn get_users_by_homeserver(hs_id: &str) -> Query {
```
ok300: Maybe a better method name is `get_active_users_by_homeserver`, to more clearly indicate the new check and align it with the existing `get_all_homeservers_with_active_users`.

nexus-watcher/src/service/user_hs_resolver.rs
```diff
@@ -18,22 +18,53 @@ use tracing::{debug, error, info, warn};
...
 static HS_RESOLVER_METRICS: LazyLock<HsResolverMetrics> = LazyLock::new(HsResolverMetrics::new);
 
+/// Resolves a user's currently published homeserver from PKDNS/DHT.
+///
+/// Abstracted behind a trait so the resolver loop can be driven with a mock in
+/// tests instead of hitting the network.
+#[async_trait::async_trait]
+pub trait PkdnsHomeserverResolver: Send + Sync {
+    /// Returns the homeserver published for `user_pk`, or `None` if none is
+    /// currently published.
```
ok300:
```suggestion
    /// Returns the HS published for `user_pk`, if any is currently published.
```

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -418,12 +418,26 @@ pub fn get_users_needing_hs_resolution(ttl_ms: u64) -> Query {
     .param("ttl_ms", ttl_ms as i64)
 }
 
-/// Retrieves all user IDs hosted on a given homeserver.
+/// Retrieves the homeserver ID a user is currently hosted on, if any.
+pub fn get_user_homeserver(user_id: &str) -> Query {
```
ok300, line 422: This query lacks `LIMIT 1`:

The existing `set_user_homeserver` query uses `OPTIONAL MATCH … DELETE old` to remove a prior `HOSTED_BY` edge before merging a new one.

If historical data or a concurrent write left a user with two edges, this query returns two rows. `fetch_key_from_graph` will likely error or pick one non-deterministically.

A `LIMIT 1` would make it safe and self-documenting.

nexus-common/src/db/graph/queries/put.rs
ok300, line 375: `set_user_homeserver` doesn't clear stale when called on an existing edge:

In the first-time registration path, `resolve_user` calls `set_user_homeserver`. The Cypher for that function does:
```
  MERGE (u)-[r:HOSTED_BY]->(hs)
  SET r.resolved_at = timestamp()
```

It does not set `r.stale = false`. If the `MERGE` matches an existing edge (e.g., leftover data or a race), the stale flag is preserved. In the PR's logic this can't happen because `get_user_homeserver` catches any existing edge first, but adding
```
SET r.stale = false, r.resolved_at = timestamp()
```
to this Cypher is cheap defensive hygiene.

nexus-watcher/src/service/user_hs_resolver.rs
```diff
@@ -376,6 +460,173 @@ mod tests {
         Ok(())
     }
 
+    #[tokio_shared_rt::test(shared)]
+    async fn test_get_user_homeserver() -> Result<(), DynError> {
+        setup().await?;
+
+        let user_id = "get_user_hs_test_user";
+        let hs_id = "get_user_hs_test_hs";
```
ok300:
```suggestion
        let user_id = Keypair::random().public_key().z32();
        let hs_id = Keypair::random().public_key().z32();
```

To be consistent with how IDs are generated in other tests.

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -393,8 +393,8 @@ pub fn get_homeserver_by_id(id: &str) -> Query {
 pub fn get_all_homeservers_with_active_users() -> Query {
```
ok300, line 393: Stale flag not tested via `get_all_homeservers_with_active_users`:

There is no test confirming that a homeserver with only stale users drops off the active list. The per-query logic is correct, but the gap means a regression there wouldn't be caught. Worth adding an assertion in one of the change/unpublish tests.

nexus-watcher/src/service/user_hs_resolver.rs
```diff
@@ -151,28 +186,56 @@ async fn get_users_needing_resolution(ttl_ms: u64) -> GraphResult<Vec<String>> {
...
-    exec_single_row(query).await?;
+    // Already bound to a homeserver: toggle the stale flag instead of switching.
+    let still_matches = resolved_hs_id
+        .as_ref()
+        .is_some_and(|hs_id| hs_id.as_ref() == stored_hs_id.as_str());
+    if still_matches {
+        exec_single_row(queries::put::set_user_homeserver_stale(&user_id, false)).await?;
+        debug!("User {user_id} still hosted on {stored_hs_id}, mapping active");
+    } else {
+        exec_single_row(queries::put::set_user_homeserver_stale(&user_id, true)).await?;
```
ok300, line 222: Stale mappings do not stop indexing from the default homeserver:

`nexus-watcher/src/service/user_hs_resolver.rs:223` marks a changed/unpublished mapping as stale, and the key-based external watcher respects that via `get_users_by_homeserver`. But the default watcher still processes every global event line from `/events/` without checking whether the event user has a non-stale `HOSTED_BY` edge for that homeserver: `nexus-watcher/src/service/indexer/homeserver.rs:115`.

If a user originally bound to the default HS later changes or unpublishes their HS, this PR will not “pause indexing” for default-HS events.

Add an active-mapping check before handling each default event, or document/test that the default homeserver is intentionally exempt.

---

IMO the default HS should not be exempt from this.

ok300 (commented):
https://github.com/pubky/pubky-nexus/blob/e3e9c8166f0ae9992133447bfcfe4fc1247c1b13/nexus-common/src/db/graph/queries/del.rs#L110

`remove_user_homeserver` in `del.rs` is now unreachable from `resolve_user`:

The previous implementation called `remove_user_homeserver` when DHT returned nothing. That call is gone, the new code marks the edge stale instead. The function still exists in `del.rs` but is now stale.

To keep things clean, it should be removed IMO.

aintnostressin:
> https://github.com/pubky/pubky-nexus/blob/e3e9c8166f0ae9992133447bfcfe4fc1247c1b13/nexus-common/src/db/graph/queries/del.rs#L110
> 
> `remove_user_homeserver` in `del.rs` is now unreachable from `resolve_user`:
> 
> The previous implementation called `remove_user_homeserver` when DHT returned nothing. That call is gone, the new code marks the edge stale instead. The function still exists in `del.rs` but is now stale.
> 
> To keep things clean, it should be removed IMO.

I left in on purpose because we will eventually need it again. Do you still think it's worth to remove it?

nexus-common/src/db/graph/queries/get.rs
aintnostressin: if there are two homeservers stored - then which one is correct ? I still prefer to error out in such case.

nexus-common/src/db/graph/queries/get.rs
aintnostressin: I added the same query condition to both functions. The naming inconsistency is pre existing.

aintnostressin: renamed [bac4edc](https://github.com/pubky/pubky-nexus/pull/884/commits/bac4edc3a15d57c0c405790f6ca5f1ad23ddeafc)

nexus-common/src/db/graph/queries/get.rs
aintnostressin: added: [740eb93](https://github.com/pubky/pubky-nexus/pull/884/commits/740eb9382f8e2e446d0907aa71b1d7e19f8439c0)

nexus-watcher/src/service/user_hs_resolver.rs
aintnostressin: That's true. This will be added alongisde with "Only process events of an user if that user's HS is explicitly set to the default HS (e.g. do not blindly ingest all events)" from the #726 checklist. I've updated title and description.

nexus-common/src/db/graph/queries/put.rs
aintnostressin: I prefer to avoid setting stale to `false` by default - this is what I would call as a defensive hygiene.
The `coalesce` function call on read query already handles null values.

ok300 (commented):
> I left in on purpose because we will eventually need it again. Do you still think it's worth to remove it?

Yes please, it's cleaner.

When we'll need it again, we might not remember this already exists and might end up just adding a new one. To prevent that, it's better to remove it now. It's also pretty small, not a lot of logic is lost.

ok300 (approved):

PR #887, opened 2026-05-28
Fix swagger `reach` dropdown in `/v0/tags/hot` endpoint

ok300:
The hot tags endpoint (`/v0/tags/hot`) , when called with `user_id` and `reach` , fails with:

> Failed to deserialize query string: reach: unknown variant Followers, expected one of followers, following, friends, wot, wot_1, wot_2, wot_3

The GET request was triggered from the Swagger UI available at `/swagger-ui`  then choosing `Followers` from the reach dropdown.

---

This PR fixes the above error, which was caused by what Swagger UI considers as valid values that it proposes in the dropdown (in this case `Followers` capitalized, when the error msg mentioned `followers` lowercase).

This PR also
- replaces the custom deserializer for `StreamReach` with `serde` annotations, while preserving logic
- updates / fixes the `reach` description in the API interface (had typo + missing `wot` variants)


SHAcollision (approved):
LGTM

PR #889, opened 2026-05-28
feat: check user homeserver before processing default hs event

aintnostressin:
Gates the default-homeserver indexer so it only processes events for users that still resolve to this homeserver:
- Graph path (cheap): read the user's HOSTED_BY mapping — process only if it points to this HS and is not stale; refuse if it diverges (points elsewhere, or here but stale).
- PKDNS fallback: if there's no mapping yet (new/unresolved user), resolve the homeserver live from PKDNS and process only if it matches.

Currently there is no way to distinct no pkds record from resolve failure - we may need https://github.com/pubky/pubky-core/issues/408 to add proper error handling instead of skipping user events.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -33,4 +33,5 @@ jobs:
           cache-on-failure: true
 
       - name: Run clippy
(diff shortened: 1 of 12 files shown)


aintnostressin:
closing this one in favour of #902

PR #890, opened 2026-05-28
Introduce, handle `HsEventsStreamRateLimitExhausted` event processor error

ok300:
When a `KeyBasedEventProcessor` tries to fetch events of a user, but repeatedly gets 429 error, to the point that all backoff retry attempts are exhausted, this results in a `PubkyClientError::TooManyRequests429` error.

This PR marks this error type as infrastructure error, to prevent the `KeyBasedEventProcessor` from attempting to fetch events for the next user, if any.


aintnostressin (approved):
lgtm

the error classification will probably undergo rework

aintnostressin (approved):

PR #891, opened 2026-05-29
Add user backoff when `/event-stream` returns 404 error (user not found)

ok300:
This PR expands the `KeyBasedEventSource` used when processing 3rd party HSs, such that receiving error 404 from the HS when trying to fetch a user's events results in a backoff delay causing that user to be skipped periodically, without affecting other users or HSs.


ok300:
@greptile review again please

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -119,26 +203,53 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
-                        error = ?err,
-                        "Infrastructure error while processing user; aborting homeserver run",
-                    );
-                    return Err(err);
-                }
-
-                error!(
+            // Users whose event fetch previously returned 404 are skipped for an
+            // increasing number of runs (see `UserNotFoundBackoff`).
+            if self.event_source.should_skip_user(user_pk) {
```
aintnostressin: Is it possible to move the responsibility of keeping backoff state to processor instead of keeping it in event_source ?

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -25,9 +81,25 @@ pub trait KeyBasedEventSource: Send + Sync + 'static {
         cursor: EventCursor,
         limit: u16,
     ) -> Result<Vec<StreamEvent>, EventProcessorError>;
+
+    /// Returns `true` if the user should be skipped this run due to a prior
```
aintnostressin: Why does `KeyBasedEventSource` starts to know anything about outer logic like "runs" ? It seems like this responsibility should  not belong here. Can responsibility of tracking it be moved to `KeyBasedEventProcessor` ?

nexus-watcher/src/service/indexer/key_based.rs
ok300: Done: f1a851a1ff940ef029b020377c535363b3110b30

nexus-watcher/src/service/indexer/key_based.rs
ok300: Done, moved out of `KeyBasedEventSource`: f1a851a1ff940ef029b020377c535363b3110b30

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -71,6 +72,9 @@ pub struct KeyBasedEventProcessor {
     pub files_path: PathBuf,
     pub event_handler: Arc<dyn EventHandler>,
     pub event_source: Arc<dyn KeyBasedEventSource>,
+    /// Per-user 404 backoff, shared with (and owned by) the runner so its state
```
aintnostressin: I think this code should not make any claims how it's used by other components.

I suggest removing this comment.

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -119,26 +123,53 @@ impl TEventProcessor for KeyBasedEventProcessor {
...
                     action = "skip_user",
-                    error = ?err,
-                    "Got error while processing user: continuing with next user",
+                    "Skipping user due to prior 404 (NotFound404) backoff",
                 );
+                continue;
+            }
+
+            match self.process_user(&hs_pk, &hs_id, user_pk, *cursor).await {
+                // Successful fetch clears any pending 404 backoff for the user.
```
aintnostressin: Ok => clear 

The code is self documenting.
It's worth removing this comment.

nexus-watcher/src/service/runner/key_based_user_backoff.rs
```diff
@@ -0,0 +1,58 @@
...
+/// subsequent runs that user should be skipped: the 1st 404 skips the user once,
+/// the 2nd skips twice, and so on, capped at `MAX_USER_NOT_FOUND_SKIPS`.
+/// A successful fetch clears the user's state.
+#[derive(Default)]
+pub struct UserNotFoundBackoff {
+    inner: Mutex<HashMap<PublicKey, BackoffEntry>>,
+}
+
+#[derive(Default)]
+struct BackoffEntry {
```
aintnostressin, line 21: Can we have some unit tests here?

nexus-watcher/tests/service/utils/key_based_event_source.rs
```diff
@@ -18,6 +18,10 @@ pub struct MockKeyBasedEventSource {
     /// Useful when graph user ordering is intentionally not part of the assertion.
     user_events: Mutex<HashMap<String, FetchEventsResult>>,
 
+    /// Error returned on every fetch, without being consumed.
+    /// Useful for simulating a persistently failing user (e.g. repeated 404s) across runs.
+    sticky_error: Mutex<Option<EventProcessorError>>,
```
aintnostressin: can `events` / `user_events` be reused instead ? just simply feed it with more results ? This would help keeping this mock simple and also control how many times it's called.

nexus-watcher/src/service/indexer/key_based.rs
ok300: Done: 7dad5f035ce6e1197d645a2e2ee674951f379620

nexus-watcher/src/service/indexer/key_based.rs
ok300: Done: 7dad5f035ce6e1197d645a2e2ee674951f379620

nexus-watcher/src/service/runner/key_based_user_backoff.rs
ok300: Done: 107755dfc0e14d97e2ecf1592a1ec8aca580fb8b

nexus-watcher/tests/service/utils/key_based_event_source.rs
ok300: Done: 12864b47b53a25c763fe231f98af8b5916e1a819

nexus-watcher/src/service/runner/key_based_user_backoff.rs
```diff
@@ -0,0 +1,116 @@
+use std::collections::HashMap;
+
+use pubky::PublicKey;
+use tokio::sync::Mutex;
+
+/// Upper bound on how many consecutive runs a user can be skipped after repeated 404s.
+const MAX_USER_NOT_FOUND_SKIPS: u32 = 10;
```
aintnostressin: This should be renamed to `MAX_FAILURES`

nexus-watcher/src/service/runner/key_based_user_backoff.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// Per-user backoff state for users whose event fetch returns HTTP 404.
+///
+/// Tracks consecutive 404s per user (keyed by public key) and derives how many
```
aintnostressin, line 11: We really don't know here what it tracks - it depends how it is used by the caller.
I would be good to make comments more general, like you did with methods, `record_not_found` -> `record_failure`

ok300:
I would keep the references to 404 (naming, comments) as this is the only use-case of this struct. For now, the places where we use backoff are so different from each other, that I don't see a good way it can be re-used. If we ever see a place where we can re-use it, IMO that's the time to generalize it.

aintnostressin (approved):

nexus-watcher/src/service/runner/key_based_user_backoff.rs
ok300: Done: 3a60f9487c0bcbcbb89a12bdbb64dd0f730bad38

PR #893, opened 2026-05-29
feat: add docker image pipeline

SpontaneousOverthrow:


.github/workflows/docker.yml
```diff
@@ -0,0 +1,40 @@
+name: Build and Publish Docker Image
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
```
SHAcollision, line 6: NIT: the workflow pushes `latest` for every tag matching `v*.*.*`. That's ok for simple release workflows, but it has a risk: rebuilding or pushing an older version tag later could move `latest` backwards unless tag creation is tightly controlled. Not worth fixing IMO unless you know of a really quick/easy fix.

SHAcollision (approved):
Tests failing. Once, maybe something flaky? Just relaunched them. LGTM otherwise, left a single NIT.

.github/workflows/docker.yml
SpontaneousOverthrow: that's ok for now, in the future we will fix that

PR #894, opened 2026-05-30
Simplify Redis ops (read/write) for user cursor handling

ok300:
This PR relocated the per-user cursor Redis plumbing from the indexer into the model layer that owns the key schema.


ok300:
@greptile please review again

aintnostressin (approved):

aintnostressin (approved):
Approved, but to me seems like user's cursors doesn't belong to details. Maybe we should find a separate place for it.

PR #898, opened 2026-06-02
feat(api): web of trust post streams, domain-trust, and post-tag filtering

### LINKED ISSUE #897: Graph-backed post streams ignore `order` and use a timestamp cursor for `total_engagement`
The graph query path (`post_stream` / `get_from_graph`), used by all non-index post streams (`following`/`followers`/`friends`), has two pre-existing bugs:

- **`order` ignored:** `post_stream` hardcodes `ORDER BY ... DESC` for both sorts; `get_from_graph` never receives `SortOrder`, so `order=ascending` is silently dropped.
- **Wrong engagement cursor:** `get_from_graph` always derives `last_post_score` from `indexed_at`, even for `sorting=total_engagement`, so `start`/`end` cursor pagination by engagement skips/repeats.

**User impact while unfixed:** a feed served by the graph path (e.g. `following` with `sorting=total_engagement`) can't be paged past the first page (the next request returns the same posts), and `order=ascending` silently returns descending results.

**Fix:** thread `SortOrder` into `post_stream`/`get_from_graph` (map to `ASC`/`DESC`, like the resource stream already does) and `RETURN` a sorting-matched `score` column (`indexed_at` for timeline, `total_engagement` for engagement) that `get_from_graph` reads into `last_post_score`.

**Files:** `nexus-common/src/db/graph/queries/get.rs` (`post_stream`), `nexus-common/src/models/post/stream.rs` (`get_from_graph`).

SHAcollision:
## What

Filters feeds and post tags through a user's Web of Trust instead of the global view.

- **`source=wot`** - posts from authors in the observer's follow graph (transitive `FOLLOWS`, depth 1-3, default 2; depth 3 gets expensive without a cache). Accounts nobody in the network follows are excluded by graph topology, no ranking involved.
- **`source=wot_domain`** - posts from authors the observer's trust network has tagged with a `domain_tags` label (e.g. `bitcoiner`). Reuses the existing `(:User)-[:TAGGED]->(:User)` edges, so no migration or schema change.
- **WoT post tags** - `GET /v0/post/{author}/{post}/tags?viewer_id=&depth=` returns tags only from taggers in the viewer's WoT, so a moderator the viewer follows isn't buried under hundreds of other tags on a popular post.
- **Uniform depth validation** - a `depth` outside 1-3 supplied with a viewer now returns 400 on every WoT endpoint, instead of silently falling back to the global view.

Graph-only for now; no new cache (can add one later if latency calls for it). Metrics counters deferred.

## API

```
GET /v0/stream/posts?source=wot&observer_id={pk}&depth=2
GET /v0/stream/posts?source=wot_domain&observer_id={pk}&depth=2&domain_tags=bitcoiner,btc-dev
GET /v0/post/{author}/{post}/tags?viewer_id={pk}&depth=2
```
`/v0/stream/posts/keys` takes the same stream params.

## Backwards compatibility

- Requests without WoT params behave exactly as before.
- The new `depth`/`domain_tags` params now return 400 on malformed input instead of ignoring it, consistent with the existing `tags` param.
- Tag endpoints: WoT activates only with `viewer_id` plus an explicit `depth`. `viewer_id` alone stays the global view, so existing callers aren't silently switched to a filtered subset.

## Also fixes #897

Graph-backed streams were ignoring `order=ascending` and returning a timestamp as the `total_engagement` cursor, so the next page just repeated the first. Both fixed, for all graph sources, not only the WoT ones.


nexus-webapi/src/routes/v0/stream/posts.rs
```diff
@@ -101,22 +114,40 @@ fn build_stream_source(
...
-                return Err(Error::invalid_input(
-                    "source 'collection' requires 'post_id' parameter",
-                ));
-            }
-            Ok(StreamSource::Collection {
-                author_id: author_id.unwrap().to_string(),
-                post_id: post_id.unwrap().to_string(),
-            })
-        }
+        StreamSourceKind::Collection => match (author_id, post_id) {
```
SHAcollision, line 117: boyscoutting to ensure complete match on all  `author_id` and `post_id` combinations

nexus-common/src/types/mod.rs
```diff
@@ -20,12 +20,67 @@ pub enum StreamSorting {
     TotalEngagement,
 }
 
+/// Web of Trust traversal depth, validated to `1..=3` at construction. The
+/// `FOLLOWS*1..n` graph traversals are expensive, so the query builders only
+/// accept this type, keeping the bound enforced for every caller (web, tests,
+/// benches, internal) rather than at the web layer alone.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, ToSchema)]
+pub struct WotDepth(u8);
```
SHAcollision, line 28: Creating a new type `WotDepth` and using it consistently on all already existing "wot related" features as a mean to cleanup. This PR ends up touching more files and lines than strictly necessary because it revamps lot of existing code.

nexus-webapi/src/models/crockford_id.rs
```diff
@@ -0,0 +1,107 @@
...
+use crate::Error;
+use pubky_app_specs::validate_crockford_id;
+use serde::de;
+use serde::Deserialize;
+use utoipa::ToSchema;
+
+/// Defines a 13-character Crockford Base32 ID newtype. Post and file IDs are
+/// structurally identical (same encoding, length and validation), so they share
+/// one definition instead of duplicating the impls and tests.
+macro_rules! crockford_id {
```
SHAcollision, line 12: boyscoutting

SHAcollision:
Opening here for review. This PR touches WoT adjacent code for clean-up/refactor/clarification. When comanding your AI to review this PR, remember to feed it the specs as context https://hackmd.io/cm_1fxHQSPuOW1lXUCVj_Q to avoid asks that are outside of scope (e.g., caching).

nexus-common/src/models/post/stream.rs
```diff
@@ -177,14 +198,35 @@ impl PostStream {
...
+        if let Some((source, depth)) = wot {
+            super::metrics::record_wot_result(
+                source,
+                depth,
+                started.elapsed(),
+                post_keys.post_keys.len(),
+            );
+        }
+
         Ok(post_keys)
```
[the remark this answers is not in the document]
SHAcollision: Fixed. `query_duration` is now recorded on the error path too (plus a new `wot.post_stream.errors` counter), so timeouts and DB errors no longer drop out of the histogram and inflate p95.

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -408,46 +409,28 @@ pub fn get_all_homeservers() -> Query {
...
-            MATCH (viewer)-[:FOLLOWS*1..{depth}]->(tagger:User)
-            RETURN DISTINCT tagger
-        }}
-        MATCH (tagger)-[tag:TAGGED]->(tagged)
-        WITH tag.label AS label, collect(tagger.id) AS taggerIds
-        RETURN 
-            taggerIds IS NOT NULL AS exists,
-            collect({{
+            WITH viewer, tagged
+            MATCH (viewer)-[:FOLLOWS*1..{depth}]->(tagger:User)-[tag:TAGGED]->(tagged)
```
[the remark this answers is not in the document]
SHAcollision: Done, added a comment at the `OPTIONAL MATCH` documenting the intentional NULL-viewer → `[]` behaviour.

SHAcollision:
On the user-view depth-leniency flagged in the Greptile summary: that is intentional and documented in the WoT spec (the profile endpoint treats `depth` as best-effort and falls back to global for out-of-range values, consistent across its single and batch paths, rather than 400-ing like the dedicated tag endpoints).

SHAcollision:
@greptile review again

nexus-common/src/models/post/metrics.rs
```diff
@@ -0,0 +1,91 @@
...
+                .build(),
+        }
+    }
+}
+
+static METRICS: LazyLock<WotStreamMetrics> = LazyLock::new(WotStreamMetrics::new);
+
+/// Count one WoT post-stream request. `source` is `"wot"` or `"wot_domain"`.
+pub(super) fn record_wot_request(source: &'static str) {
+    METRICS.requests.add(1, &[KeyValue::new("source", source)]);
```
tipogi: Should this request counter also include `depth`?
`empty`, `query_duration`, and `returned_posts` are recorded with `source + depth`, and the module docs describe `wot_feed_empty_rate{depth}`.
Without `depth` on `requests`, the empty-rate denominator is all requests for the source, not requests for that specific depth, so per-depth dashboards can report misleading rates.

nexus-common/src/models/tag/traits/collection.rs
```diff
@@ -185,35 +208,25 @@ where
...
     async fn get_from_graph(
         user_id: &str,
         extra_param: Option<&str>,
-        depth: Option<u8>,
+        depth: Option<WotDepth>,
     ) -> GraphResult<Option<Vec<TagDetails>>> {
         // We cannot use LIMIT clause because we need all data related
         let query = match depth {
-            Some(distance) => queries::get::get_viewer_trusted_network_tags(
+            Some(depth) => queries::get::get_viewer_trusted_network_tags(
```
tipogi, line 221: This WoT branch is user-specific, but the default `TagCollection` implementation is also used by `TagResource`.

`resource_tags_handler` accepts `depth` through `TagsQuery` and passes it into `TagResource::get_by_id(...)`. That reaches this default WoT branch, which calls `get_viewer_trusted_network_tags()`.

The problem is that the graph query matches a `User` node and not `Resource` node. 
```cypher
MATCH (tagged:User {id: $user_id})
```

For resource tag requests, `$user_id` is actually the resource id, so an existing resource can fail the graph lookup and be reported as not found.

What about if we reject `depth` on resource endpoints

nexus-common/src/models/user/view.rs
```diff
@@ -42,9 +43,17 @@ impl UserView {
...
-                .await?
-                .unwrap_or_default(),
+            _ => TagUser::get_by_id(
+                user_id,
+                None,
+                None,
+                None,
+                None,
+                viewer_id,
+                depth.and_then(|d| WotDepth::new(d).ok()),
```
tipogi, line 53: This currently converts invalid depths into `None`. So _GET_ `/v0/user/{user_id}?viewer_id=...&depth=99` returns `200` with global tags instead of rejecting the malformed WoT request.

The dedicated tags endpoint validates via `resolve_tag_wot_depth(...)` and returns `400` for invalid depth. Profile views should probably use the same validation behavior, likely at the route boundary, before calling `UserView::get_by_id(...)`.

nexus-webapi/src/models/crockford_id.rs
tipogi, line 1: NIT: this folder currently keeps both `response`/`domain-ish` models and validated API input value types under `nexus-webapi/src/models`.

`PostId` / `FileId` fit the same category as `ResourceId`, `TagLabel`, `GlobalPostId`, the prefix types, and the bounded vec helpers: they are `request`/`value` types used at the API boundary.

If we want a clearer boundary, these could live together in a `types` module. If you agree, we could open an issue to do the refactor

nexus-webapi/src/routes/v0/stream/posts.rs
tipogi, line 204: `tags` is accepted for reply sources here, but `source=post_replies` / `source=author_replies` with `tags` currently returns an empty stream because the model layer does not fall back to a reply-aware graph query.

Tracked separately in #927, but worth noting this route currently allows that unsupported combination through.

nexus-common/src/models/post/metrics.rs
SHAcollision: Fixed in 9a6c5893 , `requests` now carries `depth` too, so the empty-rate denominator (`empty / requests`) lines up with the per-depth `empty` count.

nexus-common/src/models/tag/traits/collection.rs
SHAcollision: Fixed in 433f2dac , resource tag/tagger endpoints now reject `depth` with 400 instead of 404-ing the resource. Resource tags have no WoT variant (the query matches a `User` node), so rejecting the unsupported combo beats silently failing the lookup.

nexus-common/src/models/user/view.rs
SHAcollision: Fixed in 59eabb23 , the profile route now validates via `resolve_tag_wot_depth` at the boundary, so an out-of-range depth (or `depth` without `viewer_id`) returns 400, consistent with the tag endpoints.

nexus-webapi/src/models/crockford_id.rs
SHAcollision: Agreed! PostId/FileId are the same value-type category as ResourceId/TagLabel/etc. Out of scope for this PR (pure reshuffle on an already-large diff). Let's open an issue for it?

nexus-webapi/src/routes/v0/stream/posts.rs
SHAcollision: Agreed. Leaving this to #927, that's where reply sources get the tag-aware graph query, and a 400 guard here would just be ripped out once they do.

SHAcollision:
@greptile review again

tipogi (approved):
ready! 🚀

SHAcollision:
Merging here, ready to fix stuff from Prod :rocket: :laughing:

PR #900, opened 2026-06-02
refactor: extract `EventLine`, move event related stuff to `nexus-watcher`

aintnostressin:
The Event struct in `nexus-common` was doing double duty: it held both the parsed event data (`URI`, `type`, `HomeserverParsedUri`) and the raw event line string used for Redis storage. The Redis write worked via a non-obvious `AsRef<[String]>` impl that returned `slice::from_ref(&self.event_line)` - a single-element slice pointing at the raw string, which `put_index_list` then iterated over to write to Redis.

This splits the two concerns cleanly:

- `RawEvent(pub String)` stays in `nexus-common` and owns the Redis key. It keeps the `AsRef<[String]>` trick, which is now the obvious and only thing it does.
- `Event` moves to `nexus-watcher` and holds only parsed data. Storing an event is now an explicit two-step: `event.to_raw().store()`.

Also pins `RawEvent::prefix()` to "Event" so the Redis key (`Event:Events`) stays stable across future renames. The RedisOps default derives the prefix from the struct name, which would have silently moved the key to `Raw:Event:Events` without the override.


ok300:
The title is a bit misleading: this is not simplifying `Event`. Things are touched, which are not `Event`, like `ParseResult`, etc.

Looks like the goal is to separate Redis-specific (`RetryManager`-specific) logic out of `Event` and into `RawEvent`? If so, would it have been simpler to just extract that (if the goal is, as per title, to simplify `Event`)?

It also looks like this moved `Event` (and other things) around, so it's hard to follow exactly what was done. Would have been simpler to separate in different commits
- 1st commit, extract `RawEvent`
- 2nd commit, move `Event` to better location
- etc

Will review later in more detail, but these are some first thoughts.

nexus-watcher/src/events/event.rs
```diff
@@ -0,0 +1,151 @@
...
+        Ok(ParseResult::Parsed(Event {
+            uri,
+            event_type,
+            parsed_uri,
+            files_path,
+            event_line,
+        }))
+    }
+
+    pub fn to_raw(&self) -> RawEvent {
```
ok300: Why not derive it idiomatically via `From`?

```rust
impl From<&Event> for RawEvent {
  fn from(e: &Event) -> Self {
    RawEvent(e.event_line.clone())
  }
}
```

Then storing becomes

```rust
RawEvent::from(&event).store().await?;
```

nexus-watcher/src/events/event.rs
ok300, line 1: Looking at how the other things in `nexus-common/src/models/` it seems the convention is to keep models there and create new files for related structs, for example

``` rust
// in nexus-common/src/models/post/details.rs

pub struct PostDetails { ... }

impl RedisOps for PostDetails { ... }
```

Why not keep the `Event` where it was, but just add a new file next to it `raw.rs` with `RawEvent`?

This would also allow you to do

```rust
pub struct RawEvent(pub(crate) String);
```

which would limit the ability to create and insert new `RawEvents` into Redis to this crate.

nexus-watcher/src/events/event.rs
```diff
@@ -0,0 +1,151 @@
...
+            EventType::Put => "PUT",
+            EventType::Del => "DEL",
+        };
+        write!(f, "{upper_case_str}")
+    }
+}
+
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug)]
+pub enum ParseResult {
```
ok300, line 38: This was moved, but the rustdocs of its fields were lost.

Moving `Event` and its utils back to the old location would automatically address this.

(A)

nexus-watcher/src/events/event.rs
```diff
@@ -0,0 +1,151 @@
...
+        Self::UnrecognizedUri {
+            event_type,
+            uri,
+            reason,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Event {
```
ok300, line 64: Similar to (A), the struct was moved but field comments were lost / left behind.

ok300 (changes requested):

nexus-watcher/src/events/event.rs
aintnostressin: These types are used only by watcher, why should we keep them in common ?

ok300: Because moving them is invasive and doesn't achieve much:
- there is a `models` is `nexus-common`, but none in `nexus-watcher`
  - so moving _some_ models adds confusion
- `Event` and `RawEvent` are basically the same model entity, jus two ways to represent it
  - storing one in one place, the other somewhere else is also a bit unusual
- moving requires `RawEvent(pub String)` which opens up the ability for any caller to create and inject events into Redis, whereas before it wasn't possible and keeping it in `common` can reduce visibility there to `RawEvent(pub(crate) String)`

Overall
- many small downsides
- one small theoretical upside ("its only used by the watcher")

Have no strong preference for this, if you think its a good idea, let's move them. To me it seems better overall to keep them in `common/models`.

nexus-watcher/src/events/event.rs
tipogi: the convention of `common` should be: It is shared/used by both crates, `watcher` and `api`

nexus-watcher/src/events/event.rs
aintnostressin: restored [8c6b8aa](https://github.com/pubky/pubky-nexus/pull/900/commits/8c6b8aac34f00e2986c424d1708da32b08671c3a)

nexus-watcher/src/events/event.rs
aintnostressin: comments restored [8c6b8aa](https://github.com/pubky/pubky-nexus/pull/900/commits/8c6b8aac34f00e2986c424d1708da32b08671c3a)

nexus-watcher/src/events/event.rs
aintnostressin: Two things to note:

1. `Event` is in `nexus-watcher` and we do not want to introduce dependency `nexus-common` -> `nexus-watcher`
2. https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from - in this case conversion is not lossless so it breaks one of the described rules

nexus-common/src/models/event/mod.rs
tipogi: to follow the convention of the models, it should renamed to `get_from_index`

tipogi (changes requested):
LGTM! ready to merge when the name is changed 🚀

nexus-common/src/models/event/mod.rs
aintnostressin: updated [ba573f2](https://github.com/pubky/pubky-nexus/pull/900/commits/ba573f2330bc1d2a8aaaa11b4128bb37bad3b775)

tipogi (approved):
Merge it! gogo 🔥

ok300 (approved):

PR #901, opened 2026-06-03
fix: add `HOSTED_BY` guard for `HsEventProcessor`

ok300:
This PR is a proposed alternative to #889.

Changes:
- in the default event processor `HsEventProcessor`:
  - Check each parsed event's user `HOSTED_BY` before processing
  - When `HOSTED_BY` is missing, enqueue those events for retry
  - When `HOSTED_BY` exists but points to a different HS, skip processing
- Add tests for above behavior
- Updated moderation tests, such that they use different random moderator IDs
  - Before, they were using the same moderator ID with different random HSs, so with the `HOSTED_BY` addition it looked as if the same moderator was hosted on multiple HSs at once
- Update user delete safety check `user_is_safe_to_delete`, so presence of `HOSTED_BY` does not prevent soft-deleting users


ok300:
One risk I see is, if the PKDNS resolution (and therefore the creation of `HOSTED_BY`) gets delayed long enough, and the retrying happens frequently enough, then the retry budget might run out. This would cause those events to never be indexed.

PR #902, opened 2026-06-04
`HsEventProcessor`: validate HS mapping with graceful fallback

ok300:
This PR is a proposed alternative to #889.

Changes:
- in the default event processor `HsEventProcessor`:
  - Check each parsed event's user `HOSTED_BY` before processing
  - When `HOSTED_BY` exists but points to a different HS, skip processing
  - When `HOSTED_BY` is missing, or is pointing to the current HS, continue processing event (graceful fallback)
- Add tests for above behavior

This is relevant for two time periods:
- first, when DX will be introduced, between the time of startup and until all existing users have their PKDNS records resolved: all existing users will belong to the default HS, but have no `HOSTED_BY` edge yet.
- second, between the time when a new user registers on the default HS and when `UserHsResolverRunner` resolves its PKDNS record: new user(s) map to the default HS and generate events there, but we have not populated their `HOSTED_BY` edge yet.

In both these cases:
- validation treats `HOSTED_BY` as a strong signal: if it points to a different HS, it means the user very recently changed HSs => skip processing
- in the absence of `HOSTED_BY`, validation gracefully falls back on other available signals: the presence of this user's events in the `/events` lines indicates this user points to the default HS => process such events

### DIFF
--- a/nexus-watcher/src/service/indexer/homeserver.rs
+++ b/nexus-watcher/src/service/indexer/homeserver.rs
@@ -6,9 +6,11 @@ use nexus_common::models::event::{Event, EventProcessorError};
 use nexus_common::models::homeserver::Homeserver;
 use pubky::Method;
 use pubky_app_specs::PubkyId;
+use std::collections::HashMap;
 use std::path::PathBuf;
 use std::sync::Arc;
 use tokio::sync::watch::Receiver;
+use tokio::sync::Mutex;
 use tracing::{debug, error, info, warn};
 
 /// A user's `HOSTED_BY` binding, classified relative to a processor's homeserver.
@@ -16,7 +18,8 @@ use tracing::{debug, error, info, warn};
 /// The `stale` flag is only carried where it is meaningful: a stale mapping means
 /// the user's published homeserver has diverged from the stored one (see
 /// [`set_user_homeserver_stale`](nexus_common::db::queries::put::set_user_homeserver_stale)).
-enum HsBinding {
+#[derive(Clone)]
+pub enum HsBinding {
     /// The user has no `HOSTED_BY` edge yet.
     Unbound,
     /// The user is bound to this processor's homeserver.
@@ -37,6 +40,10 @@ pub struct HsEventProcessor {
     pub shutdown_rx: Receiver<bool>,
     /// Scheduler used to enqueue failed events onto the retry queue
     pub retry_scheduler: Arc<RetryScheduler>,
+
+    /// Per-run cache of users' `HOSTED_BY` mappings. For a given user's events in
+    /// the events list, only the 1st one results in a graph lookup, the rest read from this cache.
+    pub hs_mapping_cache: Mutex<HashMap<String, HsBinding>>,
 }
 
 #[async_trait::async_trait]
@@ -74,7 +81,7 @@ impl TEventProcessor for HsEventProcessor {
     async fn should_process_event(&self, event: &Event) -> bool {
         let user_id = event.parsed_uri.user_id();
 
-        match self.user_hs_binding(user_id).await {
+        match self.user_hs_mapping(user_id).await {
             // No mapping yet (graceful fallback) or actively bound here: process.
             Ok(HsBinding::Unbound) | Ok(HsBinding::Current { stale: false }) => true,
 
@@ -132,25 +139,33 @@ impl TEventProcessor for HsEventProcessor {
 }
 
 impl HsEventProcessor {
-    /// Resolves a user's `HOSTED_BY` binding relative to this processor's HS.
-    ///
-    /// Performs a single graph lookup and classifies the result, attaching the
-    /// `stale` flag where it is meaningful (see [`HsBinding`]).
-    async fn user_hs_binding(&self, user_id: &PubkyId) -> GraphResult<HsBinding> {
-        let query = queries::get::get_user_homeserver(user_id.as_ref());
+    /// Resolves and caches a user's `HOSTED_BY` binding relative to this processor's HS
+    async fn user_hs_mapping(&self, user_id: &PubkyId) -> GraphResult<HsBinding> {
+        if let Some(binding) = self.hs_mapping_cache.lock().await.get(user_id.as_ref()) {
+            return Ok(binding.clone());
+        }
 
-        let Some(row) = fetch_row_from_graph(query).await? else {
-            return Ok(HsBinding::Unbound);
+        let query = queries::get::get_user_homeserver(user_id.as_ref());
+        let mapping = match fetch_row_from_graph(query).await? {
+            None => HsBinding::Unbound,
+            Some(row) => {
+                let hs_id: String = row.get("homeserver_id")?;
+                let stale: bool = row.get("stale")?;
+
+                if hs_id.as_str() == self.homeserver.id.as_ref() {
+                    HsBinding::Current { stale }
+                } else {
+                    HsBinding::Other { hs_id }
+                }
+            }
         };
 
-        let hs_id: String = row.get("homeserver_id")?;
-        let stale: bool = row.get("stale")?;
+        self.hs_mapping_cache
+            .lock()
+            .await
+            .insert(user_id.as_ref().to_string(), mapping.clone());
 
-        Ok(if hs_id.as_str() == self.homeserver.id.as_ref() {
-            HsBinding::Current { stale }
-        } else {
-            HsBinding::Other { hs_id }
-        })
+        Ok(mapping)
     }
 
     /// Polls new events from the homeserver.
--- a/nexus-watcher/src/service/indexer/homeserver.rs
+++ b/nexus-watcher/src/service/indexer/homeserver.rs
@@ -79,29 +79,26 @@ impl TEventProcessor for HsEventProcessor {
     /// - A stale edge to this homeserver (the user's published homeserver has
     ///   diverged): log a warning and skip until the resolver realigns it.
     /// - An edge to a different homeserver: log a warning and skip.
-    ///
-    /// A lookup failure fails open (the event is processed), so a transient
-    /// graph error never silently drops legitimate events.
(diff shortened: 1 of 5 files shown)


nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -46,6 +47,40 @@ impl TEventProcessor for HsEventProcessor {
         Some(self.homeserver.id.as_ref())
     }
 
+    /// Skips events from users that are bound to a *different* homeserver.
```
aintnostressin: This comment misleads the reader. No edge doesn't necessary mean user is bound to a different homeserver.

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -65,6 +100,22 @@ impl TEventProcessor for HsEventProcessor {
 }
 
 impl HsEventProcessor {
+    /// Returns the homeserver a user is bound to via `HOSTED_BY`, but only when
+    /// it differs from this processor's homeserver.
+    ///
+    /// Returns `Ok(None)` when the user has no `HOSTED_BY` edge or the edge
+    /// points at this processor's homeserver, and `Ok(Some(hs_id))` when the
+    /// edge points at a different homeserver.
+    async fn user_bound_to_other_homeserver(
```
aintnostressin: it seems like this should answer true / false but instead returns `Option<String>`.

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -46,6 +47,40 @@ impl TEventProcessor for HsEventProcessor {
...
+        let user_id = event.parsed_uri.user_id();
+
+        match self.user_bound_to_other_homeserver(user_id).await {
+            Ok(Some(other_hs_id)) => {
+                warn!(
+                    event.uri = %event.uri,
+                    user_id = %user_id,
+                    processor_homeserver = %self.homeserver.id,
+                    user_homeserver = %other_hs_id,
+                    "User is hosted on a different homeserver; skipping event"
```
aintnostressin, line 107: I think we don't know that user is hosted on a different homserver, do we?

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -46,6 +47,40 @@ impl TEventProcessor for HsEventProcessor {
...
+    /// Before an event is processed we inspect the user's `HOSTED_BY` edge:
+    /// - No edge, or the edge points at this processor's homeserver: process.
+    /// - The edge points at a different homeserver: log a warning and skip.
+    ///
+    /// A lookup failure fails open (the event is processed), so a transient
+    /// graph error never silently drops legitimate events.
+    async fn should_process_event(&self, event: &Event) -> bool {
+        let user_id = event.parsed_uri.user_id();
+
+        match self.user_bound_to_other_homeserver(user_id).await {
```
ok300: Proposal from @tipogi :

We could / should have a memory structure (`HashMap`?) where user-HS mappings can be cached. This can be a local variable, which could be dropped after each even processor loop run. Goal is to minimize amount of graph queries.

ok300: Added this type of caching: 04b27855d6e17783effebb82a5d654bd7d2491e3

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -205,6 +215,10 @@ pub trait TEventProcessor: Send + Sync + 'static {
         )
     )]
     async fn handle_event(&self, event: &Event) -> Result<(), EventProcessorError> {
+        if !self.should_process_event(event).await {
+            return Ok(());
+        }
+
         let span = tracing::Span::current();
         if let Err(e) = self.event_handler().handle(event).await {
```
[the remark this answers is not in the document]
ok300: Fixed: 8c361a7da2685c324a8763d74de10bc39c38d30e

nexus-watcher/src/service/indexer/homeserver.rs
ok300: Good point, reworded in 2427ce5d32cca543c1f5a876a3b3ce25d3adad52

nexus-watcher/src/service/indexer/homeserver.rs
ok300: Indeed this check needed some rework.

The method was renamed to `user_hs_binding` (2427ce5d32cca543c1f5a876a3b3ce25d3adad52) and then to `user_hs_mapping` (04b27855d6e17783effebb82a5d654bd7d2491e3).

Initially I was going for `true` / `false` but later realized we need to also check if the mapping to this HS is stale, so now it's returning

```rust
/// A user's `HOSTED_BY` binding, classified relative to a processor's homeserver.
enum HsBinding {
    /// The user has no `HOSTED_BY` edge yet.
    Unbound,
    /// The user is bound to this processor's homeserver.
    Current { stale: bool },
    /// The user is bound to a different homeserver.
    Other { hs_id: String },
}
```

The caller can then clearly decide in which situation the event gets handled, in which not.

Let me know if this new shape is acceptable.

nexus-watcher/src/service/indexer/homeserver.rs
ok300: True, the situation was a bit ambiguous.

I think 2427ce5d32cca543c1f5a876a3b3ce25d3adad52 clarified things, now the log statement clearly describes and matches that case.

ok300:
@greptile please review again

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -46,6 +72,58 @@ impl TEventProcessor for HsEventProcessor {
...
+    /// - A stale edge to this homeserver (the user's published homeserver has
+    ///   diverged): log a warning and skip until the resolver realigns it.
+    /// - An edge to a different homeserver: log a warning and skip.
+    ///
+    /// A lookup failure fails open (the event is processed), so a transient
+    /// graph error never silently drops legitimate events.
+    async fn should_process_event(&self, event: &Event) -> bool {
+        let user_id = event.parsed_uri.user_id();
+
+        match self.user_hs_mapping(user_id).await {
```
tipogi: dont you think that logs are a bit noisy, or are we in that case were is it a bit rare to have that situations? thats why `warn!` implementation and not `debug!`

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -189,6 +189,16 @@ pub trait TEventProcessor: Send + Sync + 'static {
...
 
+    /// Decides whether the given event should be processed by this processor.
+    ///
+    /// Called at the start of [`Self::handle_event`], before the event is handed
+    /// to the event handler. The default always processes the event;
+    /// [`HsEventProcessor`](crate::service::indexer::HsEventProcessor) overrides
+    /// this to skip events from users bound to a different homeserver.
+    async fn should_process_event(&self, _event: &Event) -> bool {
+        true
+    }
```
tipogi: nice 🚀

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -206,6 +216,12 @@ pub trait TEventProcessor: Send + Sync + 'static {
     )]
     async fn handle_event(&self, event: &Event) -> Result<(), EventProcessorError> {
         let span = tracing::Span::current();
+
+        if !self.should_process_event(event).await {
+            span.record("otel.status_code", "OK");
```
tipogi: what about status code, `SKIPPED` or `HS_NOT_SYNC` instead of `OK`

tipogi (approved):
NIT comments, ready to merge 🛤️

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -46,6 +72,58 @@ impl TEventProcessor for HsEventProcessor {
...
+            }
+
+            // Lookup failed: fail open so transient graph errors don't drop events.
+            Err(e) => {
+                warn!(
+                    event.uri = %event.uri,
+                    user_id = %user_id,
+                    "Failed to resolve user's homeserver; processing event anyway: {e}"
+                );
+                true
```
aintnostressin: I would consider returning error here and decide (probably retry in place event line) handling at callsite

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -65,6 +143,35 @@ impl TEventProcessor for HsEventProcessor {
 }
 
 impl HsEventProcessor {
+    /// Resolves and caches a user's `HOSTED_BY` mapping relative to this processor's HS
+    async fn user_hs_mapping(&self, user_id: &PubkyId) -> GraphResult<HsMapping> {
```
aintnostressin, line 134: This looks like something that should belong to models.

aintnostressin (approved):

nexus-watcher/src/service/indexer/homeserver.rs
ok300: `HsMapping` is a "locally-relevant model", its only created and used in this `rs` file, so it seems best to keep it here.

nexus-watcher/src/service/indexer/homeserver.rs
ok300: Good idea, done: fa659937deb00eaf9b26ac40344b8c17c17768c0

nexus-watcher/src/service/indexer/mod.rs
ok300: Done in fa659937deb00eaf9b26ac40344b8c17c17768c0

nexus-watcher/src/service/indexer/homeserver.rs
ok300: It's pretty rare to have these, at least until we'll support full HS migrations. For now, these are rare (user changing HS) and when they happen, it's useful to have the details in the logs.

PR #903, opened 2026-06-04
fix: track "no HS found via PKDNS lookup" in failure metric

ok300:
This PR includes the cases when "PKDNS lookup found no HS" for a user into the `HS_RESOLVER_METRICS.run_failed` metric.


aintnostressin (approved):
Change looks good, however I doubt we should treat "not found" as a failure.

ok300:
IMO a failure is clear enough, we can add further granularity later on if needed.

PR #906, opened 2026-06-05
feat: blacklist external HSs from indexing and ingestion

ok300:
This PR adds a configurable blacklist of external HS PKs. While an HS PK is on the list, Nexus refuses to index events from it or to ingest users hosted on it.

The blacklist is configured via `StackConfig.external_hs_pk_blacklist` (so it applies to both the watcher and the webapi) and is enforced in the following places:

- **3rd-party HS indexing** (`nexus-watcher`)
  - `KeyBasedEventProcessorRunner` excludes blacklisted HSs from the set of HSs it monitors (`pre_run`).
  - `KeyBasedEventProcessor` re-checks before running and aborts on a blacklisted HS (defense-in-depth).

- **User ingestion**
  - Via the REST API (`PUT /v0/ingest/{user_id}`): returns `403 Forbidden` if user is hosted on a blacklisted HS.
  - Via event dependencies on a not-yet-ingested user hosted by a blacklisted HS — a follow of such a user, a tag on them or their post, or a reply/repost referencing their post — the event fails with a **non-retryable** error and is dropped (rather than queued for retry, since the dependency can never resolve).
  - Posts that merely **mention** such a user are still indexed; only the `MENTIONED` edge is not materialized.

- **File ingestion** (`nexus-watcher`)
  - A file whose `src` blob is hosted on a blacklisted HS is refused: Nexus does not reach out to the blacklisted HS, the blob is not downloaded, and no `FileDetails` is indexed.

### Notes / scope

- Existing data is **not** affected: users already ingested, and existing `HOSTED_BY` graph state, are left untouched when their HS is later blacklisted — only *new* indexing/ingestion is blocked. Removing an HS from the list does not recover events dropped while it was blacklisted.
- A new non-retryable `EventProcessorError::HsBlacklisted` (from `ModelError::HsBlacklisted`) drives the "drop instead of retry" behavior and the API's 403 response.

### Implementation

Because enforcement is spread across all 3 crates (`nexus-common`, `nexus-watcher`, `nexus-webapi`), the blacklist and the ingestion logic are encapsulated in a `UserIngestor` (built from config, holding an `HsBlacklist`), which is threaded through the event handlers and the API context.

As part of this, the user-ingestion logic was moved out of `UserDetails` into `UserIngestor`, and on successful ingestion the user is now bound to their HS (`HOSTED_BY` + `resolved_at`).

See `docs/decentralization.md` for the configuration/behavior reference.

---

# Pre-submission Checklist


nexus-watcher/src/events/handlers/post.rs
```diff
@@ -40,7 +41,7 @@ pub async fn sync_put(
                     .map_err(EventProcessorError::generic)?;
                 dependency_event_keys.push(replied_uri_str);
 
-                if let Err(e) = PostDetails::maybe_ingest_author_of_post(replied_to_uri).await {
+                if let Err(e) = ingestor.maybe_ingest_author_of_post(replied_to_uri).await {
```
SHAcollision: Umm, this means a reply/repost whose parent author is hosted on a blacklisted HS is queued for dependency retry, right? even though the dependency can never be ingested by policy. Maybe same thing happens in `follow.rs` and `tag.rs`.

Because `MissingDependency` is retryable, this will retry until max retries, repeatedly attempting an ingestion that the blacklist forbids. I guess a fix is to propagate the blacklist decision as a terminal/non-retryable error.

nexus-webapi/src/api_context.rs
```diff
@@ -46,12 +49,15 @@ impl ApiContextBuilder {
...
             None => {
                 let dc = DaemonConfig::read_or_create_config_file(self.config_dir.clone()).await?;
-                ApiConfig::from(dc)
+                // Enforce the watcher's HS blacklist on API-triggered ingestion too.
+                let ingestor = UserIngestor::from_config(&dc.watcher);
+                (ApiConfig::from(dc), ingestor)
             }
-            Some(ac) => ac.clone(),
+            // No daemon config in scope (explicit ApiConfig): ingest everything.
+            Some(ac) => (ac.clone(), UserIngestor::default()),
```
SHAcollision: Nit: this explicit `ApiConfig` branch constructs `UserIngestor::default()`, so `/v0/.../ingest-user` accepts users from blacklisted homeservers when the API is started from an `api-config.toml`. That means the API and watcher could enforce different blacklist policies in the same deployment, right?

nexus-common/src/config/watcher.rs
```diff
@@ -165,6 +165,10 @@ pub struct WatcherConfig {
     // Moderation
     pub moderation_id: PubkyId,
     pub moderated_tags: Vec<String>,
+
+    /// External HS PKs which are forbidden from being indexed.
+    #[serde(default)]
+    pub external_hs_pk_blacklist: Vec<PubkyId>,
```
SHAcollision: Nit: if we expect this list to become large, maybe consider using a `HashSet<PubkyId>` rather than `Vec<PubkyId>` . I don't think this is the case as of now.

SHAcollision (approved):
Good work. 

2 Nits: 
- we could document whether the blacklist is intended to affect already-ingested users or existing `HOSTED_BY` graph state. Currently preventing new ingestion and key-based processing, but it does nothing wit existing graph/index state for users on a homeserver that becomes blacklisted later.
- Could expose a test helper that constructs `DefaultEventHandler::from_config` with a blacklist and exercises the retry scheduler behavior end-to-end. The current tests cover filtering and direct API ingestion, but not dependency retry.

SHAcollision (changes requested):
Sorry, did not mean to "Approve", only to "Comment". Need to submit a "Request changes" now to "Un-approve" it.

aintnostressin (commented):
Two findings from my side, maybe not relevant/important or good candidates for follow up PRs to tighten checks:

1. `RetryProcessor` skips blacklist: `checks.process_retry_event` reconstructs and handles events from any `origin_homeserver_id` without checking it against the blacklist, so queued events from blacklisted HSs will still be fetched and processed on retry. 
2. `file::sync_put`  skips blacklist checks: Since `PubkyAppFile.src` can reference a file hosted on any HS (including a blacklisted one), Nexus will reach out to blacklisted HSs when processing file events.

nexus-watcher/src/events/handlers/post.rs
ok300: > Because `MissingDependency` is retryable, this will retry until max retries

Good question.

With only a basic understanding of the new retry manager, I would say this is ok because the user could update their PKDNS record and point to a non-blacklisted HS _before_ we reach max retries. Then that user would correctly be ingested. In other words, IMO this doesn't intersect or overlap with the retry handling, as it doesn't need special handling.

@aintnostressin how do you see this?

nexus-common/src/config/watcher.rs
ok300: Will be possible once https://github.com/pubky/pubky-app-specs/pull/128 is in.

For now I don't think it's critical.

nexus-watcher/src/events/handlers/post.rs
aintnostressin: If author of referenced post is on a blacklisted HS and because of his post being unavailable we cannot complete stop here and do not retry.

This is a different kind of a situation than a `MissingDependency` which we believe is a race condition between two event streams and will eventually self-heal.

ok300: I see, thanks.

Indeed, this is a sub-check inside the handling of a `MissingDependency`. I don't fully see the race condition, but I agree this will self-heal:
- user updates record and points to non-blacklisted-HS _before_ we reach max retries
  - then user is ingested (added to DB, ready for event indexing)
  - this also means the `MissingDependency` will resolve, since this user was the missing dep
- user keeps pointing to blacklisted-HS until we reach max retries
  - max retries run out, we stop because we cannot resolve this dependency

nexus-watcher/src/events/handlers/follow.rs
```diff
@@ -35,7 +35,7 @@ pub async fn sync_put(
             return Ok(());
         }
         OperationOutcome::MissingDependency => {
-            if let Err(e) = UserDetails::maybe_ingest_user(&followee_id).await {
+            if let Err(e) = ingestor.maybe_ingest_user(&followee_id).await {
                 tracing::error!("Failed to ingest user: {e}");
```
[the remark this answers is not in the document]
ok300: Fixed in fcc04b9d21bc505764375aef634a01dc6801f95f

nexus-common/src/models/user/ingestor.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// and persists the user node in the graph.
+    ///
+    /// If the resolved HS is blacklisted, throws  [`ModelError::HomeserverBlacklisted`].
+    #[tracing::instrument(name = "user.ingest", skip_all)]
+    pub async fn maybe_ingest_user(&self, user_id: &PubkyId) -> ModelResult<()> {
+        let user_id_str = user_id.to_string();
+        if UserDetails::get_by_id(&user_id_str).await?.is_some() {
+            tracing::debug!("Skipping ingestion: {user_id_str} already known");
+            return Ok(());
+        }
```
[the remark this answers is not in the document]
ok300: Added clarifying comments: fcc04b9d21bc505764375aef634a01dc6801f95f

nexus-webapi/src/api_context.rs
ok300: Good catch, fixed: 3596ea4e6d86f8ea601e15edaabf1e4229f3186a

ok300:
@greptile please review again

ok300:
@aintnostressin about your findings:

> 1. `RetryProcessor` skips blacklist: `checks.process_retry_event` reconstructs and handles events from any `origin_homeserver_id` without checking it against the blacklist, so queued events from blacklisted HSs will still be fetched and processed on retry.

IMO it's a short-enough window and it's fine for now if we retry those events. I documented that with d02fd09ff16453e5c8e369d46c2d6df7293292f2

> 2. `file::sync_put`  skips blacklist checks: Since `PubkyAppFile.src` can reference a file hosted on any HS (including a blacklisted one), Nexus will reach out to blacklisted HSs when processing file events.

Good catch, fixed in 1ad0958afa7a0ecf02348b61a893f72bfe14f119

ok300:
@SHAcollision about your findings:

> * we could document whether the blacklist is intended to affect already-ingested users or existing `HOSTED_BY` graph state. Currently preventing new ingestion and key-based processing, but it does nothing wit existing graph/index state for users on a homeserver that becomes blacklisted later.

Done, documented in [docs/decentralization.md](https://github.com/ok300/pubky-nexus/blob/ok300-dx-add-hs-blacklist/docs/decentralization.md) 

> * Could expose a test helper that constructs `DefaultEventHandler::from_config` with a blacklist and exercises the retry scheduler behavior end-to-end. The current tests cover filtering and direct API ingestion, but not dependency retry.

I added

- a non-retryable `HsBlacklisted` error that is thrown before `MissingDependency`, where appropriate
- a test to confirm this error (and this situation: necessary dependency hosted on a blacklisted HS) results in dropping the dependant too

in 2ff4d88a91a70440607dabf0f7c7b7e1f8123106

nexus-watcher/src/events/handlers/post.rs
ok300: Non-retryable error + handlnig was added, see https://github.com/pubky/pubky-nexus/pull/906#issuecomment-4682450368

nexus-common/src/models/user/ingestor.rs
tipogi, line 1: It does feel like this folder is not the right location. The models folder should contain low-level models that persist information in the graph/cache. This one feels like an upper-layer service.

nexus-watcher/src/events/handlers/utils.rs
```diff
@@ -1,12 +1,29 @@
...
+/// (`maybe_ingest_user` already logs the underlying failure.)
+pub(super) fn fail_on_blacklisted_hs(
+    ingest_result: ModelResult<()>,
+) -> Result<(), EventProcessorError> {
+    match ingest_result {
+        Err(e @ ModelError::HsBlacklisted { .. }) => Err(e.into()),
+        _ => Ok(()),
+    }
+}
+
```
tipogi, line 23: Interesting that function 🧑‍🏫

nexus-watcher/src/events/handlers/file.rs
```diff
@@ -51,7 +53,19 @@ async fn ingest(
     file_id: &str,
     pubkyapp_file: &PubkyAppFile,
     files_path: PathBuf,
+    ingestor: &UserIngestor,
 ) -> Result<FileMeta, EventProcessorError> {
+    let file_src = &pubkyapp_file.src;
+    let parsed_source_uri = ParsedUri::try_from(file_src.to_string()).map_err(|e| {
+        EventProcessorError::generic(format!("Invalid file source URI {file_src}: {e}"))
+    })?;
```
tipogi, line 61: why do we need to be that harsh, is there a possibility to have `http`/`https`/... urls right? according to [specs](https://github.com/pubky/pubky-app-specs/blob/main/src/models/file.rs#L115-L189), the allow list is more flexible. Why if the `ParsedUri` is successful, we execute the block of _64-68_ lines

nexus-watcher/src/events/retry/processor.rs
```diff
@@ -123,6 +122,11 @@ impl RetryProcessor {
         let ev_uri = &retry_event.event_uri;
         let ev_retry_count = retry_event.retry_count;
 
+        // In principle, it's possible to check if `origin_homeserver_id` is blacklisted before
+        // handling the event. A retry entry may have been queued before that HS got blacklisted.
+        // Retrying those pre-existing events is acceptable for now. Newly discovered events from a
+        // blacklisted HS are blocked before they can be enqueued.
+        //
```
tipogi, line 131: 👍

nexus-common/src/models/user/details.rs
```diff
@@ -124,46 +124,16 @@ impl UserDetails {
...
-            .await
-            .inspect(|_| tracing::info!("Ingested user {user_id} from homeserver {hs_id}"))
-            .inspect_err(|e| tracing::error!("Failed to ingest user {user_id}: {e}"))?;
-
-        // Store the start point of the homeserver cursor
-        UserHsCursor::write(user_id, hs_id, 0).await?;
+/// Binds a user to their homeserver, recording the `HOSTED_BY` relationship and `resolved_at`.
+pub async fn set_user_homeserver(user_id: &str, homeserver_id: &str) -> GraphResult<()> {
+    exec_single_row(queries::put::set_user_homeserver(user_id, homeserver_id)).await
+}
```
tipogi, line 132: good refactorization ✨

nexus-common/src/utils/test_utils.rs
tipogi, line 1: NIT: Not sure if I like that refactorization, adding in the `src` folder the test utils... AI had some suggestion, but I will let you decide, if it is ok or not

nexus-common/src/config/daemon.rs
```diff
@@ -102,6 +102,7 @@ mod tests {
                 "il_adult_nu_sex_act",
             ]
         );
+        assert!(c.stack.external_hs_pk_blacklist.is_empty());
```
aintnostressin, line 106: Would be good to test that list is correctly parsed.

nexus-common/src/models/post/details.rs
```diff
@@ -169,15 +168,6 @@ impl PostDetails {
         Ok(())
     }
 
-    /// If a referenced post is authored by a new, unknown user, this method triggers ingestion of that user.
-    ///
-    /// ### Arguments
-    ///
-    /// - `referenced_post_uri`: The parent post (if current post is a reply to it), or a reposted post (if current post is a Repost)
-    pub async fn maybe_ingest_author_of_post(referenced_post_uri: &ParsedUri) -> ModelResult<()> {
```
aintnostressin, line 177: I like this being moved out of model.

nexus-common/src/models/user/ingestor.rs
```diff
@@ -0,0 +1,99 @@
+use pubky_app_specs::{ParsedUri, PubkyId};
+
+use crate::db::PubkyConnector;
+use crate::models::error::{ModelError, ModelResult};
+use crate::models::homeserver::HsBlacklist;
+use crate::models::traits::Collection;
+use crate::models::user::{set_user_homeserver, UserDetails, UserHsCursor};
+use crate::StackConfig;
+
+/// Ingests previously-unknown users unless their HS is blacklisted.
```
aintnostressin, line 10: `UserIngestor` doesn't perform any ingestion - it only saves user, cursor and hosted_by relation.

`UserIngestor` tells this thing `ingests` something (user).
Rust doc says it ingests previously unknown users

There is `fn maybe_ingest_user` - to this point i think it really ingests.
I read source and rust doc and i realize it does not. 

This is preexisting issue but now it leaks to first class struct.
I would suggest renaming it so it better reflects what it does.

nexus-common/src/models/user/ingestor.rs
```diff
@@ -0,0 +1,99 @@
...
+            .inspect_err(|e| tracing::warn!("Aborting ingestion of {user_id}: {e}"))?;
+
+        let Some(hs_id) = maybe_hs_id else {
+            tracing::warn!("Skipping ingestion: {user_id} has no published HS or is an HS PK");
+            return Ok(());
+        };
+
+        let user_details = UserDetails::from_pubky(user_id.clone());
+
+        // Do not add to index, as this would affect the timeline of events for this user.
```
aintnostressin, line 93: Would `UserDetails::reindex(&user_ids_refs)` put this stub into index ?
If yes - how do we plan to prevent that?

nexus-common/src/models/error.rs
```diff
@@ -21,6 +21,10 @@ pub enum ModelError {
     #[error("FileOperationFailed")]
     FileOperationFailed(#[from] std::io::Error),
 
+    /// The user's HS is blacklisted; the user must not be ingested.
```
aintnostressin: Is this rust doc complete ? 
Files are also not ingested if hosted by blacklisted homeserver.

I would suggest avoid making any claims here how other components behave:
1. list is already incomplete
2. it drifts away with the time
3. keeping it in sync costs additional time

nexus-watcher/tests/user_ingestion/ingest_user_blacklisted_homeserver.rs
```diff
@@ -0,0 +1,103 @@
...
+        bio: Some("test_follow_of_user_on_blacklisted_homeserver".to_string()),
+        image: None,
+        links: None,
+        name: "Watcher:UserIngestion:FollowBlacklisted".to_string(),
+        status: None,
+    };
+    test.create_user(&follower_kp, &follower_user).await?;
+
+    let ingestor = UserIngestor::new([PubkyId::from(hs_pk.clone())]);
+    let err = handlers::follow::sync_put(
```
aintnostressin, line 96: I think there are more handlers to cover like this, can we have similiar tests ?

tipogi:
What about a user key that has been indexing and for some reason, we stop nexus to add that key homeserver to the blacklisted list? I think, to add notification type related that one might be useful. "Your homeserver was blacklisted. Your events are not going to be ingested by nexus"

nexus-watcher/src/events/handlers/file.rs
ok300: If you look a few lines below, you will see this method already had this logic before:
- if `file.src` is parsed as `PubkyAppObject::Blob`
  - ingest file, return `Ok`
- else
   - `Err("The file has a source uri that is not a blob path")`

So my diff here doesn't add or change this logic[^1].

My diff in this method addresses the 2nd point raised by @aintnostressin here: https://github.com/pubky/pubky-nexus/pull/906#pullrequestreview-4448774134

To track the idea of changing this logic, I added #934. I would not do it as part of this PR, as it's not related (the above file ingestion logic also exists on `main`).

nexus-common/src/utils/test_utils.rs
ok300: TBD: investigate if it can be moved to `common/utils/test.rs`

ok300: Done: 0baa6e4fe2f4244b2f8c7e4b3ddb91c8e55db614

nexus-common/src/models/post/details.rs
ok300: Done, its moved outside of the `PostDetails` model.

nexus-common/src/models/user/ingestor.rs
ok300: > Would `UserDetails::reindex(&user_ids_refs)` put this stub into index ?

Yes it would, in
- the `user_details` cache and
- the `UserSearc` name/ID search indexes

This is pre-existing behavior. This PR only relocated the stub-creation, it didn't introduce it.

nexus-common/src/models/user/ingestor.rs
ok300: > `UserIngestor` doesn't perform any ingestion - it only saves user, cursor and hosted_by relation.

That is exactly what ingestion means.

If you look in `main`, you will find that `maybe_ingest_for_user` does the same:

https://github.com/pubky/pubky-nexus/blob/a3d93962ab4661d9298eac69c40d3741c1472621/nexus-common/src/models/homeserver.rs#L178

If you have better ideas for the name, please propose them.

nexus-common/src/models/error.rs
ok300: Done: 04d942812e11aa3793bfeccff6fb0a5af263c9ab

nexus-watcher/tests/user_ingestion/ingest_user_blacklisted_homeserver.rs
ok300: Done: eb0869cac8c1e58024b4da50b7d738490f249d82

nexus-common/src/config/daemon.rs
ok300: Done: 82bf313cadb4362e8e49c63698d411e26320f412

nexus-common/src/models/user/ingestor.rs
ok300: I looked at different locations, but none seemed much better than here.

One thing that weighted much in favor of this location is, there are other user-related user-helper structs in `nexus-common/src/models/user/` so placing this next to them seemed natura.

If you see a better location for this, let me know.

nexus-common/src/models/user/ingestor.rs
aintnostressin: Ok, for me ingestion is more like a continous process but i must have misunderstood it.

Let's keep it as it is.

nexus-common/src/models/user/ingestor.rs
aintnostressin: Ony this one remain but it's not introduced here so i'm resolving it for now.

nexus-watcher/src/events/mod.rs
```diff
@@ -27,20 +28,34 @@ pub trait EventHandler: Send + Sync {
...
-        Self { moderation }
+    pub fn new(moderation: Arc<Moderation>, ingestor: Arc<UserIngestor>) -> Self {
+        Self {
+            moderation,
+            ingestor,
+        }
+    }
+
+    /// Builds a handler, deriving its moderation rules and user ingestor from config.
+    pub fn from_config(config: &WatcherConfig) -> Self {
```
aintnostressin, line 43: If each struct has this `from_config` function then objects will stop sharing dependencies between each other. I think we should avoid that and `new` function was enough.

I suggest creating `Moderation` and `Ingestor` at the spin up layer - the top most - and pass it down so single instances can be reused. 

If you think this PR is not a place to change it - let's just don't introduce another `from_config` constructor.

This rule should apply to other `from_config` functions in this PR as well.

aintnostressin: decided to keep it as it is.

aintnostressin (approved):

tipogi (approved):
LRTM!

PR #907, opened 2026-06-07
Simplify `PubkyClientError` structure, conversion from `pubky::Error`

ok300:
This PR simplifies the 3-level-deep nested `From<pubky::Error> for PubkyClientError` conversion, which was hard to understand and reason about.

Part of the reason that nesting was necessary was because `PubkyClientErrorr` derived `Clone, Serialize, Deserialize`, which are never really needed. So this PR removes those traits too. Neither `PubkyClientErrorr`, nor its parent `EventProcessorError` (which internally uses it) need to serialize/deserialize their content, so removing those traits doesn't affect any callers.

The `PubkyClientError` enum values are now slightly inconsistent:
- some are `ErrorType { message: String }` 
- some are `ErrorType(#[from] AnotherErrorType)`

IMO that's a minor inconvenience, more than justified by the clearer conversion logic in `From<pubky::Error> for PubkyClientError`.

### DIFF
--- a/nexus-common/src/db/connectors/pubky.rs
+++ b/nexus-common/src/db/connectors/pubky.rs
@@ -1,5 +1,5 @@
+use pubky::errors::{AuthError, BuildError, PkarrError, RequestError};
 use pubky::{Pubky, PubkyHttpClient, StatusCode};
-use serde::{Deserialize, Serialize};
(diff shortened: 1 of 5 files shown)


tipogi (approved):
good clean up! LGTM to merge

PR #910, opened 2026-06-08
fix: parse universal file uris

gillohner:
## Summary
- Allow file metadata URI parsing for universal app file namespaces under `/pub/<app>/files/<id>`.
- Keep Pubky ID validation while avoiding `ParsedUri`, which only accepts `pubky.app` file paths.
- Add focused tests for `pubky.app`, `mapky.app`, `eventky.app`, and non-file URI rejection.

## Validation
- `cargo fmt --check`
- `cargo test -p nexus-common file_key_from_uri`


gillohner:
accidentally chose wrong upstream repo...

PR #911, opened 2026-06-08
Test plugin tag dispatch

gillohner:
## Summary
- Add a dispatcher regression test proving app-specific `tags/` paths remain plugin-dispatched.
- Protects MapKy tags from being routed only through generic universal tag handling, which would bypass MapKy plugin Place/Post tag edges.

## Validation
- `cargo test -p nexus-watcher --lib dispatcher`


gillohner:
Opened against upstream by mistake; recreating against gillohner/pubky-nexus.

PR #913, opened 2026-06-09
fix: dropping large graph

aintnostressin:
Dropping large graphs will fail due to transaction limit. This change splits deletion into smaller txns. 

```
  thread 'main' (116158) panicked at nexus-webapi/src/mock.rs:54:14:                                                                                                                                                                                                                                                                                                                    
  Could not drop graph nodes.: Neo4j(Neo4jError { kind: Transient, code: "Neo.TransientError.General.MemoryPoolOutOfMemoryError", message: "The allocation of an extra 2.0 MiB would use more than the limit 1.4 GiB. Currently using 1.4 GiB. dbms.memory.transaction.total.max threshold reached" })                                                                                  
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace   
```

### DIFF
--- a/nexus-webapi/src/mock.rs
+++ b/nexus-webapi/src/mock.rs
@@ -46,8 +46,11 @@ impl MockDb {
         info!("Dropping Graph database...");
         let graph = get_neo4j_graph().expect("Failed to get Neo4j graph connection");
 
-        // drop and run the queries again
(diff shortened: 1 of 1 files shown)


aintnostressin:
@greptile review again

ok300 (approved):

tipogi (approved):

PR #914, opened 2026-06-09
fix: detect index existence

aintnostressin:
Fixes string match in index existence check.

```
thread 'main' (504782) panicked at nexus-webapi/benches/setup.rs:16:47:
called `Result::unwrap()` on an `Err` value: CommandFailed("\"Index\": already exists")
```


aintnostressin:
@greptile review again

ok300 (approved):

PR #920, opened 2026-06-11
chore: hash `RetryEvent` redis index keys via new `IndexKey` type

aintnostressin:
Replaces raw event URIs as Redis index keys (sorted-set members / JSON state keys) for the retry queue with a fixed-length `BLAKE3` hash, reusing the same hashing scheme as `resource_id` (now factored into `nexus_common::utils::hash_str_hex`).

Introduces `IndexKey`, a newtype wrapping the hashed key, so a raw unhashed URI can no longer be passed where a hashed key is expected. `RetryStore::put` now derives the key internally from `event.event_uri via IndexKey::for_uri`, removing the redundant caller-supplied key parameter throughout the retry processor and tests.

Why: avoids unbounded/odd-character URIs being used directly as Redis keys, and the new type prevents accidentally mixing raw URIs with hashed index keys


tipogi (approved):
LGTM

PR #926, opened 2026-06-15
fix: enforce file/resource fetch size limit

aintnostressin:
A malicious or buggy homeserver could serve an unbounded body that the watcher buffered fully into RAM and then wrote to disk. This caps every fetched body.

**Changes**
- `max_file_size` config (50 MiB default) for file blobs; fixed 2 MiB cap for JSON descriptors.
- `fetch_capped` enforces both: rejects early on `Content-Length`, then streams with a running byte-count to catch absent/lying headers. Used by the file, resource, and tag fetch paths.
- Oversized fetches → `FetchSizeExceeded` (non-retryable: logged, counted via `watcher.fetch.rejected`, not enqueued).
- Error-response bodies truncated to 4 KiB before embedding in errors.

**Tests**
- Unit: `fetch_capped` pre-check + streaming branches over a real `reqwest::Response` (lying/absent CL).
- Integration: oversized file → nothing written to graph or disk, ingestion continues; small file still ingested.


nexus-watcher/src/events/mod.rs
```diff
@@ -10,9 +11,198 @@ pub mod retry;
...
+    let mut buf = Vec::new();
+    let mut total = 0usize;
+
+    while let Some(chunk) = stream.next().await {
+        let bytes = chunk?;
+        total += bytes.len();
+        buf.extend_from_slice(&bytes);
+        if total >= cap {
+            return Ok((buf, total > max));
+        }
```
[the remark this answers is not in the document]
aintnostressin: @greptile it should be fixed, check

aintnostressin:
@greptile review again

nexus-watcher/src/service/processor.rs
tipogi: This `/events` response is still read with unbounded `response.text().await`. Since the homeserver controls this body and can ignore `limit`, can we use the same capped streaming read here before parsing lines?

nexus-watcher/src/events/mod.rs
```diff
@@ -10,9 +11,82 @@ pub mod retry;
...
+        .await
+        .map_err(|e| EventProcessorError::client_error(e.to_string()))?;
+    if exceeded {
+        return Err(EventProcessorError::FetchSizeExceeded(
+            buf.len() as u64,
+            max,
+        ));
+    }
+    Ok(buf)
+}
```
tipogi: What do you think about moving these response-related helpers into a separate module, e.g. `events/fetch.rs`? That would keep `events/mod.rs` focused on event routing and make it easier to reuse the capped fetch logic for `/events` polling too.

nexus-watcher/src/events/mod.rs
```diff
@@ -10,9 +11,82 @@ pub mod retry;
 
 pub use moderation::Moderation;
 
-pub async fn handle(event: &Event, moderation: Arc<Moderation>) -> Result<(), EventProcessorError> {
+/// Max bytes to read from an error response body.
+pub(super) const MAX_ERROR_BODY: usize = 4 * 1024;
+
+/// Max bytes for a JSON resource descriptor (user, post, tag, file meta, etc).
+pub(super) const MAX_RESOURCE_SIZE: usize = 2 * 1024 * 1024;
```
tipogi: agreed to hard code that ones. No need to add in `.toml` file

nexus-watcher/src/service/processor.rs
aintnostressin: Good catch.

In default homeserver impl event line can be max ~4160 bytes.
watcher's `events_limt` max is `1000`

So we need to cap it slighlty above 4 MB

nexus-watcher/src/events/mod.rs
aintnostressin: moved: [f97de44](https://github.com/pubky/pubky-nexus/pull/926/commits/f97de4488fd8411156cba74a6ac5eeab459b24ee)

nexus-watcher/src/service/processor.rs
aintnostressin: done: [9e40996](https://github.com/pubky/pubky-nexus/pull/926/commits/9e40996796296dba7eff90ca6c6ebd505f3e9741)

tipogi (approved):
RTM!

PR #929, opened 2026-06-17
fix: cap limit and skip params

aintnostressin:
Paginated endpoints used to deserialize raw `Pagination` and improvise defaults with `unwrap_or(...)`. Several had no cap at all. This adds `BoundedLimit` / `BoundedSkip` types that reject out-of-range values at deserialization, plus a clamp inside the Cypher builders as a backstop.

⚠️ Breaking changes

* Out-of-range now returns `400`, not `200.` Policy is reject, not clamp. Applies to limit over the max, limit=0, and skip over the max. Body: `{"error":"Invalid input: Failed to deserialize query string: limit: Invalid input: limit exceeds maximum of 1000"}`.
* Follows & user-search default dropped `200` → `50`. Omitting limit on `followers/following/friends/search` now returns `50` items, not `200`. Pass limit explicitly (up to `200`) for more.
* `/v0/events rejects` limit > `1000`. Previously silently clamped to `1000` and returned `200`; now `400`. Default (`500`) unchanged.


aintnostressin:
@greptile review

aintnostressin:
@greptile review again

aintnostressin:
@greptile review again

nexus-webapi/src/routes/v0/stream/users.rs
```diff
@@ -42,11 +44,12 @@ pub struct UserStreamQuery {
         ("timeframe" = Option<Timeframe>, Query, description = "Timeframe for sources supporting a range"),
         ("preview" = Option<bool>, Query, description = "Provide a random selection of size 3 for sources supporting preview. Passing preview ignores skip and limit parameters."),
```
ok300, line 45: > Passing preview ignores skip and limit parameters.

This is not true anymore.

[stream/users.rs:23](https://github.com/pubky/pubky-nexus/blob/c1a1668ee73972935b172c13d6dc71ae3c8020b7/nexus-webapi/src/routes/v0/stream/users.rs#L23): `preview=true` no longer really “ignores skip and limit.” The bounded pagination is deserialized before `build_user_stream_input`, so `/v0/stream/users?source=influencers&preview=true&limit=99999999` returns `400` before `Influencers::get_influencers` can replace pagination with the preview `(random_skip, 3)` behavior. This also affects `/v0/stream/users/ids`.

I'm not sure if it should be true (if there was a real reason for this), or if this sentence should be removed.

(A)

nexus-webapi/src/routes/v0/stream/users.rs
```diff
@@ -89,11 +92,12 @@ pub async fn stream_users_handler(
...
```
ok300, line 93: > Passing preview ignores skip and limit parameters.

Same as (A)

nexus-webapi/src/routes/v0/events.rs
```diff
@@ -69,7 +70,7 @@ pub async fn get_events_handler(Query(q): Query<EventsQuery>) -> Result<Response
 }
 
 fn parse_query(q: &EventsQuery) -> Result<(usize, Option<u64>), Error> {
```
ok300: NIT: this can't fail anymore, so it doesn't need to return `Result`.

nexus-webapi/src/models/bounded_limit.rs
```diff
@@ -0,0 +1,128 @@
...
+            type Value = usize;
+            fn expecting(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                f.write_str("a string containing a positive integer")
+            }
+            fn visit_str<E: de::Error>(self, v: &str) -> Result<usize, E> {
+                let n: u64 = v.parse().map_err(de::Error::custom)?;
+                usize::try_from(n).map_err(|_| de::Error::custom("limit out of range"))
+            }
+        }
+        let n = deserializer.deserialize_str(V)?;
```
ok300:
```suggestion
        let s = String::deserialize(deserializer)?;
        let n: u64 = s.parse().map_err(de::Error::custom)?;
        let n = usize::try_from(n).map_err(|_| de::Error::custom("limit out of range"))?;
```

NIT: the visitor can be rewritten in a much simpler way.

(B)

nexus-webapi/src/models/bounded_skip.rs
```diff
@@ -0,0 +1,113 @@
...
+            type Value = usize;
+            fn expecting(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                f.write_str("a string containing a non-negative integer")
+            }
+            fn visit_str<E: de::Error>(self, v: &str) -> Result<usize, E> {
+                let n: u64 = v.parse().map_err(de::Error::custom)?;
+                usize::try_from(n).map_err(|_| de::Error::custom("skip out of range"))
+            }
+        }
+        let n = deserializer.deserialize_str(V)?;
```
ok300:
```suggestion
        let s = String::deserialize(deserializer)?;
        let n: u64 = s.parse().map_err(de::Error::custom)?;
        let n = usize::try_from(n).map_err(|_| de::Error::custom("skip out of range"))?;
```

NIT: same as (B)

nexus-webapi/src/routes/v0/types.rs
```diff
@@ -25,9 +25,9 @@ pub(crate) fn resolve_tag_wot_depth(
 
 #[derive(Default, Deserialize, Debug, ToSchema)]
 pub struct TagsQuery {
```
ok300, line 27: NIT: Add small helper methods for `TagsQuery`.

Several handlers repeat:

- `query.skip_tags.map(|s| s.value())`
- `query.limit_tags.map(|l| l.value())`
- `query.limit_taggers.map(|l| l.value())`

Examples:

- `nexus-webapi/src/routes/v0/post/tags.rs:63-65`
- `nexus-webapi/src/routes/v0/post/tags.rs:73-75`
- `nexus-webapi/src/routes/v0/resource/mod.rs:88-90`
- `nexus-webapi/src/routes/v0/resource/mod.rs:139-141`
- `nexus-webapi/src/routes/v0/user/tags.rs:49-51`

Since `TagsQuery` is the shared abstraction, it could expose methods like:

- `skip_tags_value() -> Option<usize>`
- `limit_tags_value() -> Option<usize>`
- `limit_taggers_value() -> Option<usize>`

nexus-webapi/src/routes/v0/stream/users.rs
tipogi: had the same feedback. My approach would be to add in the docs how to use `preview` instead of change the  code. do you agree? Finally, it is an error, because you cannot use `limit`/`skip` params with `preview` and the user might be confused if it request `limit=8` and after we return just 3

tipogi (approved):
RTM!

nexus-webapi/src/routes/v0/stream/users.rs
ok300: > My approach would be to add in the docs how to use `preview` instead of change the code. (..)
> Finally, it is an error (...)

So you're saying
- `preview` with `skip` or `limit` should throw error
- the doc sentence "Passing preview ignores skip and limit parameters" should be changed to "When using preview, do not pass skip or limit params" or similar

If so, sounds good to me.

aintnostressin: I have to push back on this. It was never ignored the way you describe.

```
curl  -v   'http://localhost:8082/v0/stream/users?user_id=59w4pmgjze54dd4iim7ahd6wx8q33gnqj5kk4k4uuudt46idy5by&post_id=00000039YD9DP&reach=wot_2&preview=true&skip=10&limit=10aa'-H 'accept: application/json'
< HTTP/1.1 400 Bad Request
< content-type: application/json
< vary: origin, access-control-request-method, access-control-request-headers
< vary: accept-encoding
< access-control-allow-origin: *
< content-length: 99
< date: Mon, 22 Jun 2026 10:20:27 GMT
< 
* Connection #0 to host localhost left intact
{"error":"Invalid input: Failed to deserialize query string: limit: invalid digit found in string"}* URL rejected: Malformed input to a URL function
* Closing connection
curl: (3) URL rejected: Malformed input to a URL function
```

@ok300 if you have a clear path how to proceed with this - please suggest so  I can add it to this PR (even if it's unrelated).

nexus-webapi/src/routes/v0/events.rs
aintnostressin: It could never fail 😂 but i'll change it.

aintnostressin: done [4816886](https://github.com/pubky/pubky-nexus/pull/929/commits/48168860e9e6b6824a1ddd198bd403b5460b7a49)

nexus-webapi/src/routes/v0/types.rs
aintnostressin: done: [f2372fc](https://github.com/pubky/pubky-nexus/pull/929/commits/f2372fc8d11b8118d9b60cfba08289e93e85f312)

nexus-webapi/src/models/bounded_limit.rs
aintnostressin: done [765cbd6](https://github.com/pubky/pubky-nexus/pull/929/commits/765cbd6db310b762c0941f914a036562994f498d)

ok300 (approved):

nexus-webapi/src/routes/v0/stream/users.rs
ok300: As this behavior existed before (its unrelated to this PR), it can be left as is.

PR #932, opened 2026-06-18
fix(counts): add collections field and exclude collection-follows from bookmarks

### LINKED ISSUE #930: Fix /v0/user/{user_id}/counts for Collections
# Fix `/v0/user/{user_id}/counts` for Collections

**Area:** `nexus-common` (UserCounts + Cypher), `nexus-watcher` (post/bookmark handlers)
**Related:** #857, #881 · deleted-posts over-count split out to a separate issue.

## Source (Slack)

> **@secondl1ght** — updates needed on `/v0/user/{user_id}/counts`: a new collections field; bookmarks count is wrong because "following" a collection is a bookmark; posts count includes collections; (deleted posts included — pre-existing, tracked separately).

> **@SHAcollision** — Keep bookmark counts unless bookmarks go private on the homeserver (then drop them from Nexus). `posts` stays a total of all posts incl. replies and collections; serve `replies` and `collections` so the FE subtracts to show "main posts".

## Problem

Collections are now first-class posts (`kind='collection'`, #857/#881), so the endpoint is wrong two ways: no `collections` field, and `bookmarks` is inflated because following a collection writes a bookmark. `UserCounts` is served from Redis and backfilled from `get.rs::user_counts`. **Live watcher counters and the Cypher must apply identical filters** — `reindex` re-derives from the query, so a mismatch flips the number on cache rebuild.

## 1. Add `collections` field

Mirror the existing `replies` subset; `posts` stays the grand total (FE renders main = `posts − replies − collections`).

- `counts.rs` — add `pub collections: u32` with `#[serde(default)]`. No backfill needed: old cached entries read as `0` (the correct value today — no collections exist yet), and the first increment materializes the field via `JSON.SET $.collections`.
- `get.rs::user_counts` — `COUNT { (u)-[:AUTHORED]->(p:Post) WHERE p.kind = 'collection' } AS collections`.
- `post.rs` PUT (~120): increment when `post.kind == Collection` (kind in scope).
- `post.rs` DEL (~492): recover kind (`PostDetails::get_by_id`; node still exists, graph-delete runs last), decrement under the `post_in_index` guard.

## 2. De-inflate `bookmarks`

Exclude collection-targets so the count reflects real post bookmarks only.

- `get.rs::user_counts` — `COUNT { (u)-[:BOOKMARKED]->(p:Post) WHERE (p.kind IS NULL OR p.kind <> 'collection') }`.
- `bookmark.rs` — gate increment (line 50) / decrement (~78) on target not being a collection (look up target kind; target is guaranteed to exist).

> Deferred: a separate `collection_followers` count — depends on the bookmarks-privacy roadmap.

## Acceptance

- `collections` exposed; `posts` semantics unchanged; `bookmarks` excludes collection-follows.
- Live counters == Cypher after a forced `reindex`.
- Tests — webapi: counts returns `collections`, bookmarks excludes a bookmarked collection (update `docker/test-graph/mocks` seed). watcher: PUT/DEL collection moves `collections` not `replies`; bookmark on a collection leaves `bookmarks` unchanged.
- `fmt` + `clippy` clean.

SHAcollision:
fixes #930

### Known limitation: counts keyed on a target's mutable kind

`collections` counts a user's authored posts by kind, and `bookmarks` excludes bookmarks whose target is a collection. Both the live counters and the Cypher backfill read the target's current kind, so a bookmarked post that crosses the collection boundary can make the live `bookmarks` counter drift from the graph.

A post's kind can change two ways: an edit (the client does not expose this), and soft-delete, which rewrites a still-referenced deleted post as a `[DELETED]` post with kind `Short`. So deleting a followed (bookmarked) collection flips its kind to `Short`. The follow was suppressed from `bookmarks` at PUT, but the later un-follow is evaluated against the now-`Short` kind, so it can decrement a value that was never incremented. The counter is floored at 0 (no negative value, no failed read), so the worst case is a bounded undercount for that user. Other users' counters are likewise not repaired when a post they bookmarked changes kind (that would require scanning every bookmarker on each edit).

There is no automatic reindex, so this does not self-heal; a manual reindex recomputes the counts from the graph. The author's own `collections` counter is kept correct live across both edit and delete. This is a property of any live counter keyed on a target's mutable state, not specific to collections.

Accepted and documented rather than fixed: the trigger is narrow and the impact is a bounded, non-crashing undercount that a reindex corrects.

### DIFF
--- a/nexus-common/src/models/user/counts.rs
+++ b/nexus-common/src/models/user/counts.rs
@@ -18,6 +18,9 @@ pub struct UserCounts {
     pub unique_tags: u32,
     pub posts: u32,
     pub replies: u32,
+    // Defaulted so counts cached before this field existed still deserialize.
+    #[serde(default)]
+    pub collections: u32,
     pub following: u32,
     pub followers: u32,
     pub friends: u32,
--- a/nexus-watcher/src/events/handlers/bookmark.rs
+++ b/nexus-watcher/src/events/handlers/bookmark.rs
@@ -5,6 +5,7 @@ use nexus_common::models::user::UserCounts;
 use pubky_app_specs::{ParsedUri, PubkyAppBookmark, PubkyId, Resource};
 use tracing::debug;
 
+use super::utils::post_is_collection;
 use crate::events::EventProcessorError;
 
 #[tracing::instrument(name = "bookmark.put", skip_all, fields(user_id = %user_id, bookmark_id = %id))]
@@ -27,6 +28,11 @@ pub async fn sync_put(
         }
     };
 
+    // Decide collection-follow BEFORE any write: if this lookup failed after the
+    // graph/index write, the retry would see `existed = true` and skip the
+    // increment forever, leaving a real bookmark uncounted.
+    let target_is_collection = post_is_collection(&author_id, &post_id).await?;
+
     // Save new bookmark relationship to the graph, only if the bookmarked user exists
     let indexed_at = Utc::now().timestamp_millis();
     let existed =
@@ -46,7 +52,8 @@ pub async fn sync_put(
         .put_to_index(&author_id, &post_id, &user_id)
         .await?;
 
-    if !existed {
+    // A collection-follow is stored as a bookmark; don't count it.
+    if !existed && !target_is_collection {
         UserCounts::increment(&user_id, "bookmarks", None).await?;
     }
     Ok(())
@@ -71,10 +78,16 @@ pub async fn sync_del(user_id: PubkyId, bookmark_id: String) -> Result<(), Event
         .await?
         .is_some();
 
+    // Decide collection-follow BEFORE the cleanup: if this lookup failed after
+    // `del_from_index`, the retry would see `existed_in_index = false` and skip
+    // the decrement forever, leaving a real bookmark counted.
+    let target_is_collection = post_is_collection(&author_id, &post_id).await?;
+
     // 3. Redis cleanup (idempotent)
     Bookmark::del_from_index(&user_id, &post_id, &author_id).await?;
 
-    if existed_in_index {
+    // Mirror the PUT gate: collection-follows never incremented `bookmarks`.
+    if existed_in_index && !target_is_collection {
         UserCounts::decrement(&user_id, "bookmarks", None).await?;
     }
 
--- a/nexus-watcher/src/events/handlers/post.rs
+++ b/nexus-watcher/src/events/handlers/post.rs
@@ -15,7 +15,7 @@ use pubky_app_specs::{
 };
 use tracing::{debug, Instrument};
 
-use super::utils::post_relationships_is_reply;
+use super::utils::{post_is_collection, post_relationships_is_reply};
 
 #[tracing::instrument(name = "post.put", skip_all, fields(user_id = %author_id, post_id = %post_id))]
 pub async fn sync_put(
@@ -28,6 +28,7 @@ pub async fn sync_put(
     let post_details = PostDetails::from_homeserver(post.clone(), &author_id, &post_id);
     // We avoid indexing replies into global feed sorted sets
     let is_reply = post.parent.is_some();
+    let is_collection = post.kind == PubkyAppPostKind::Collection;
     // PRE-INDEX operation, identify the post relationship
     let mut post_relationships = PostRelationships::from_homeserver(&post);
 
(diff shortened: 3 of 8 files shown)


SHAcollision:
Known limitation: counters keyed on a target's mutable kind

`collections` counts a user's authored posts by kind, and `bookmarks` excludes bookmarks whose target is a collection. If an existing post's `kind` is edited across the collection boundary, the author's own `collections` counter is corrected live, but the `bookmarks` counters of other users who bookmarked that post are not (repairing them would mean scanning every bookmarker on each edit). There is no automatic reindex, so this does not self-heal; it would require a manual reindex. This is not specific to collections: any live counter keyed on a target's mutable state has the same property today. In practice the pubky client does not expose changing a post's kind, so this path is not reachable through normal use. I think this is worth tracking on a new issue.

nexus-watcher/src/events/handlers/bookmark.rs
```diff
@@ -74,7 +76,8 @@ pub async fn sync_del(user_id: PubkyId, bookmark_id: String) -> Result<(), Event
     // 3. Redis cleanup (idempotent)
     Bookmark::del_from_index(&user_id, &post_id, &author_id).await?;
 
-    if existed_in_index {
+    // Mirror the PUT gate: collection-follows never incremented `bookmarks`.
+    if existed_in_index && !post_is_collection(&author_id, &post_id).await? {
         UserCounts::decrement(&user_id, "bookmarks", None).await?;
     }
```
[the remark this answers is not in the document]
SHAcollision: Real, but the decrement is floored at 0 so it cannot go negative or fail to deserialize; the worst case is a bounded undercount, which we have accepted and documented in the PR body and tracked in #933.

nexus-watcher/src/events/handlers/bookmark.rs
```diff
@@ -74,7 +76,8 @@ pub async fn sync_del(user_id: PubkyId, bookmark_id: String) -> Result<(), Event
     // 3. Redis cleanup (idempotent)
     Bookmark::del_from_index(&user_id, &post_id, &author_id).await?;
 
-    if existed_in_index {
+    // Mirror the PUT gate: collection-follows never incremented `bookmarks`.
+    if existed_in_index && !post_is_collection(&author_id, &post_id).await? {
         UserCounts::decrement(&user_id, "bookmarks", None).await?;
```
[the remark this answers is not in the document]
SHAcollision: Confirmed reachable via soft-delete; we have accepted this bounded, floored-at-0 undercount (documented in the PR body, tracked in #933) rather than fix it now.

nexus-common/src/models/user/counts.rs
```diff
@@ -18,6 +18,9 @@ pub struct UserCounts {
     pub unique_tags: u32,
     pub posts: u32,
     pub replies: u32,
+    // Defaulted so counts cached before this field existed still deserialize.
+    #[serde(default)]
+    pub collections: u32,
```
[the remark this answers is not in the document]
SHAcollision: Collections are not deployed yet, so no cached count holds collection data and the serde default of 0 is correct; no migration is needed.

SHAcollision:
@greptile review again

nexus-watcher/src/events/handlers/bookmark.rs
```diff
@@ -46,7 +47,8 @@ pub async fn sync_put(
         .put_to_index(&author_id, &post_id, &user_id)
         .await?;
 
-    if !existed {
+    // A collection-follow is stored as a bookmark; don't count it.
+    if !existed && !post_is_collection(&author_id, &post_id).await? {
```
tipogi: Maybe better if we compute `is_collection` before `Bookmark::put_to_graph`? If `post_is_collection()` fails after the graph edge/index write, the event retries with the bookmark already present, so `existed` becomes true. For a normal post bookmark, that means the bookmark can be indexed without ever incrementing `UserCounts.bookmarks`.

nexus-watcher/src/events/handlers/bookmark.rs
```diff
@@ -74,7 +76,8 @@ pub async fn sync_del(user_id: PubkyId, bookmark_id: String) -> Result<(), Event
     // 3. Redis cleanup (idempotent)
     Bookmark::del_from_index(&user_id, &post_id, &author_id).await?;
 
-    if existed_in_index {
+    // Mirror the PUT gate: collection-follows never incremented `bookmarks`.
+    if existed_in_index && !post_is_collection(&author_id, &post_id).await? {
```
tipogi: Same here. if `post_is_collection()` fails after Redis cleanup, the graph edge is still present for retry, but `existed_in_index` will be false on retry. That means a normal post bookmark can be removed from the Redis bookmark index without ever decrementing `UserCounts.bookmarks`.

nexus-watcher/src/events/handlers/bookmark.rs
SHAcollision: Good catch, fixed in 52c24c8e: `post_is_collection` now runs before `put_to_graph`.

nexus-watcher/src/events/handlers/bookmark.rs
SHAcollision: Fixed in 52c24c8e: the check now runs before `del_from_index`.

SHAcollision:
Soft-delete double-decrement fixed in 52c24c8e (thanks @tipogi): `sync_edit` persists the new kind before the `collections` counter moves, so a retry sees the new kind and the transition no longer fires. Added a soft-delete regression test.

nexus-watcher/src/events/handlers/post.rs
```diff
@@ -482,6 +499,10 @@ pub async fn sync_del(author_id: PubkyId, post_id: String) -> Result<(), EventPr
     let is_reply =
         matches!(&post_relationships_opt, Some(relationship) if relationship.replied.is_some());
 
+    // Kind isn't in the relationships; recover it from PostDetails (still present,
+    // graph-delete runs last).
+    let is_collection = post_in_index && post_is_collection(&author_id, &post_id).await?;
```
tipogi: some as bookmarks. Lets compute `is_collection` before `PostRelationships::delete` removes the retry gate? If `post_is_collection()` fails after the gate is deleted, retry observes `post_in_index = false` and skips the `collections` decrement, leaving a hard-deleted collection counted.

tipogi (approved):

nexus-watcher/src/events/handlers/post.rs
SHAcollision: Fixed in e4737d1d: `is_collection` is now computed before `PostRelationships::delete`, so a failed lookup cannot strand the decrement on retry.

tipogi (approved):
LGTM! merge it

PR #940, opened 2026-06-23
fix: build key-based event processor without Redis homeserver lookup

### LINKED ISSUE #939: bug: key-based runner should not load homeserver from Redis index
## Summary

`KeyBasedEventProcessorRunner::build` loads the homeserver via `Homeserver::get_by_id`, which reads **Redis**. External homeservers are never written to the Redis homeserver index, so processor builds fail with `"Homeserver not found"` even when the HS is already known in Neo4j and has ingested users.

External (key-based) indexing uses **per-user** cursors (`UserHsCursor`), not a homeserver-level cursor. The runner only needs the HS public key to poll user event streams — it does not need a Redis `Homeserver` record.

### Current flow

1. A user on an external HS is ingested (`UserIngestor::maybe_ingest_user`):
   - Creates stub `User` in Neo4j
   - `MERGE`s `Homeserver` + `HOSTED_BY` via `set_user_homeserver`
   - Writes per-user cursor at `0` via `UserHsCursor::write`
   - Does **not** call `Homeserver::put_to_index`

2. `KeyBasedEventProcessorRunner::pre_run` discovers the HS via `Homeserver::get_all_active_from_graph()` (Neo4j).

3. `build` fails:

   ```rust
   let homeserver = Homeserver::get_by_id(homeserver_id)
       .await?
       .ok_or("Homeserver not found")?;
   ```

   `get_by_id` is Redis-only (no graph fallback). External HS entries are never seeded there — only the **default** HS is, via `persist_if_unknown` at watcher startup.

4. Build failure → `ProcessorRunStatus::FailedToBuild` → exponential backoff. The key-based processor never runs, so **no events are fetched** from that HS.

### Why Redis is unnecessary here

| Concern | Default HS (`HsEventProcessor`) | External HS (`KeyBasedEventProcessor`) |
| --- | --- | --- |
| Cursor storage | Redis `Homeserver.cursor` | Redis `UserHsCursor` (per user) |
| Target list source | Config + `persist_if_unknown` | Neo4j `get_all_active_from_graph` |
| Uses `homeserver.cursor` at poll time? | Yes (`/events/?cursor=…`) | No — only `homeserver.id` |

`KeyBasedEventProcessor::run_internal` only uses `self.homeserver.id` (and `to_public_key()`). Cursors come from `UserHsCursor::read` per user. The `Homeserver` struct's `cursor` field is unused in this path.

tipogi:
# Pre-submission Checklist

### DIFF
--- a/nexus-watcher/src/service/indexer/key_based.rs
+++ b/nexus-watcher/src/service/indexer/key_based.rs
@@ -4,9 +4,10 @@ use crate::errors::EventProcessorError;
 use crate::events::Event;
 use futures::StreamExt;
 use nexus_common::db::PubkyConnector;
-use nexus_common::models::homeserver::{Homeserver, HsBlacklist};
+use nexus_common::models::homeserver::HsBlacklist;
 use nexus_common::models::user::UserHsCursor;
 use pubky::{Event as StreamEvent, EventCursor, PublicKey};
+use pubky_app_specs::PubkyId;
 use tokio::sync::watch::Receiver;
 use tracing::{debug, error, info, warn};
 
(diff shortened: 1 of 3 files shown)


ok300 (approved):
Since we're `KeyBasedEventProcessor` only needs the HS ID, I replaced the `Homeserver` field with a `PubkyId` for the HS ID: 84dade1dfaed769a13eb70393e3becc22c2bb53e

Let me know if you see any issues with that.

PR #942, opened 2026-06-23
fix: only read limit events from `/events-stream`

ok300:
This is an alternative implementation of #941.

It enforces the `limit` when reading from `/events-stream`, which prevents malicious external HSs from serving large amounts of data, which would then be read, loaded, parsed and processed by Nexus.

---
# Pre-submission Checklist

### DIFF
--- a/nexus-watcher/src/service/indexer/key_based.rs
+++ b/nexus-watcher/src/service/indexer/key_based.rs
@@ -53,8 +53,18 @@ impl KeyBasedEventSource for PubkyKeyBasedEventSource {
             .await
             .inspect_err(|e| error!("Failed to subscribe to event stream: {e:?}"))?;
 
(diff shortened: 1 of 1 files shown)


aintnostressin:
What do you think about returning error instead of capping events at the limit ? If homeserver is misbehaving we should not accept it.

aintnostressin (approved):
Looks fine. 
The PR description is a bit over promising. It only prevents large amounts of events from being processed by watcher.
The decision how to behave when homeserver misbehaves (returns more events than requested) and capping response size at byte level will be done as a follow ups.

ok300:
The follow-up issues are:
- https://github.com/pubky/pubky-nexus/issues/943
- https://github.com/pubky/pubky-core/issues/457

PR #948, opened 2026-06-26
fix: bounded cypher queries

aintnostressin:
Adds bounded keyset pagination for graph-bases queries and post notificaiton fan out. This should prevent from loads of objects being loaded into memory at once.


aintnostressin:
@greptile review

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -86,25 +86,34 @@ pub fn post_bookmark(author_id: &str, post_id: &str, viewer_id: &str) -> Query {
...
 
-// Get all the bookmarks that a post has received (used for edit/delete notifications)
-pub fn get_post_bookmarks(author_id: &str, post_id: &str) -> Query {
+// BOOKMARKED has no unique/indexed property, so keyset pagination uses elementId(b).
+pub fn get_post_bookmarks(author_id: &str, post_id: &str, cursor: &str, limit: i64) -> Query {
     Query::new(
         "get_post_bookmarks",
         "MATCH (bookmarker:User)-[b:BOOKMARKED]->(p:Post {id: $post_id})<-[:AUTHORED]-(author:User {id: $author_id})
-         RETURN b.id AS bookmark_id, bookmarker.id AS bookmarker_id",
+         WHERE elementId(b) > $cursor
```
tipogi: NOTE: `elementId()` is only officially guaranteed within one transaction, while each page here runs in a separate transaction. Given our single-instance deployment and short scans, the practical risk is very low. The remaining edge case is a bookmark relationship being deleted and recreated during a fan-out exceeding one page, which could move it across the cursor and skip or duplicate a notification. The same applies to tags at line _203_

nexus-common/src/db/graph/pagination.rs
```diff
@@ -0,0 +1,147 @@
...
+    Fut: Future<Output = Result<(usize, Option<String>), E>>,
+{
+    let mut cursor = String::new();
+    loop {
+        let (count, last_cursor) = step(cursor).await?;
+        if count == 0 || count < batch_size {
+            break;
+        }
+        match last_cursor {
+            Some(next) => cursor = next,
```
tipogi: *Non-blocking. robustness*: the helper does not verify that `next` is strictly greater than the `current` cursor. An unchanged or decreasing cursor could repeatedly fetch the same full page. Also, when a full page returns no cursor, the `None` branch logs and stops but ultimately returns Ok(()), so remaining records would be silently skipped. The current queries should produce valid monotonic cursors under normal conditions, but could we return an error for missing, empty, or non-increasing cursors to make this reusable driver fail safely?

Issue #935, opened 2026-06-20
user_counts query is O(tags x posts), hangs on heavy users


SHAcollision:
## What

`user_counts` (`nexus-common/src/db/graph/queries/get.rs`) is pathologically slow for users with many received tags. For a heavy user it does not finish.

## Why

It starts with `OPTIONAL MATCH (u)<-[t:TAGGED]-(:User)`, which produces one row per received tag. The next `WITH` mixes `COUNT(DISTINCT t.label)` with several `COUNT { ... }` subqueries (posts, followers, friends, ...). Those subqueries become grouping keys, so Neo4j evaluates each one once per row. `posts` (a 100k scan) runs once per received tag. A user with 20k tags and 100k posts is ~2e9 ops.

`post_counts` does NOT have this bug: it aggregates the tag count into a single row before its `COUNT { }` subqueries, so they run once.

## Repro

Synthetic whale: 100k authored posts, 100k followers, 5k following, 20k received tags. The query ran 30+ minutes and never returned (transaction had to be terminated).

## Impact

`UserCounts::get_from_graph` runs on cache miss and on reindex. Redis usually answers first, so it stays hidden, but a whale cache-miss or a reindex over a whale will hang.

## Fix

Compute each count as an independent `COUNT { }` subquery off the single `(u)` row, with no leading row-multiplying `OPTIONAL MATCH`. Same whale returns in ~310 ms with that shape. Keep each field's exact semantics (e.g. `tagged` stays `TAGGED->User` + `TAGGED->Post`, not all outgoing tags; validate the rewrite returns identical values to the current query on a normal user).

PR #952, opened 2026-06-30
chore(bench): graph query performance audit harness

SHAcollision:
## What

Tooling for the graph query performance audit (#953): a reusable harness under `docker/bench-graph/` to find graph queries whose cost grows faster than the data (the class behind #935).

## How it works

Seeds a synthetic graph at multiple scales, runs every query under Neo4j `PROFILE`, and reports **baseline-subtracted** db hits with a **fitted growth exponent** (~1 linear, ~2 quadratic, ~3 cubic), so the signal is the cost *shape*, not a single ratio.

| Piece | Role |
|---|---|
| `seed.sh` | Parametric synthetic graph (`:Bench`-labeled, `audit_*` ids, one-shot `DETACH DELETE`; real data untouched). |
| `examples/graph-audit` (Rust bin) | Emits `PROFILE <cypher>` for each query from the **real** query fns via `Query::to_cypher_populated()` (no drift). Reads tagged `@@QUERY`; writes tagged `@@WRITE` and profiled under `:begin`/`:rollback`, so they never mutate. |
| `run.sh` | Profiles a baseline (real data only) + scales {1,3,10}, subtracts the baseline, fits the exponent, dumps per-query PROFILE plans, writes a worst-first report. |
| `equiv.sh` | Proves a query rewrite returns identical results (used to verify the fixes). |
| `concurrency.sh` | Server-side contention probe. |

## Run

```sh
cd docker && docker compose up -d
./docker/bench-graph/run.sh   # writes docker/bench-graph/report.md (gitignored)
```

## Coverage

Whole graph-query surface (50 `queries/` fns + reindex + search), plus a caller-side N+1 sweep and an index/schema audit. The anti-pattern checklist in the README is grounded in the Neo4j 5.26 docs; PROFILE db hits are the ground truth.

Findings, metrics, and verified fixes: **#953**.


SHAcollision:
Baseline audit produced by this harness: #953 (Tier-0 read queries, query-by-query metrics + recommendations).

Issue #953, opened 2026-06-30
Graph query performance audit


SHAcollision:
# Graph query performance audit (full surface)

Performance audit of the whole graph-query surface: all 50 `queries/` fns plus reindex and full-text search, with a caller-side N+1 sweep, an index/schema audit, and a concurrency probe. Goal: find read queries whose cost grows faster than the data (so they hang on heavy users, e.g. #935) and land verified, behavior-preserving fixes. Tooling and full methodology: #952.

## Method
- **db hits** = PROFILE `Total database accesses` (the ground-truth cost metric), **baseline-subtracted**: the real dev-graph data is profiled on its own and removed, so the figures are the work the synthetic load alone causes. Fixes are proven result-identical before measuring. Numbers are at the synthetic scale; the signal is the cost *shape*, not the absolute.
- **exp** = growth exponent fitted over synthetic-graph scales M ∈ {1, 3, 10}, where M is the seed's size multiplier (~1 linear, ~2 quadratic, ~3 cubic). Tables cite db hits at a stated M. Useful regularity: for the cartesian pattern below, **exp ≈ the number of branches** (3 `OPTIONAL MATCH` → exp ~3), which is how the structure was confirmed.
- Caveats: warm-cache only (cold-disk behavior unprofiled); the concurrency probe hits server connections, not the app's 16-connection neo4rs pool.

## Headline
The dominant problem is one pattern: **N independent `OPTIONAL MATCH` branches collapsed by `COUNT(DISTINCT)`**, which the planner runs as a cartesian product per row. Four instances; each has a healthy sibling already using the fix (independent `COUNT{}`/`CALL{}` subqueries). Legend: ✅ DONE = fixed and merged (PR noted); 🟡 WIP = in an open PR; 🔵 = needs a design/feature decision; ⚪ = observation.

## A. Cartesian aggregation (super-linear)
| id | query (loc) | metric | fix (short) | status |
|---|---|---|---|---|
| F1 | `global_tags_by_post_engagement` (get.rs:234) | 4 branches; **exp n/a (times out at M≥3 before it can be fit)**; 154M hits at M=1 | each engagement count as a `COUNT{}` off `(post)` | ✅ DONE #954; **154M→28k** (M=1), no longer times out |
[...]

PR #954, opened 2026-06-30
perf(graph): query performance audit fixes (#953)

SHAcollision:
Performance fixes from the graph query audit (#953). Six independent, behavior-preserving query optimizations, **one commit each** for per-fix review.

| fix | what | db hits (before → after) |
|---|---|---|
| F1 | `global_tags_by_post_engagement`: 4 `OPTIONAL MATCH` (cartesian) → `COUNT{}` per post | 154M → 28k (M=1; was timing out at scale) |
| F2 | `get_global_influencers`: 3 `OPTIONAL MATCH` → `CALL(user){}` per count (mirrors `get_influencers_by_reach`) | 1.9M → 60k (M=3) |
| F7 | `post_stream` TotalEngagement: 3 `OPTIONAL MATCH` → `COUNT{}` | 6.96M → 8.9k (M=3) |
| F8 | per-post queries (×5): seek `Post` by its unique id + verify author, instead of expanding all the author's posts | e.g. `get_post_by_id` 1217 → 19 (M=3) |
| F9 | `user_/post_is_safe_to_delete`: `EXISTS{}` instead of counting every relationship | 2,707 → 3 |
| IDX-1 | drop 4 dead no-op `CREATE INDEX` (the UNIQUE constraints already index those schemas) | n/a |

**Pure performance, no behavior change.** Each rewrite was diffed old-vs-new on a synthetic graph and shown **result-identical**, including edge cases (author mismatch, post-not-found, zero-relationship/clean nodes, each allowed-relationship branch for F9). The cartesian / per-post plans simply stop multiplying work. Numbers are at the audit's synthetic scale; the wins grow with data.

**Validation:** no new Cypher surface, `COUNT{}` and scoped `CALL(var){}` already ship on `main` (`user_counts`, `get_influencers_by_reach`); CI runs the watcher/webapi integration suites over these paths against the pinned `neo4j:5.26.27`.

Harness + methodology: #952. Findings: #953.

### DIFF
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -239,11 +239,13 @@ pub fn global_tags_by_post_engagement() -> Query {
         WITH post, COUNT(tag) AS tags_count, tag.label AS label, author.id + ':' + post.id AS key
         WITH DISTINCT key, label, post, tags_count
         WHERE tags_count > 0
-        OPTIONAL MATCH (post)<-[reply:REPLIED]-()
-        OPTIONAL MATCH (post)<-[repost:REPOSTED]-()
-        OPTIONAL MATCH (post)-[mention:MENTIONED]->()
-        OPTIONAL MATCH (post)<-[tagged:TAGGED]-()
-        WITH COUNT(DISTINCT tagged) AS taggers, COUNT(DISTINCT reply) AS replies_count, COUNT(DISTINCT repost) AS reposts_count, COUNT(DISTINCT mention) AS mention_count, key, label
+        // Each engagement count is its own COUNT{} subquery, so they don't
+        // multiply into a cartesian product per post.
+        WITH key, label,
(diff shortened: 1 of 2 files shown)


nexus-common/src/db/graph/queries/get.rs
```diff
@@ -21,7 +21,11 @@ pub fn get_post_by_id(author_id: &str, post_id: &str) -> Query {
     Query::new(
         "get_post_by_id",
         "
-            MATCH (u:User {id: $author_id})-[:AUTHORED]->(p:Post {id: $post_id})
+            // The WITH barrier is load-bearing: without it the planner expands all
```
SHAcollision: This PR purposefully adds "Why" comments on the decisions that matter for perf. It is likely we keep creating poorly optimized queries in the future, so these comments are valuable.

aintnostressin (approved):
Nice improvements!

PR #971, opened 2026-07-02
fix(moderation): route moderated post deletion through the safe-to-delete gate

### LINKED ISSUE #961: Moderated post deletion bypasses the safe-to-delete gate, orphaning tag/bookmark indexes and permanently inflating counts
The normal post-delete path, `post::del` (`handlers/post.rs:436-471`), first runs `post_is_safe_to_delete`: a post with no other edges is hard-deleted via `sync_del`, but a post that still has tags, bookmarks, or incoming replies is **tombstoned** (content replaced with `[DELETED]`) rather than removed, precisely because `sync_del` assumes an edge-free post.

Moderation skips that gate. `apply_moderation` calls `handlers::post::sync_del(user_id, post_id)` directly for a `Resource::Post` (`events/moderation.rs:57-61`), while every other moderated resource type correctly uses its full `del` handler (`tag::del`, `user::del`, `file::del`). `sync_del` cleans only the post's own `PostCounts`/`PostDetails`, the parent's reply/repost counts, and then `DETACH DELETE`s the graph node. It never touches the `TagPost` tagger sets, `PostsByTagSearch` timelines, `Bookmark` entries, or the taggers' `tagged` / bookmarkers' `bookmarks` counts.

**Impact:** moderation targets exactly the highly-engaged content where this does the most damage. Moderating a post with, say, 50 tags, 20 bookmarks, and 10 replies leaves:
- the post as a dangling member of every `Tags:Global:Post:Timeline:{label}` sorted set (tag streams return ids that hydrate to nothing),
- all 50 taggers with a permanently inflated `tagged` count and stale tagger-set membership,
- all 20 bookmarkers with a phantom `bookmarks` count and a `Bookmark` entry pointing at nothing,
- 10 replies with Redis relationships to a nonexistent parent.

None of this self-heals; `sync_del` returns `Ok`, so retry does not help.

**Suggested fix:** route moderated post deletion through `post::del` (tombstone when edges exist), or extend `sync_del` to enumerate and clean the tag/bookmark/reply indexes and their counts when it is invoked on a post that still has such edges.

SHAcollision:
Moderation deleted posts by calling `sync_del` directly, skipping the `post_is_safe_to_delete` gate every other resource type goes through. A moderated post that still had tags, bookmarks or replies was DETACH DELETEd, leaving tagger/bookmarker counts and tag timelines permanently orphaned, and destroying the edges means the users' own later tag/bookmark DELs hit the "edge already gone" no-op, so it never self-heals.

Fix: route `Resource::Post` through `post::del`, threading the `UserIngestor` the tombstone path needs. Behavior change: a moderated post with engagement is now tombstoned (`[DELETED]`) like a user-deleted one instead of vanishing with corrupted indexes; edge-free posts are hard-deleted as before.

The new test tombstones a tagged post, then has the tagger delete their tag and asserts the cleanup actually lands (red on main at four points).

Fixes #961

## Pre-submission Checklist

- [x] I asked one or more LLMs to check if this PR can be simplified
- [ ] If appropriate, I added performance benchmarks for the APIs added in this PR (n/a)

### DIFF
--- a/nexus-watcher/src/events/mod.rs
+++ b/nexus-watcher/src/events/mod.rs
@@ -138,7 +138,7 @@ pub async fn handle_put_event(
         (PubkyAppObject::Tag(tag), Resource::Tag(tag_id)) => {
             if moderation.should_delete(&tag, user_id.clone()) {
                 moderation
-                    .apply_moderation(tag, event.files_path.clone())
+                    .apply_moderation(tag, event.files_path.clone(), &ingestor)
                     .await?
             } else {
                 // Route universal tag events (non-pubky.app apps) to sync_put_resource
--- a/nexus-watcher/src/events/mod.rs
+++ b/nexus-watcher/src/events/mod.rs
@@ -147,7 +147,9 @@ pub async fn handle_put_event(
         }
         (PubkyAppObject::Tag(tag), Resource::Tag(tag_id)) => {
             if moderation.should_delete(&tag, &user_id) {
-                moderation.apply_moderation(tag, files_path).await?
+                moderation
+                    .apply_moderation(tag, files_path, &ingestor)
+                    .await?
             } else {
                 // Route universal tag events (non-pubky.app apps) to sync_put_resource
                 // which handles Resource nodes for InternalUnknown/InternalUnknown URIs.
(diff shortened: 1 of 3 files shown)


SHAcollision:
@greptile review again

aintnostressin (approved):
Looks ok, and it's consistent with how the other resources are handled by moderation.

Separate from this PR: I'd rather `handlers::*::del` stayed an event-handler entry point and moderation went through something that doesn't inherit DEL-event semantics. That ship has already sailed for `tag`/`user`/`file`, so no objection to `post` joining them here

PR #978, opened 2026-07-06
chore: gate postgres behind dev compose profile

tipogi:
## Summary
Make Postgres optional in the local Docker stack using Compose profiles, and pin the Compose project/network name so external containers can join reliably.

Postgres is only needed for `nexus-watcher` integration tests (homeserver scenarios). Previously, `docker compose up -d` always started it alongside Neo4j and Redis. This change keeps the default stack lean and starts Postgres only when the dev profile is enabled.

Also pins the Compose project and Docker network to `pubky-nexus`, so other containers or compose projects can attach via a stable network name instead of the directory-derived _docker_default_.

### Usage

```bash
cd docker
# Default: Neo4j + Redis + Redis Insight
docker compose up -d
# With Postgres (watcher tests)
docker compose --profile dev up -d
# Or set once in .env
COMPOSE_PROFILES=dev
```

### Pre-submission Checklist

### DIFF
--- a/docker/.env-sample
+++ b/docker/.env-sample
@@ -1,3 +1,6 @@
+# Uncomment to always start optional services (redisinsight, postgres) with `docker compose up -d`
+# COMPOSE_PROFILES=dev
+
 # Neo4J Community Edition does not support custom db_name and db_username
 # the default ones are `neo4j`
 NEO4J_DB_NAME=neo4j
--- a/docker/.env-sample
+++ b/docker/.env-sample
@@ -1,4 +1,4 @@
-# Uncomment to always start optional services (redisinsight, postgres) with `docker compose up -d`
+# Uncomment to always start optional services (postgres) with `docker compose up -d`
 # COMPOSE_PROFILES=dev
 
 # Neo4J Community Edition does not support custom db_name and db_username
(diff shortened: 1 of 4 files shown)


docker/.env-sample
```diff
@@ -1,3 +1,6 @@
+# Uncomment to always start optional services (postgres) with `docker compose up -d`
+# COMPOSE_PROFILES=dev
```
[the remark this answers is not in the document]
tipogi: addressed: https://github.com/pubky/pubky-nexus/pull/978/commits/a1cf329e2e73a025be59408a2f0dda936e574883

docker/docker-compose.yml
```diff
@@ -57,8 +59,9 @@ services:
       - redis
     restart: always
 
-  # Only used for tests that involve homeservers
+  # Only used for tests that involve homeservers (requires --profile dev)
```
aintnostressin: NIT: seems like profile name `tests` fits the purpose better

aintnostressin (approved):

docker/docker-compose.yml
tipogi: makes sense!

PR #984, opened 2026-07-08
feat: add trust score to user nodes

aintnostressin:
Adds seeded pagerank trust scores to User nodes.

* Trust scores are computed either by cron job or ad hoc by CLI
* PageRank teleportation vector factor can be cusomized through `alpha` parameter. It also affects damping factor which is `1 - alpha`
* Trust computation runs till one of two conditions are met: max iterations or convergance
* Scheduled recomputation can be disabled
* Contains optional report generation

### DIFF
--- a/nexusd/src/trust/engine.rs
+++ b/nexusd/src/trust/engine.rs
@@ -15,6 +15,8 @@ pub struct TrustRankParams {
     pub max_iterations: u32,
     /// Convergence tolerance.
     pub tolerance: f64,
+    /// Optional hard cap on the GDS projection size (bytes).
+    pub max_projection_bytes: Option<u64>,
 }
 
 impl From<&TrustRankConfig> for TrustRankParams {
@@ -24,6 +26,7 @@ impl From<&TrustRankConfig> for TrustRankParams {
             alpha: config.alpha,
             max_iterations: config.max_iterations,
             tolerance: config.tolerance,
+            max_projection_bytes: config.max_projection_bytes,
         }
     }
 }
--- a/nexusd/src/trust/engine.rs
+++ b/nexusd/src/trust/engine.rs
@@ -1,0 +1,49 @@
+use async_trait::async_trait;
+use nexus_common::types::DynError;
+use nexus_common::TrustRankConfig;
+
+/// Inputs for a seeded PageRank computation, decoupling [`TrustRankEngine`]
+/// from [`TrustRankConfig`] so implementations ignore config fields they don't
+/// use (e.g. `report_dir`).
+#[derive(Debug, Clone)]
+pub struct TrustRankParams {
+    /// Users trust originates from, each weighted equally (`v = 1/N`).
+    pub seed_ids: Vec<String>,
+    /// Teleport weight `alpha` in `trust = alpha*v + (1-alpha)*M*trust`.
+    pub alpha: f64,
+    /// Cap on power-iteration rounds.
+    pub max_iterations: u32,
+    /// Convergence tolerance.
+    pub tolerance: f64,
+}
+
+impl From<&TrustRankConfig> for TrustRankParams {
+    fn from(config: &TrustRankConfig) -> Self {
+        Self {
+            seed_ids: config.seed.iter().map(|id| id.to_string()).collect(),
+            alpha: config.alpha,
+            max_iterations: config.max_iterations,
+            tolerance: config.tolerance,
+        }
+    }
+}
+
+/// Summary of a completed trust-rank run.
+#[derive(Debug, Clone)]
+pub struct TrustRankStats {
+    /// Number of users a score was written for.
+    pub users_written: u64,
+    /// Power-iteration rounds actually run.
+    pub ran_iterations: u64,
+    /// Whether the computation converged within `max_iterations`.
+    pub did_converge: bool,
+}
+
+/// Computes seeded (personalized) PageRank trust, persisting each user's
+/// score as a side effect and returning only run stats.
+#[async_trait]
+pub trait TrustRankEngine: Send + Sync {
+    /// Runs the computation for `params` and returns run stats. Input
+    /// validation (seed set, empty graph) is the implementation's job.
+    async fn compute(&self, params: &TrustRankParams) -> Result<TrustRankStats, DynError>;
+}
--- a/nexusd/src/trust/export.rs
+++ b/nexusd/src/trust/export.rs
@@ -1,0 +1,176 @@
+use chrono::{DateTime, Utc};
+use nexus_common::db::fetch_all_rows_from_graph;
+use nexus_common::models::error::ModelResult;
+use nexus_common::types::DynError;
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+
+use super::neo4j::queries::{read_trust_scores, trust_report_user_details};
+
+/// Reads the `trust` scores written by the last recompute, highest first.
+/// Used by the CSV report and callers needing scores back after a
+/// [`super::TrustRankEngine::compute`].
+pub async fn read_scores() -> ModelResult<Vec<(String, f64)>> {
+    let rows = fetch_all_rows_from_graph(read_trust_scores()).await?;
+    let mut scores = Vec::with_capacity(rows.len());
+    for row in rows {
+        let user_id: String = row.get("user_id")?;
+        let score: f64 = row.get("score")?;
+        scores.push((user_id, score));
+    }
+    Ok(scores)
+}
+
+#[derive(Default, Clone)]
+struct UserReportRow {
+    name: String,
+    following: i64,
+    followers: i64,
+    friends: i64,
+    posts: i64,
+    replies: i64,
+    collections: i64,
+    bookmarks: i64,
+    tagged: i64,
+    tags: i64,
+    unique_tags: i64,
+}
+
+/// Writes a CSV report joining trust scores with user details and counts, for
+/// investigating who ranked where and why. Read straight from the graph (not
+/// Redis), so it reflects current state even where the cache is stale or empty.
+///
+/// Columns: id, name, score, followers, following, friends, posts, replies,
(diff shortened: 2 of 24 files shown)


aintnostressin:
@greptile review

aintnostressin:
@greptile review

nexusd/src/trust/neo4j/queries.rs
```diff
@@ -0,0 +1,162 @@
...
+/// Drops trust projections left by crashed or shutdown-cancelled runs (see
+/// scheduler.rs): entries under [`TRUST_GRAPH_PREFIX`] older than six hours.
+/// The 6h gate only sweeps a live run if a single PageRank exceeds 6h, at
+/// which point scheduler overlap — not the sweep — is the real failure.
+pub fn drop_stale_trust_graphs() -> Query {
+    Query::new(
+        "drop_stale_trust_graphs",
+        "CALL gds.graph.list() YIELD graphName, creationTime
+         WHERE graphName STARTS WITH $prefix AND creationTime < datetime() - duration({hours: 6})
+         CALL gds.graph.drop(graphName, false) YIELD graphName AS dropped
```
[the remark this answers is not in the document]
aintnostressin: pagerank default max itearation count is 200. Given the current prod graph size computation takes seconds-to-minutes.

aintnostressin:
@greptile review again

nexusd/src/trust/engine.rs
```diff
@@ -0,0 +1,49 @@
...
+    /// Power-iteration rounds actually run.
+    pub ran_iterations: u64,
+    /// Whether the computation converged within `max_iterations`.
+    pub did_converge: bool,
+}
+
+/// Computes seeded (personalized) PageRank trust, persisting each user's
+/// score as a side effect and returning only run stats.
+#[async_trait]
+pub trait TrustRankEngine: Send + Sync {
```
SHAcollision, line 48: I think we should drop this trait, or test it different. Only `GdsNeo4j` implement it and the test call the concrete `GdsNeo4j.compute()` directly, so the `Box<dyn TrustRankEngine>` never really used as polymorphic. So maybe better hold `GdsNeo4j` direct, OR add a mock engine test for the report part (then the trait make sense). The Params/Config split I would keep, that one is fine.

SHAcollision (commented):
Overall the trust part looks correct to me: seeded `sourceNodes`, `dampingFactor = 1 - alpha`, `alpha 0.35` like the doc, and the empty-seed case is guarded 2 times so it can not fall back to global pagerank. Nice.

Good stuff worth saying: empty-seed double guard, token-checked unlock, GDS jar with checksum + procedure allowlist in docker, csv formula-injection mitigation, a real GDS integration test, and the sybil script is exactly the canary we wanted from the design review. Few smaller things in the inline comments.

nexusd/src/trust/neo4j/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+        if matched < configured {
+            warn!(
+                matched,
+                configured, "Some configured trust seeds do not exist in the graph"
+            );
+        }
+
+        // Per-run unique name so a concurrent run can't clobber this
+        // run's in-flight projection.
+        let graph_name = format!(
```
SHAcollision, line 83: Just to note the lock TTL thing from #995 apply here too, is the same root, not a new bug. At current graph size the run is sub-second so almost no risk, and the fix go in #995 lock not here. (small note: a run would need to pass >2h to also race the projection cleanup, because projections have unique name + the sweep is 2h gated, so the real risk is the shared `trust` write, not the projection.)

nexusd/src/trust/neo4j/mod.rs
```diff
@@ -0,0 +1,131 @@
...
+            let dropped: Vec<String> = row.get("dropped").unwrap_or_default();
+            if !dropped.is_empty() {
+                warn!(
+                    ?dropped,
+                    "Dropped stale trust graph projections left by crashed runs"
+                );
+            }
+        }
+
+        let projected = fetch_row_from_graph(project_trust_graph(&graph_name)).await?;
```
SHAcollision, line 146: I am not sure I get this right, but the trust doc say dangling nodes (out-degree 0, follow nobody) must redistribute their mass by the teleport vector. The integration test check a node that is unreachable (`c` -> 0 trust, good) but not a node that IS reachable but follow nobody (like `b`, it have trust but follow nobody, where the mass go?). With `L1Norm` at the end we can not see a leak, and the sybil test pass, so risk low, but is in our own doc as must-have so maybe we confirm GDS do this, or we say explicit we skip it.

nexusd/src/trust/export.rs
```diff
@@ -0,0 +1,176 @@
...
+    scores: &[(String, f64)],
+) -> Result<PathBuf, DynError> {
+    tokio::fs::create_dir_all(dir).await?;
+    let path = dir.join(report_file_name(Utc::now()));
+    write_csv(&path, scores).await?;
+    Ok(path)
+}
+
+/// Second precision is enough: scheduled runs are minutes apart at their densest.
+fn report_file_name(now: DateTime<Utc>) -> String {
```
SHAcollision, line 127: The name is only seconds, and cron allow every second, so 2 runs in the same second overwrite the same csv. For trust is not real (run take more than 1s and the lock serialize) so is low, but the fix is 1 line (add millis or a counter) so I think we should just do it.

nexusd/src/trust/neo4j/queries.rs
```diff
@@ -0,0 +1,176 @@
...
+    )
+}
+
+/// Batched name + counts for a list of user ids, read straight from the graph
+/// (no Redis) so the report's numbers reflect current state, not a stale cache.
+///
+/// `OPTIONAL MATCH` keeps one row per requested id even if a user is missing;
+/// each `COUNT {}` subquery mirrors `nexus_common`'s `user_counts` and stays
+/// O(1) per field, as no row-multiplying match precedes them.
+pub fn trust_report_user_details(user_ids: &[String]) -> Query {
```
SHAcollision, line 142: minor: this copy the `user_counts` logic. 2 copies will drift (the collection/bookmark filters). Is report only so not urgent, but maybe share a helper.

nexusd/src/trust/neo4j/queries.rs
```diff
@@ -0,0 +1,176 @@
...
+    Query::new(
+        "drop_trust_graph",
+        "CALL gds.graph.drop($graph_name, false) YIELD graphName RETURN graphName",
+    )
+    .param("graph_name", graph_name.to_string())
+}
+
+/// Age gate for the stale-projection sweep, in hours. Set to 2× the run-lock
+/// TTL (`LOCK_TTL_SECS`, 1h) so the sweep can never race a lock-holding run.
+const STALE_GRAPH_SWEEP_HOURS: i64 = 2;
```
SHAcollision: minor: the 2h is only correct because is 2x the 1h lock TTL, but coupled by a comment across crates. Maybe better derive it from the TTL constant instead of magic 2.

nexusd/src/trust/neo4j/queries.rs
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/// Runs personalized PageRank over the projected follow graph in GDS **write
+/// mode**, storing each score on its `:User` node as the `trust` property.
+/// Teleports uniformly across `seed_ids` (`v = 1/N`); GDS 2.13 (the only
+/// version on Neo4j 5.26) has no weighted `sourceNodes`.
+///
+/// `scaler: 'L1Norm'` writes scores as an L1-normalized distribution (summing
+/// to 1) rather than raw rank values. Write mode overwrites `trust` on every
+/// projected `:User` each run, so scores never go stale.
```
SHAcollision: doc nit: users created between the projection and next run are stale until next batch. Is normal for a batch job, just maybe soften the comment.

nexusd/src/trust/engine.rs
aintnostressin: Good catch. I have added a mock impl and two unit tests.
[f41bd32](https://github.com/pubky/pubky-nexus/pull/984/commits/f41bd325955dc174aa3187137a91f1147a5c5dc6)

nexusd/src/trust/neo4j/mod.rs
aintnostressin: Great finding! Yes, dangling mass is dropped not redistributed.
I've added test for checking that - it doesn't sum up to 1 without L1Norm.

aintnostressin: We can redistribute this trust by creating follow edges in projected graph from dangling nodes to:
1. seed nodes
2. all nodes

I think we can think about 1 in the future but for now let's leave it as it is.

nexusd/src/trust/export.rs
aintnostressin: Done in [1eb4deb](https://github.com/pubky/pubky-nexus/pull/984/commits/1eb4deb1ab0d8664e8501dc80ce0d4929ac7ffd6)

nexusd/src/trust/neo4j/queries.rs
aintnostressin: done [8e5b31a](https://github.com/pubky/pubky-nexus/pull/984/commits/8e5b31a7d2591a819d9e9424c4450ac428f5fe63)

nexusd/src/trust/neo4j/queries.rs
aintnostressin: decoupled [42c9428](https://github.com/pubky/pubky-nexus/pull/984/commits/42c94283f035498e97bc94d044059e942d4cf3fe)

nexusd/src/trust/neo4j/queries.rs
aintnostressin: done: [f6ecb5a](https://github.com/pubky/pubky-nexus/pull/984/commits/f6ecb5ae128a9e52fd5642cfe35f7cf3a969f1c2)

tipogi (approved):
ready to merge!

PR #995, opened 2026-07-15
feat: Add a job-agnostic scheduling engine to nexusd

aintnostressin:
- `Job` trait + `run_job` scheduler loop (overrun-skip, MAX_SLEEP chunking, shutdown-responsive) backed by a `JobSpec` registry.
- `job_cron` gate + `JobConfig` (`[jobs.<name>]`, cron-only).
- `nexusd jobs run <name>` / `jobs list` CLI; the default daemon also drives scheduled jobs alongside webapi + watcher.
- Unit tests for the scheduler and `job_cron` validation.
- TrustRank job from #984

### DIFF
--- a/nexus-common/src/config/job.rs
+++ b/nexus-common/src/config/job.rs
@@ -1,0 +1,10 @@
+use serde::{Deserialize, Serialize};
+
+/// Scheduling config for a cron job, keyed under `[jobs.<name>]`.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
+#[serde(default, deny_unknown_fields)]
+pub struct JobConfig {
+    /// 7-field cron (`sec min hour dom month dow [year]`, seconds-first).
+    /// `None` leaves the job unscheduled.
+    pub cron: Option<String>,
+}
--- a/nexus-common/src/config/job.rs
+++ b/nexus-common/src/config/job.rs
@@ -4,7 +4,7 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
 #[serde(default, deny_unknown_fields)]
 pub struct JobConfig {
-    /// 7-field cron (`sec min hour dom month dow [year]`, seconds-first).
+    /// 6- or 7-field cron (`sec min hour dom month dow [year]`, seconds-first).
     /// `None` leaves the job unscheduled.
     pub cron: Option<String>,
 }
--- a/nexusd/src/jobs/lock.rs
+++ b/nexusd/src/jobs/lock.rs
@@ -87,13 +87,49 @@ fn key(job: &str) -> String {
     format!("lock:job:{job}")
 }
 
-/// RAII lock release. Call `release()` on the happy path to await unlock; on a
-/// panic the explicit call is skipped but Drop still frees the lock so it
-/// doesn't stay stuck for its TTL.
+/// The outcome of [`acquire`].
+pub(super) enum Acquired {
+    /// Taken; release through the guard.
+    Taken(LockGuard),
+    /// Another run holds it. Nothing to release.
+    Held,
+    /// The backend failed. Anything we might have taken is already released.
+    Failed(LockError),
+}
+
+/// Takes `job`'s run slot, arming the guard *before* the acquire so a lost SET
+/// reply (or a cancel mid-acquire) still releases; the compare-and-delete no-ops
+/// if our token was never stored.
 ///
-/// Drop releases fire-and-forget via `tokio::spawn` (so it must be dropped
-/// inside a runtime context), falling back to the TTL if the spawn is dropped
-/// during shutdown.
+/// One `job` feeds both the acquire and the guard, so a run can't release a key
+/// it never acquired. On backend error the guard is released inline rather than
+/// left to Drop: callers may return straight into process exit, where a
+/// Drop-spawned unlock never gets polled.
+pub(super) async fn acquire(job: &'static str, lock: &Arc<dyn RunLock>) -> Acquired {
+    let token = lock.new_token();
+    let guard = LockGuard::new(job, token.clone(), Arc::clone(lock));
+
+    match lock.acquire(job, &token).await {
+        Ok(true) => Acquired::Taken(guard),
+        Ok(false) => {
+            guard.disarm();
+            Acquired::Held
+        }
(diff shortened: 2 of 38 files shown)


aintnostressin:
@greptile review

nexusd/src/jobs/lock.rs
```diff
@@ -0,0 +1,153 @@
+use async_trait::async_trait;
+use nexus_common::db::{release_lock, try_acquire_lock, RedisResult};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+/// TTL on a job's run lock, and the crash backstop. Must exceed the longest
+/// expected run, or a run outliving its lease could overlap another.
+const LOCK_TTL_SECS: u64 = 3600;
```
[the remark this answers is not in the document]
aintnostressin: For now it's good enough. We need to ensure jobs don''t exceed this time .This is just a clean up mechanism in case lock is not released.

nexusd/src/jobs/scheduler.rs
```diff
@@ -0,0 +1,741 @@
...
+            tokio::select! {
+                biased;
+
+                // Dropping the future leaves the lock to its TTL, but we're exiting anyway.
+                _ = wait_for_shutdown(&mut shutdown_rx) => {
+                    info!(job = name, "Shutdown received during run; abandoning it");
+                    return;
+                }
+                () = self.run_locked(name, job) => {}
+            }
```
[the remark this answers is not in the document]
aintnostressin: fixed: [be8c707](https://github.com/pubky/pubky-nexus/pull/995/commits/be8c70771ce9dd206d2eb76651d44db35e07c6df)

nexusd/src/jobs/scheduler.rs
```diff
@@ -0,0 +1,741 @@
...
+                return;
+            }
+        }
+    }
+}
+
+/// Parses `cron_expr` and checks it has at least one upcoming fire time,
+/// returning the [`Schedule`] or an error. Used to fail fast when building a job.
+pub fn validate_cron(cron_expr: &str) -> Result<Schedule, CronParseError> {
+    let schedule = Schedule::from_str(cron_expr).map_err(|e| CronParseError {
```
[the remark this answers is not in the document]
aintnostressin: This is false positive. I've added a unit test in [e034088](https://github.com/pubky/pubky-nexus/pull/995/commits/e0340889ae5e7c2b56fb4be23489a940a9644827)

aintnostressin:
@greptile review again

SHAcollision (commented):
Overall the scheduler looks good and well tested, nice work.

One thing about landing it: I think we should not merge this alone to main. Right now the registry is empty, only #984 give it a real job, and the lock issue below only really happen when a job run. So maybe better we keep #984 on top of feat/jobs and land both together, or just fold #984 into feat/jobs and make one PR. I don't like to leave the framework alone in main with nobody using it.

nexusd/src/jobs/lock.rs
```diff
@@ -0,0 +1,245 @@
...
+use nexus_common::db::{release_lock, try_acquire_lock, RedisError};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use super::error::{LockError, LockResult};
+
+/// TTL on a job's run lock, and the crash backstop. Must exceed the longest
+/// expected run, or a run outliving its lease could overlap another.
+const LOCK_TTL_SECS: u64 = 3600;
```
SHAcollision: I think we have one real thing here. The "no overlap" only work until the 1h TTL. `run()` have no timeout and we don't renew the lease, so if a job run more than 1h it lose the lock and another process can start the same job at the same time. To happen you need 2 process AND a >1h run so is not so common, but we promise no overlap so better we fix.

Maybe we don't need the full lease renew/fencing. Simpler idea: add a timeout in the `select!` of `run_locked` so the job can not run longer than the lease (if pass, we abandon + log), and make lease+timeout one config value. Like this the rule is always true and no extra renew loop. We can add a test on the boundary. Only if some job really need >1h we do the renew thing later.

nexus-common/src/config/job.rs
```diff
@@ -0,0 +1,10 @@
+use serde::{Deserialize, Serialize};
+
+/// Scheduling config for a cron job, keyed under `[jobs.<name>]`.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Default)]
+#[serde(default, deny_unknown_fields)]
+pub struct JobConfig {
+    /// 7-field cron (`sec min hour dom month dow [year]`, seconds-first).
```
SHAcollision: nit: comment say "7-field" but we accept 6 and 7. Maybe reword.

nexusd/src/jobs/lock.rs
aintnostressin: done [49ffe59](https://github.com/pubky/pubky-nexus/pull/995/commits/49ffe593c6fcfa6325e9410475ba15de747f4b55)

nexus-common/src/config/job.rs
aintnostressin: done: [62a2a32](https://github.com/pubky/pubky-nexus/pull/995/commits/62a2a322ef8383cfda1b6141f54cfde922d7a826)

nexusd/src/jobs/lock.rs
```diff
@@ -188,7 +189,7 @@ impl LockMetrics {
         ];
         self.operations_counter.add(1, &tags);
         self.duration_histogram
-            .record(duration.as_secs_f64(), &tags);
+            .record(duration.as_secs_f64() * 1000.0, &tags);
```
tipogi: there is a helper function in `common/src/db/graph/instrumented.rs`, 
```rust
fn ms(d: Duration) -> f64 {
    d.as_secs_f64() * 1000.0
}
```

aintnostressin: done in [21d5c26](https://github.com/pubky/pubky-nexus/pull/995/commits/21d5c2613c75d50e53aa3d8ff50239bce6bd27cf)

tipogi (approved):
ready!

SHAcollision (approved):
LGTM 

Heads up `trust.recompute.max_iterations_reached` is a permanent no-op in production.

PR #997, opened 2026-07-17
fix: prevent HS reindexing, only allow cursor to increment

ok300:
This PR prevents homeserver event reindexing by enforcing monotonic cursors:

- The primary HS cursor may never rewind. Non-empty batches must advance past the requested cursor, are validated before handlers run, and are checkpointed only after successful processing.
- External HS per-user cursors must advance past both the stored cursor and earlier events in the same batch. Violations abort that HS run and trigger the existing configurable per-HS exponential backoff.
- Primary HS cursors are represented numerically while remaining compatible with numeric values previously stored as strings.

Records the following metrics:

| Metric | Meaning |
|---|---|
| `watcher.primary_hs.cursor.invalid` | Primary HS cursor was unparseable or would rewind |
| `watcher.primary_hs.cursor.stalled` | A non-empty primary HS batch could not advance its checkpoint |
| `watcher.external_hs.cursor.out_of_order` | External HS returned a replayed or out-of-order per-user cursor |

Supersedes #662 by reimplementing it on the latest `main`.

---

### Pre-submission Checklist

### DIFF
--- a/nexus-common/src/db/kv/error.rs
+++ b/nexus-common/src/db/kv/error.rs
@@ -38,6 +38,18 @@ impl RedisError {
     }
 }
 
+/// Shared monotonicity rule for Primary HS global cursors and External HS user cursors:
+/// a cursor may hold or advance, but never rewind. Equal is allowed so an
+/// idempotent re-persist of a boundary event is not treated as an error.
+pub fn ensure_cursor_not_backwards(new_cursor: u64, stored_cursor: u64) -> RedisResult<()> {
+    if new_cursor < stored_cursor {
+        return Err(RedisError::InvalidInput(
+            "Cursor cannot move backwards".into(),
+        ));
+    }
+    Ok(())
+}
+
 impl From<redis::RedisError> for RedisError {
     fn from(e: redis::RedisError) -> Self {
         if e.is_connection_refusal() || e.is_timeout() || e.is_io_error() {
--- a/nexus-common/src/db/kv/index/sorted_sets.rs
+++ b/nexus-common/src/db/kv/index/sorted_sets.rs
@@ -111,7 +111,12 @@ pub async fn put(
 ///
 /// `ZADD NX` makes the check and the write one atomic operation, so an existing
 /// member keeps its score whatever concurrent writers are doing.
-pub async fn put_if_absent(prefix: &str, key: &str, score: f64, member: &str) -> RedisResult<()> {
+pub async fn add_member_if_absent(
+    prefix: &str,
+    key: &str,
+    score: f64,
+    member: &str,
+) -> RedisResult<()> {
     let index_key = format!("{prefix}:{key}");
     let mut redis_conn = get_redis_conn().await?;
     let _: () = redis::cmd("ZADD")
--- a/nexus-common/src/db/kv/index/sorted_sets.rs
+++ b/nexus-common/src/db/kv/index/sorted_sets.rs
@@ -107,6 +107,23 @@ pub async fn put(
     Ok(())
 }
 
+/// Seeds `member` with `score` only if it is not already in the sorted set.
+///
+/// `ZADD NX` makes the check and the write one atomic operation, so an existing
+/// member keeps its score whatever concurrent writers are doing.
+pub async fn put_if_absent(prefix: &str, key: &str, score: f64, member: &str) -> RedisResult<()> {
+    let index_key = format!("{prefix}:{key}");
+    let mut redis_conn = get_redis_conn().await?;
+    let _: () = redis::cmd("ZADD")
+        .arg(index_key)
+        .arg("NX")
+        .arg(score)
+        .arg(member)
+        .query_async(&mut redis_conn)
+        .await?;
+    Ok(())
+}
+
 /// Updates the score of a member in a Redis sorted set.
 ///
 /// This function modifies the score of a member in the specified Redis sorted set by incrementing or decrementing it
--- a/nexus-common/src/db/kv/traits.rs
+++ b/nexus-common/src/db/kv/traits.rs
@@ -602,7 +602,7 @@ pub trait RedisOps: Serialize + DeserializeOwned + Send + Sync {
     /// Atomic (`ZADD NX`), unlike a `check_sorted_set_member` read followed by
     /// [`Self::put_index_sorted_set`], which would clobber a member written
     /// between the two calls.
-    async fn put_index_sorted_set_if_absent(
+    async fn add_index_sorted_set_if_absent(
         key_parts: &[&str],
         score: f64,
         member: &str,
@@ -610,7 +610,7 @@ pub trait RedisOps: Serialize + DeserializeOwned + Send + Sync {
     ) -> RedisResult<()> {
         let prefix = prefix.unwrap_or(SORTED_PREFIX);
         let key = key_parts.join(":");
-        sorted_sets::put_if_absent(prefix, &key, score, member).await
+        sorted_sets::add_member_if_absent(prefix, &key, score, member).await
     }
 
     /// Updates the score of a member in a Redis sorted set.
--- a/nexus-common/src/db/kv/traits.rs
+++ b/nexus-common/src/db/kv/traits.rs
@@ -597,6 +597,22 @@ pub trait RedisOps: Serialize + DeserializeOwned + Send + Sync {
         sorted_sets::put(prefix, &key, elements, expiration).await
     }
 
+    /// Seeds a member of a Redis sorted set, leaving an already-present one untouched.
+    ///
+    /// Atomic (`ZADD NX`), unlike a `check_sorted_set_member` read followed by
+    /// [`Self::put_index_sorted_set`], which would clobber a member written
+    /// between the two calls.
+    async fn put_index_sorted_set_if_absent(
+        key_parts: &[&str],
+        score: f64,
+        member: &str,
+        prefix: Option<&str>,
+    ) -> RedisResult<()> {
+        let prefix = prefix.unwrap_or(SORTED_PREFIX);
+        let key = key_parts.join(":");
+        sorted_sets::put_if_absent(prefix, &key, score, member).await
+    }
+
     /// Updates the score of a member in a Redis sorted set.
     ///
     /// This method updates the score associated with a specific member in a Redis sorted set
--- a/nexus-common/src/models/user/cursor.rs
+++ b/nexus-common/src/models/user/cursor.rs
@@ -58,8 +58,14 @@ impl UserHsCursor {
             .collect()
     }
 
-    /// Persists a single user's event cursor for `hs_id` to its `USER_HS_CURSOR` sorted set.
+    /// Persists a single user's event cursor for `hs_id` without moving it backward.
     pub async fn write(user_id: &str, hs_id: &str, cursor: u64) -> RedisResult<()> {
+        let stored_cursor = Self::read(&[user_id], hs_id)
+            .await?
+            .into_iter()
+            .next()
+            .unwrap_or_default();
+        let cursor = cursor.max(stored_cursor);
         let key = user_hs_cursor_key(user_id);
         Self::put_index_sorted_set(&key, &[(cursor as f64, hs_id)], None, None).await
     }
@@ -72,7 +78,7 @@ mod tests {
     use crate::{types::DynError, utils::test_utils::random_pubky_id, StackConfig, StackManager};
 
     /// `write` persists a per-user cursor that `read` reads back, missing entries
-    /// default to 0, and re-writing overwrites the value.
+    /// default to 0, and lower re-writes do not rewind the value.
     #[tokio_shared_rt::test(shared)]
     async fn test_hs_cursor_read_write_roundtrip() -> Result<(), DynError> {
         StackManager::setup(&StackConfig::default()).await?;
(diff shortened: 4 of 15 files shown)


ok300:
@greptile please review again

ok300:
@greptile please review again

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -235,7 +235,7 @@ impl HsEventProcessor {
 
             if let Some(cursor) = line.strip_prefix("cursor: ") {
                 info!("Received cursor for the next request: {cursor}");
-                match Homeserver::try_from_cursor(self.homeserver.id.clone(), cursor) {
+                match Homeserver::try_from_cursor(self.homeserver.id.clone(), cursor).await {
                     Ok(hs) => hs.put_to_index().await?,
                     Err(e) => warn!("{e}"),
                 }
```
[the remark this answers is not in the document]
ok300: If a HS returns a cursor below the user's floor ( b1b9854af29bcc829626b91dfef14eab63e9e02f ), then "warn and don't advance" is correct.

For invalid cursors, a metric was added in 877fdcd6e3203f25dedbfaab0c9ab50c4333c3d0

ok300:
@greptile please review again

ok300:
@greptile please review again

ok300:
@greptile please review again

nexus-common/src/models/user/cursor.rs
```diff
@@ -58,11 +58,40 @@ impl UserHsCursor {
...
+        let key = user_hs_cursor_key(user_id);
+        // Check the raw score, not `read` (which maps a missing entry to 0 and
+        // so can't tell "absent" from "present == 0").
+        if Self::check_sorted_set_member(None, &key, &[hs_id])
+            .await?
+            .is_some()
+        {
+            return Ok(());
+        }
+        Self::put_index_sorted_set(&key, &[(0.0, hs_id)], None, None).await
```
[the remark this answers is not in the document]
ok300: There is only a single-writer for the cursor init here, as documented in b1e1535381335f35b145de4ef78176a7e682d815 .

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -235,9 +245,14 @@ impl HsEventProcessor {
 
             if let Some(cursor) = line.strip_prefix("cursor: ") {
                 info!("Received cursor for the next request: {cursor}");
-                match Homeserver::try_from_cursor(self.homeserver.id.clone(), cursor) {
+                match Homeserver::try_from_cursor(self.homeserver.id.clone(), cursor).await {
```
SHAcollision: Not sure how important this is, but it looks like this accepts a cursor equal to the stored cursor. Equality is only valid when this response had no event lines. If a response requested at `10` contains events but ends with `cursor: 10`, this code processes the events, leaves the checkpoint at `10`, and reprocesses the same batch forever. Maybe primary cursor validation should be batch-aware ?

SHAcollision (approved):
Looks good. It's a shame numerically monotonically increasing cursors are going to disapear from the homeserver soon so we might need to redo this.

nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -310,6 +331,25 @@ impl KeyBasedEventProcessor {
             }
 
             let cursor_id = stream_event.cursor.id();
+            // Fetches are cursor-exclusive, so a correct HS only returns events
+            // above the floor; anything at or below it is a replay. Seeding the
+            // floor from `persisted_cursor` matters because `latest_cursor`
+            // resets each batch — the in-batch check alone would miss a replay
+            // of an already-indexed event.
+            let ordering_floor = latest_cursor.unwrap_or(persisted_cursor);
+            if cursor_id <= ordering_floor {
```
tipogi: Returning here stops the entire user batch, so one stale event can repeatedly block valid newer events behind it. For example, with persisted cursor `12`, a response `[9, 13]` stops at `9` and never processes `13`; the next poll may receive the same response. Since these malformed events are already treated as non-retryable, could we increment the metric, log the skipped cursor, and continue instead? This preserves forward progress while still preventing the invalid event from reaching its handler or rewinding the cursor.

nexus-watcher/src/service/indexer/homeserver.rs
tipogi: Could we extract and validate the `cursor` before processing the event lines, then persist it after the batch succeeds (246 line)? Since the `cursor` is normally last, malformed/backward cursors and Redis validation failures (`ModelError::KvOperationFailed`) are currently discovered only after handlers have run. 
The Err branch (line 250) also treats Redis infrastructure failures as invalid HS input and returns success without advancing the checkpoint, causing the batch to be reprocessed. Invalid HS cursors can still stop the batch and increment the metric, while Redis failures should propagate so the run is recorded as failed and retried on the next scheduled execution.

tipogi (changes requested):

nexus-watcher/src/service/indexer/homeserver.rs
ok300: > Maybe primary cursor validation should be batch-aware ?

Done: a2199f05c4a8238844c36cde10845ea93d95b8be

nexus-watcher/src/service/indexer/homeserver.rs
ok300: Done: a2199f05c4a8238844c36cde10845ea93d95b8be

nexus-watcher/src/service/indexer/key_based.rs
ok300: A bit tricky, because the two cases of "external HS returned events out of order" and "external HS started with a cursor below the given floor" were treated as one.

A solution is added in 3bb4009d0d0341876b40d4b86c558c6d625b1fd2

The single `ordering_floor` check became two:

- `cursor_id <= persisted_cursor` → replay of already-indexed data: metric + warn! + continue. This is the scenario you mentioned; `[9, 13]` behind cursor `12` now indexes `13` and checkpoints there instead of wedging.
- `cursor_id <= latest_cursor` (above the stored cursor) → unsorted batch: unchanged abort with `EventCursorOutOfOrder`.

ok300: Reverted the fix for now.

> For example, with persisted cursor `12`, a response `[9, 13]` stops at `9` and never processes `13`

IMO this broken or malicious behavior from an external HS. If the request provides the `cursor` arg as `12`, a response with `[9, 13]` should be seen as invalid.

If we don't enforce that and allow more "flexible" handling (A) :
> "increment the metric, log the skipped cursor, and continue"

this risks that we lose events.

For example:
- what do we do for `[9, 13, 12]` with persisted cursor `11`?
  - if we persist `13`, then when we get `12` we will skip it (as per the logic in A)
- what do we do if
  - on first call, we get `[9, 15]` with persisted cursor `12` -> then persist `15`
  - then on 2nd call get `[10, 12, 16]` with persisted cursor `15`
  - we would skip `12`, which would be a lost event

IMO cleanly enforcing the rules that
- events should be incremental
- responses should respect the request's `cursor` arg

helps us identify broken / malicious HSs and avoids all the above complications.

ok300: As discussed: add simple (hardcoded) backoff, if a HS returns cursors out-of-order.

ok300: Done: e8a03d05d24149353206d2c017413fa75f8069db

ok300:
@greptile please review again

tipogi:
NOTE: @SpontaneousOverthrow, these OpenTelemetry counters are exported to Prometheus/Grafana. We’d like to configure alerts when any of the following metrics keeps increasing for the same hs_id across consecutive polling cycles:
- `watcher.primary_hs.cursor.stalled`
- `watcher.primary_hs.cursor.invalid`
- `watcher.external_hs.cursor.out_of_order`

Sustained increases indicate that the watcher cannot advance its primary or key-based external homeserver checkpoint and may require some investigation.

nexus-common/src/db/kv/index/sorted_sets.rs
```diff
@@ -107,6 +107,23 @@ pub async fn put(
     Ok(())
 }
 
+/// Seeds `member` with `score` only if it is not already in the sorted set.
+///
+/// `ZADD NX` makes the check and the write one atomic operation, so an existing
+/// member keeps its score whatever concurrent writers are doing.
+pub async fn put_if_absent(prefix: &str, key: &str, score: f64, member: &str) -> RedisResult<()> {
```
tipogi: I prefer `add_member_if_absent` or similar because `put` usually implies that an existing value may be replaced, while this function leaves existing members unchanged.

nexus-common/src/db/kv/error.rs
```diff
@@ -38,6 +38,18 @@ impl RedisError {
...
+/// idempotent re-persist of a boundary event is not treated as an error.
+pub fn ensure_cursor_not_backwards(new_cursor: u64, stored_cursor: u64) -> RedisResult<()> {
+    if new_cursor < stored_cursor {
+        return Err(RedisError::InvalidInput(
+            "Cursor cannot move backwards".into(),
+        ));
+    }
+    Ok(())
+}
+
```
tipogi, line 52: This looks like a cursor-domain validation error rather than a Redis error, since no Redis operation has failed here. Would it be clearer to move this validation into a dedicated cursor module and return `CursorError::WouldRewind { stored_cursor, new_cursor }` (or similar)? Since the validation is shared , we could place it in `nexus-common/src/models/homeserver/cursor.rs`, with the current `homeserver.rs` becoming `homeserver/mod.rs`. Maybe the refactoring, we could let for the next PR and now create the file in `nexus-common/src/models/cursor.rs`

nexus-common/src/db/kv/traits.rs
```diff
@@ -597,6 +597,22 @@ pub trait RedisOps: Serialize + DeserializeOwned + Send + Sync {
         sorted_sets::put(prefix, &key, elements, expiration).await
     }
 
+    /// Seeds a member of a Redis sorted set, leaving an already-present one untouched.
+    ///
+    /// Atomic (`ZADD NX`), unlike a `check_sorted_set_member` read followed by
+    /// [`Self::put_index_sorted_set`], which would clobber a member written
+    /// between the two calls.
+    async fn put_index_sorted_set_if_absent(
```
tipogi: What do you think, if we rename this method to `add_sorted_set_member_if_absent`

nexus-common/src/models/user/cursor.rs
```diff
@@ -58,11 +58,40 @@ impl UserHsCursor {
...
+    /// Seeded atomically (`ZADD NX`) rather than under [`Self::write`]'s
+    /// single-writer invariant, which does not reach here: the caller is the user
+    /// ingestor, running on the primary-HS and retry-processor tasks, and it
+    /// publishes the user's `HOSTED_BY` edge before seeding — so the concurrent
+    /// external-HS task can fetch, index and `write` a real cursor in between. A
+    /// read-then-write would clobber that with 0 and re-index the user's history.
+    pub async fn init(user_id: &str, hs_id: &str) -> RedisResult<()> {
+        let key = user_hs_cursor_key(user_id);
+        Self::put_index_sorted_set_if_absent(&key, 0.0, hs_id, None).await
+    }
```
tipogi, line 94: good one to assure that we do not overwrite the already persisted homeservers cursor

nexus-watcher/src/service/indexer/homeserver.rs
```diff
@@ -40,6 +66,57 @@ pub enum HsMapping {
...
+///
+/// The homeserver sends the cursor line last, but that is positional convention
+/// rather than something we can rely on from an untrusted peer: the split is
+/// order-independent, and if a batch carries several cursor lines the last one
+/// wins, being the position the homeserver ends the batch at.
+struct EventBatch<'a> {
+    event_lines: Vec<&'a str>,
+    /// Raw, still unparsed value of the batch's cursor line, if it carried one.
+    cursor: Option<&'a str>,
+}
```
tipogi, line 79: nice helper

tipogi (commented):
apart of the new helper added in the redis domain, that encapsulates cursor errors under `RedisError::InvalidInput` variant, looks good improvement and final one

nexus-common/src/db/kv/error.rs
ok300: This was intended as a data validation check, just before persisting to Redis.

That's why it's checked inside `UserHsCursor::write`: 

```rust
    pub async fn write(user_id: &str, hs_id: &str, cursor: u64) -> RedisResult<()> {
        // ... let stored_cursor = Self::read(&[user_id], hs_id) ...

        ensure_cursor_not_backwards(cursor, stored_cursor)?;

        // ... Self::put_index_sorted_set(&key, &[(cursor as f64, hs_id)], None, None).await
    }
```

Since that only does Redis operations (reads existing cursor, compares to new, validates) and returns a `RedisResult`, it made sense to throw a Redis error.

Since it validates the new value against the existing (Redis) one, IMO it makes sense to keep it as it is. If we extract it outside of `write`, then every caller who writes will have to also validate just before. From my view, it makes sense that validation is done as part of `write` rather than separately.

nexus-common/src/db/kv/traits.rs
ok300: Done: fb5f9d12eca69fc29187e2789e9629e9986548e5

nexus-common/src/db/kv/index/sorted_sets.rs
ok300: Good idea, done: fb5f9d12eca69fc29187e2789e9629e9986548e5

tipogi (approved):

PR #999, opened 2026-07-20
fix: bound concurrent media processing

aintnostressin:
Unbounded variant generation let a burst of static-file requests spawn one ImageMagick subprocess each, starving the host. Cap the concurrent processes with a semaphore, and cap the queue waiting on it so a burst parks a bounded number of clients instead of arbitrarily many.

The process limit is configured with [stack.media].max_concurrency in config.toml; it defaults to the CPU count (min 4) and rejects 0.

Requests that cannot get a slot degrade instead of failing: the avatar endpoint serves the original `main` image, while /static/files returns 503, since there the variant is explicit in the URL.

### DIFF
--- a/nexus-common/src/config/stack.rs
+++ b/nexus-common/src/config/stack.rs
@@ -5,6 +5,49 @@ use std::{fmt::Debug, path::PathBuf};
 use super::net::NetConfig;
 use super::{file::validate_and_expand_path, Level, LOG_LEVEL};
 
+/// Media processing concurrency configuration.
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq)]
+pub struct MediaConfig {
+    /// Maximum number of concurrent media subprocesses (ImageMagick/ffmpeg).
+    /// Defaults to the number of available parallelism (CPU cores), minimum 4.
+    #[serde(
+        default = "MediaConfig::default_max_concurrency",
+        deserialize_with = "deserialize_max_concurrency"
+    )]
+    pub max_concurrency: usize,
+}
+
+/// Rejects 0, which would otherwise shed every variant request forever.
+fn deserialize_max_concurrency<'de, D>(deserializer: D) -> Result<usize, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let max_concurrency = usize::deserialize(deserializer)?;
+    if max_concurrency == 0 {
+        return Err(serde::de::Error::custom(
+            "stack.media.max_concurrency must be greater than 0",
+        ));
+    }
+    Ok(max_concurrency)
+}
+
+impl MediaConfig {
+    fn default_max_concurrency() -> usize {
+        std::thread::available_parallelism()
+            .map(|n| n.get())
+            .unwrap_or(4)
+            .max(4)
+    }
+}
+
+impl Default for MediaConfig {
+    fn default() -> Self {
+        Self {
+            max_concurrency: Self::default_max_concurrency(),
+        }
+    }
+}
+
 fn deserialize_and_expand<'de, D>(deserializer: D) -> Result<PathBuf, D::Error>
 where
     D: Deserializer<'de>,
@@ -41,6 +84,8 @@ pub struct StackConfig {
     pub db: DatabaseConfig,
     #[serde(default)]
     pub net: NetConfig,
+    #[serde(default)]
+    pub media: MediaConfig,
 }
 
 /// Utility function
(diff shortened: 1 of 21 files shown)


aintnostressin:
@greptile review

nexus-common/src/media/processors/mod.rs
```diff
@@ -96,6 +102,8 @@ pub trait VariantProcessor {
             ));
         };
 
+        // Held only around the subprocess, not the path/option work above.
+        let _permit = gate.acquire().await?;
         Self::process(origin_file_path, output_path, &options).await?;
```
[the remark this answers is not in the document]
aintnostressin: fixed in [a878bab](https://github.com/pubky/pubky-nexus/pull/999/commits/a878bab3b085cf0a28e2b4405c768eea45773061)

nexus-common/src/config/stack.rs
```diff
@@ -5,6 +5,49 @@ use std::{fmt::Debug, path::PathBuf};
...
+where
+    D: Deserializer<'de>,
+{
+    let max_concurrency = usize::deserialize(deserializer)?;
+    if max_concurrency == 0 {
+        return Err(serde::de::Error::custom(
+            "stack.media.max_concurrency must be greater than 0",
+        ));
+    }
+    Ok(max_concurrency)
```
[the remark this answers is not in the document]
aintnostressin: Keep it as it is. It fails on startup so nothing bad will happen.
This may only happen if operator sets absurd config value.

aintnostressin:
@greptile review again

nexus-common/src/models/file/blob.rs
```diff
@@ -49,36 +47,15 @@ impl Blob {
                 file, variant,
             ))
         } else {
-            Self::put_variant(file, variant, file_path)
+            controller
+                .create_file_variant(file, variant, file_path)
                 .await
+                .map_err(Into::into)
                 .inspect_err(|e| {
                     tracing::error!("Creating variant failed for file: {file:?} with error: {e}")
```
tipogi: I’ll skip the `error!` in Blob::get_by_id when `e.is_at_capacity()`, so bursts don’t spam ERROR logs. The web layer already `warn!`s
```rust
.inspect_err(|e| {
        if !e.is_at_capacity() {
            tracing::error!("Creating variant failed for file: {file:?} with error: {e}")
        }
    })
```

nexus-webapi/src/routes/static/avatar.rs
```diff
@@ -58,19 +62,55 @@ pub async fn user_avatar_handler(
...
+    // Falls back to the untouched main image when the media gate is at capacity,
+    // so an avatar still renders instead of failing.
+    let controller = &app_state.variant_controller;
+    let variant = match Blob::get_by_id(
         &file_details,
         &FileVariant::Small,
         file_path.clone(),
+        controller,
+    )
+    .await
```
tipogi, line 77: This fallback may take up to 5 seconds because `Blob::get_by_id` waits for a media permit before returning `AtCapacity`. Since avatars can degrade gracefully, could we propagate a fail-fast option to the gate and immediately serve `main` when no permit is available?

For example, `MediaGate` could expose:

```rust
pub fn try_acquire(&self) -> Result<OwnedSemaphorePermit, MediaProcessorError> {
    Arc::clone(&self.semaphore)
        .try_acquire_owned()
        .map_err(|_| MediaProcessorError::AtCapacity)
}
```

The avatar variant-generation path would use this instead of the queued `acquire()`.

nexus-common/src/media/processors/mod.rs
```diff
@@ -96,8 +105,138 @@ pub trait VariantProcessor {
...
+        // into the task so it is released when the child exits, not when the caller's future
+        // is dropped: a cancelled request must not hand its permit on while its child runs.
+        let permit = gate.acquire().await?;
+        let content_type = options.content_type();
+        let origin_file_path = origin_file_path.to_string();
+        let output_path = output_path.to_string();
+
+        tokio::spawn(async move {
+            let _permit = permit;
+            Self::process(&origin_file_path, &output_path, &options).await
```
tipogi: Non-blocking but a hung `identify` or `convert` process can now retain its permit indefinitely. The underlying subprocess hang risk is pre-existing, and keeping the permit while the child runs is correct, but without a deadline enough hung processes could exhaust the media gate until restart.

Could we track a shared subprocess runner that applies a configurable timeout, explicitly kills and reaps the child on expiry, and only then returns a timeout error? `kill_on_drop(true)` would also be useful as a safety net.

nexus-webapi/src/routes/static/avatar.rs
aintnostressin: I've added fail-fast in [df1ef7c](https://github.com/pubky/pubky-nexus/pull/999/commits/df1ef7c9bbbff61f8cdbafe51e6421c412c3f787)

nexus-common/src/models/file/blob.rs
aintnostressin: Updated logging [1e80d9c](https://github.com/pubky/pubky-nexus/pull/999/commits/1e80d9ce1d475d3edc7c87adc04f8d327b263f2a)

nexus-common/src/media/processors/mod.rs
aintnostressin: subprocess struct and command timeout added in [54fc301](https://github.com/pubky/pubky-nexus/pull/999/commits/54fc301869de81d1c2dbbe288deeff6ec7b6258c)

aintnostressin: Now if we kill a running command it may leave a broken variant on disk. Since we added killing we should also fix this.

tipogi (approved):
LGTM

PR #1006, opened 2026-07-27
chore: gate `test_utils` module behind `test-utils` Cargo feature

aintnostressin:
Add a `test-utils` feature to `nexus-common` and gate the `test_utils` module behind `#[cfg(feature = "test-utils")]` so that test-only code is excluded from production builds.


aintnostressin:
@greptile review

tipogi (approved):

PR #1009, opened 2026-07-31
feat: Replace [DELETED] sentinel with boolean deleted flag

aintnostressin:
Tombstoned users get `deleted = true` instead of having their name overwritten, so a live user named "[DELETED]" is no longer treated as deleted.

 New-code tombstones leave `name` empty, not `'[DELETED]'`. Rolling back to code that filters by the sentinel name makes them appear live, with an empty profile.

### DIFF
--- a/nexus-common/src/models/user/details.rs
+++ b/nexus-common/src/models/user/details.rs
@@ -208,13 +208,11 @@ mod tests {
         );
     }
 
-    /// Deserializing a UserDetails from a BoltNode with `deleted: true` should
-    /// preserve the flag — this is the shape after the one-shot migration has
-    /// marked a sentinel tombstone.
+    /// A present `deleted: true` deserializes as true — the other branch of the custom deserializer.
     #[test]
     fn deserialize_from_node_with_deleted_true() {
         let mut props = BoltMap::new();
-        props.put(BoltString::from("name"), BoltType::from("[DELETED]"));
+        props.put(BoltString::from("name"), BoltType::from(""));
         props.put(
             BoltString::from("id"),
             BoltType::from("uo7jgkykft4885n8cruizwy6khw71mnu5pq3ay9i8pw1ymcn85ko"),
--- a/nexus-common/src/models/user/details.rs
+++ b/nexus-common/src/models/user/details.rs
@@ -45,6 +45,8 @@ pub struct UserDetails {
     pub status: Option<String>,
     pub image: Option<String>,
     pub indexed_at: i64,
+    #[serde(deserialize_with = "deserialize_user_deleted", default)]
+    pub deleted: bool,
 }
 
 fn deserialize_user_links<'de, D>(
@@ -82,6 +84,19 @@ where
     }
 }
 
+fn deserialize_user_deleted<'de, D>(deserializer: D) -> Result<bool, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    // Neo4j drops null/missing properties, so pre-migration nodes will lack the
+    // `deleted` field entirely. Redis JSON cache entries written before the
+    // migration similarly lack this key. Deserialize through Option<bool> so a
+    // missing property yields None (→ false), while a present value is respected.
+    // Serializes as a plain non-optional `bool` for the API.
+    let value = Option::<bool>::deserialize(deserializer)?;
+    Ok(value.unwrap_or(false))
+}
+
 impl UserDetails {
     /// Retrieves details by user ID, first trying to get from Redis, then from Neo4j if not found.
     pub async fn get_by_id(user_id: &str) -> ModelResult<Option<Self>> {
@@ -101,6 +116,7 @@ impl UserDetails {
             links: None,
             status: None,
             image: None,
+            deleted: false,
         }
     }
 
@@ -113,6 +129,23 @@ impl UserDetails {
             image: homeserver_user.image,
             id: user_id.clone(),
             indexed_at: Utc::now().timestamp_millis(),
+            deleted: false,
+        }
+    }
+
+    /// Cleared profile written when a user with relationships is deleted.
+    /// Every field is wiped; `deleted` is the only signal. `name` is emptied
+    /// rather than dropped because it is not optional.
+    pub fn tombstone(user_id: &PubkyId) -> Self {
+        UserDetails {
+            name: String::new(),
+            bio: None,
+            id: user_id.clone(),
+            links: None,
+            status: None,
+            image: None,
+            indexed_at: Utc::now().timestamp_millis(),
+            deleted: true,
         }
     }
 
@@ -169,5 +202,36 @@ mod tests {
             .expect("should deserialize without links property (Neo4j drops null properties)");
         assert_eq!(details.name, "Dave");
         assert!(details.links.is_none());
+        assert!(
+            !details.deleted,
+            "missing 'deleted' property should deserialize as false (pre-migration compatibility)"
+        );
+    }
+
+    /// Deserializing a UserDetails from a BoltNode with `deleted: true` should
+    /// preserve the flag — this is the shape after the one-shot migration has
+    /// marked a sentinel tombstone.
+    #[test]
+    fn deserialize_from_node_with_deleted_true() {
+        let mut props = BoltMap::new();
+        props.put(BoltString::from("name"), BoltType::from("[DELETED]"));
+        props.put(
+            BoltString::from("id"),
+            BoltType::from("uo7jgkykft4885n8cruizwy6khw71mnu5pq3ay9i8pw1ymcn85ko"),
+        );
+        props.put(
+            BoltString::from("indexed_at"),
+            BoltType::from(1724134095000_i64),
+        );
+        props.put(BoltString::from("deleted"), BoltType::from(true));
+
+        let node = Node::new(BoltNode::new(
+            BoltInteger::new(1),
+            BoltList::from(vec![BoltType::from("User")]),
+            props,
+        ));
+
+        let details: UserDetails = node.to().expect("should deserialize with deleted: true");
+        assert!(details.deleted);
     }
 }
--- a/nexus-common/src/models/user/mod.rs
+++ b/nexus-common/src/models/user/mod.rs
@@ -21,8 +21,3 @@ pub use stream::{
     USER_INFLUENCERS_KEY_PARTS, USER_MOSTFOLLOWED_KEY_PARTS,
 };
 pub use view::UserView;
-
-/// Sentinel value used to mark deleted users in the system.
-/// When a user with relationships is deleted, their name field is set to this value
-/// instead of fully removing their profile data.
-pub const USER_DELETED_SENTINEL: &str = "[DELETED]";
--- a/nexus-watcher/src/events/handlers/user.rs
+++ b/nexus-watcher/src/events/handlers/user.rs
@@ -21,13 +21,15 @@ pub async fn sync_put(user: PubkyAppUser, user_id: PubkyId) -> Result<(), EventP
     // Step 2: Save to graph
     user_details.put_to_graph().await?;
 
-    // Step 3: Run in parallel the cache process: SAVE TO INDEX
+    // Step 3: Reindex search BEFORE refreshing the details cache. `put_to_index`
+    // resolves the stale `name:id` member from the cached JSON, so a concurrent
+    // `UserDetails::put_to_index` could overwrite it with the new name first and
+    // leak the old member forever.
+    UserSearch::put_to_index(&[&user_details]).await?;
+
+    // Step 4: Run in parallel the remaining cache writes: SAVE TO INDEX
     let indexing_results = nexus_common::traced_join!(
         tracing::info_span!("index.write");
-        async {
-            UserSearch::put_to_index(&[&user_details]).await?;
-            Ok::<(), EventProcessorError>(())
-        },
         async {
             // TODO: Use SCARD on a set for unique tag count to avoid race conditions in parallel processing
             // If new user (no existing counts), save a new `UserCounts`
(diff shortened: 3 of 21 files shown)


aintnostressin:
@greptile review

nexusd/src/migrations/migrations_list/user_deleted_flag_1780617600.rs
```diff
@@ -0,0 +1,144 @@
...
+    async fn backfill(&self) -> Result<(), DynError> {
+        let graph = get_neo4j_graph()?;
+
+        // MATCH stays outside the subquery so IN TRANSACTIONS batches the rows it feeds in;
+        // the trailing count forces the batches to run. Drain predicate: see # Idempotency.
+        let query = Query::new(
+            "user_deleted_flag_backfill",
+            "MATCH (u:User)
+             WHERE u.deleted IS NULL OR (u.name = '[DELETED]' AND u.deleted = false)
+             CALL (u) {
```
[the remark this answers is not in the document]
aintnostressin: It's impossible to create an user with `[DELETED]` name because pubky-app-specs prohibits that.

nexus-watcher/src/events/handlers/user.rs
```diff
@@ -81,15 +80,20 @@ pub async fn del(user_id: PubkyId) -> Result<(), EventProcessorError> {
             exec_single_row(queries::del::delete_user(&user_id)).await?;
```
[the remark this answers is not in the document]
aintnostressin: Is this a preexisting issue?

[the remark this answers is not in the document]
aintnostressin: let's keep it as it is

nexusd/src/migrations/migrations_list/user_deleted_flag_1780617600.rs
```diff
@@ -0,0 +1,144 @@
...
+/// `migrations_backfill_ready`. The second disjunct
+/// (`u.name = '[DELETED]' AND u.deleted = false`) re-catches users tombstoned by
+/// pre-cutover code. `IN TRANSACTIONS` commits per batch, so a mid-run failure leaves the
+/// backfill partial and not Done, as the old client loop did; re-running finishes it.
+///
+/// # Rollback caveat
+/// New-code tombstones leave `name` empty, not `'[DELETED]'`. Rolling back to code that
+/// filters by the sentinel name makes them appear live, with an empty profile.
+///
+/// # Deploy ordering
```
SHAcollision, line 53: Can we move this to the PR description? The ordering is part of the correctness here and a doc comment is not where someone looks while deploying. Same for the rollback caveat above: once we write the first new tombstone an old binary reads it as live, so a binary only rollback is not safe anymore, it needs a data fix.

Maybe also add the check to run after it, something like `MATCH (u:User) WHERE u.deleted IS NULL RETURN count(u)` returning 0, so we know the backfill really drained before starting the new services.

nexusd/src/migrations/migrations_list/user_deleted_flag_1780617600.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    fn is_multi_staged(&self) -> bool {
+        false
+    }
+
+    async fn dual_write(_data: Box<dyn std::any::Any + Send + 'static>) -> Result<(), DynError> {
+        Ok(())
+    }
+
+    async fn backfill(&self) -> Result<(), DynError> {
```
SHAcollision, line 82: No migration in the repo has tests so I am not asking for one here, but this backfill is the only part of the PR that no test touches. Could you run it once against a staging copy and paste the two counts it logs? Mostly to see the duration and that `CALL ... IN TRANSACTIONS` behaves on a real sized graph.

nexus-common/src/models/user/details.rs
```diff
@@ -113,6 +129,23 @@ impl UserDetails {
...
             id: user_id.clone(),
             indexed_at: Utc::now().timestamp_millis(),
+            deleted: false,
+        }
+    }
+
+    /// Cleared profile written when a user with relationships is deleted.
+    /// Every field is wiped; `deleted` is the only signal. `name` is emptied
+    /// rather than dropped because it is not optional.
+    pub fn tombstone(user_id: &PubkyId) -> Self {
```
SHAcollision, line 139: After the migration we end up with two shapes of tombstone: the old ones keep `name = "[DELETED]"` and the new ones get `name = ""`. For us it does not matter since we only read the flag, but a client that shows the name sees two different things forever.

Two ways I see: a small follow up migration that blanks the old names once we are sure we will not roll back, or the opposite, write "[DELETED]" here as a plain display name and let the flag be the only thing code reads. I slightly prefer the follow up. wdyt?

nexus-common/src/models/user/details.rs
```diff
@@ -45,6 +45,8 @@ pub struct UserDetails {
     pub status: Option<String>,
     pub image: Option<String>,
     pub indexed_at: i64,
+    #[serde(deserialize_with = "deserialize_user_deleted", default)]
+    pub deleted: bool,
```
SHAcollision, line 49: I grepped pubky-app and it does not look at the user name sentinel anywhere, only at post content. So today a deleted author renders literally as "[DELETED]" and after this it renders empty. Small but user visible. Can we open an issue there to read `deleted` instead, so both land close in time?

nexus-watcher/tests/event_processor/users/del_with_relations.rs
```diff
@@ -300,3 +303,207 @@ async fn test_delete_recommended_user() -> Result<()> {
...
+        status: None,
+    };
+    let user_id = test.create_user(&user_kp, &user).await?;
+
+    // Rename to the legacy sentinel while staying live.
+    let mut user_details = find_user_details(&user_id).await?;
+    user_details.name = "[DELETED]".to_string();
+    user_details.put_to_graph().await?;
+
+    // Order matters: `UserSearch::put_to_index` drops the stale `name:id` member
```
SHAcollision: Related to this comment, not to the PR itself. `sync_put` runs `UserSearch::put_to_index` and `UserDetails::put_to_index` inside the same `traced_join!`, so in prod the details cache can be rewritten before search resolves the old name, and the old `name:id` member leaks. The test passes because here we order them by hand.

Since we are already in this area, maybe just await the search reindex before the details one in `sync_put`? Or if you prefer to leave it out of this PR, fine, but then maybe soften the claim in this comment.

nexus-common/src/models/user/mod.rs
```diff
@@ -21,8 +21,3 @@ pub use stream::{
     USER_INFLUENCERS_KEY_PARTS, USER_MOSTFOLLOWED_KEY_PARTS,
 };
 pub use view::UserView;
```
SHAcollision, line 25: nexus-common is published on crates.io (0.4.0 is up there, workspace says 0.4.1). This removes a pub const and adds a pub field to `UserDetails`, both break a downstream build. So next publish should be 0.5.0 and not a patch, no? Not for this PR, just so we do not ship it as 0.4.2.

SHAcollision (approved):
Nice, keeping this as a flag instead of hiding the state inside the name is much cleaner, and the del path ordering (search first, then graph, then cache) looks right to me.

I went through the read paths and I think the code is correct: the details query returns the whole node so the flag survives, `create_user` is the only place we MERGE a `:User` and it always writes `deleted`, and the ingestor skips users that already exist so a tombstone cannot come back as a stub. Also the null name case, the old `u.name <> '[DELETED]'` was dropping rows with no name and the new predicate keeps them, but we never write a User without a name so it ends up the same.

Comments are mostly about the rollout, plus one about the crate version. Nothing that changes the design.

Small thing on the description: I think today nobody can actually have "[DELETED]" as a name, pubky-app-specs rewrites it to "anonymous" on sanitize. So maybe better to say we are replacing an implicit sentinel with an explicit field clients can read, instead of saying a live user was being treated as deleted, just so the motivation reads true.

nexusd/src/migrations/migrations_list/user_deleted_flag_1780617600.rs
```diff
@@ -0,0 +1,144 @@
...
+        let graph = get_neo4j_graph()?;
+
+        // MATCH stays outside the subquery so IN TRANSACTIONS batches the rows it feeds in;
+        // the trailing count forces the batches to run. Drain predicate: see # Idempotency.
+        let query = Query::new(
+            "user_deleted_flag_backfill",
+            "MATCH (u:User)
+             WHERE u.deleted IS NULL OR (u.name = '[DELETED]' AND u.deleted = false)
+             CALL (u) {
+                 SET u.deleted = coalesce(u.name = '[DELETED]', false)
```
tipogi: do we need to write `deleted = false` on every live `:User`?

Reads already treat a missing property as live — Rust via `deserialize_user_deleted` / default, and Cypher via `NOT coalesce(u.deleted, false)` (which this PR keeps on every filter). So for correctness it should be enough to mark tombstones only:

```cypher
MATCH (u:User)
WHERE u.name = '[DELETED]' AND coalesce(u.deleted, false) = false
SET u.deleted = true
```

The full-label false backfill only pays off if we also drop coalesce and switch to `WHERE NOT u.deleted` in the same change. As written we keep coalesce anyway, so we take the larger write without using that benefit. Thin migration = same semantics, shorter cutover.

Drain check would become something like: 
```cypher
MATCH (u:User) 
WHERE u.name = '[DELETED]' AND coalesce(u.deleted, false) = false 
RETURN count(u)
```

nexus-common/src/models/user/details.rs
tipogi: Agree on blanking rather than keeping `[DELETED]` as a display name. I’d do it in this migration instead of a follow-up — set deleted = true and blank ("") the name field

nexus-common/src/models/user/details.rs
tipogi: https://github.com/pubky/pubky-app/issues/2338

nexusd/src/migrations/migrations_list/user_deleted_flag_1780617600.rs
aintnostressin: done [ea637b7](https://github.com/pubky/pubky-nexus/pull/1009/commits/ea637b736561b880f5c63d22112682bd7b8b66ef)

nexusd/src/migrations/migrations_list/user_deleted_flag_1780617600.rs
aintnostressin: I've updated PR description and added check in [ea637b7](https://github.com/pubky/pubky-nexus/pull/1009/commits/ea637b736561b880f5c63d22112682bd7b8b66ef)

nexus-common/src/models/user/details.rs
aintnostressin: included in migration [64f2ab0](https://github.com/pubky/pubky-nexus/pull/1009/commits/64f2ab04520ce4cce2050e2bd1de25afa0140a58)

nexus-common/src/models/user/mod.rs
aintnostressin: yes, but there are more breaking changes (in previous commits)

nexusd/src/migrations/migrations_list/user_deleted_flag_1780617600.rs
aintnostressin: I synced local nexus with staging homeserver. Staging has 7 users marked as deleted while my copy has only one. I run migration (both values fit into transaction size) on my local neo4j.

```
2026-08-17T11:50:38.027978Z  INFO nexusd::migrations::manager:169: Storing new migration UserDeletedFlag1780617600...
2026-08-17T11:50:38.031446Z  INFO nexusd::migrations::manager:190: Migration UserDeletedFlag1780617600 is at phase backfill
2026-08-17T11:50:38.072141Z  INFO nexusd::migrations::migrations_list::user_deleted_flag_1780617600:103: UserDeletedFlag migration: 1 tombstones flagged and names cleared
2026-08-17T11:50:38.072213Z  INFO neo4rs::pool:49: creating new connection...
2026-08-17T11:50:38.084448Z  INFO neo4rs::pool:49: creating new connection...
2026-08-17T11:50:38.114693Z  INFO nexusd::migrations::migrations_list::user_deleted_flag_1780617600:167: UserDeletedFlag migration: invalidated 1 tombstone cache entries (1 total)
2026-08-17T11:50:38.114770Z  INFO nexusd::migrations::migrations_list::user_deleted_flag_1780617600:178: UserDeletedFlag migration: complete — 1 tombstones flagged and cleared, 1 cache entries invalidated
2026-08-17T11:50:38.114806Z  INFO nexusd::migrations::manager:204: Migration UserDeletedFlag1780617600 completed phase backfill successfully!
```

Please note that i changed cypher slightly in [ea637b7](https://github.com/pubky/pubky-nexus/pull/1009/commits/ea637b736561b880f5c63d22112682bd7b8b66ef) so it only affects users with '[DELETED]' name so total rows changed will be much lower compared to expected number from version you commented on.

nexus-watcher/tests/event_processor/users/del_with_relations.rs
aintnostressin: That's true. 

I've fixed the race condition in [25c0f5f](https://github.com/pubky/pubky-nexus/pull/1009/commits/25c0f5f98e5956d2d5c0bcea6b5e04380e414573). The test also uses handler (cannot go through full path because it uses [DELETED] user name which is forbidden by specs).

tipogi (approved):

talosmachina:
Marking as blocked: this PR needs pubky/pubky-app#2338 (client adaptation to the Nexus user `deleted` flag) to land first.

PR #1011, opened 2026-08-06
fix: detect HS transport errors on HS indexing

ok300:
This PR introduces transport error detection when processing events from an External HS.

Now, when a transport error is seen, the current processor run will skip processing any other users from this HS, until the next processor run.

This fixes possible delays introduced by trying to index events of users from an External HS, in the case when that HS went offline.

---

### Pre-submission Checklist


tipogi (approved):
nice! 🚀

PR #1012, opened 2026-08-06
fix: preserve key-based indexing progress on timeout

### LINKED ISSUE #1013: bug: gracefully stop key-based indexing when its time budget expires
External key-based indexing currently enforces its custom per-homeserver limit by immediately aborting the processor task. If the timeout occurs while processing a user, the in-memory cursor is discarded before it can be persisted to Redis. Successfully handled events may therefore be fetched and processed again during the next run. Later users are also skipped, and the timeout triggers homeserver backoff.

## Possible solution

Replace immediate timeout abortion with two-phase cancellation. When the timeout expires, signal a per-run cancellation token and allow processing to stop at the next safe event boundary. The event loop should return its latest safe cursor, `process_user` should persist it using the existing cursor write, and the processor should stop before starting another user.

The runner should wait for cleanup for a short grace period and force-abort only if the task remains stuck. Normal budget exhaustion should be represented separately from a homeserver failure so that a healthy but busy homeserver does not enter exponential backoff.

## Acceptance criteria

- [ ] Budget expiry requests cooperative cancellation before aborting.
- [ ] A completed event's cursor is persisted before the processor exits.
- [ ] Cancellation does not interrupt `handle_event()` mid-operation under normal conditions.
- [ ] No later user starts after the budget expires.
- [ ] A bounded grace period prevents indefinitely stuck tasks.
- [ ] Forced abort remains available as a last resort.
- [ ] Budget exhaustion does not trigger homeserver failure backoff.
- [ ] Tests cover cancellation during event handling, cursor persistence, stuck handlers, and later-user behavior.

tipogi:
### Summary
Adds cooperative timeout handling for key-based indexing. When its processing budget expires, the processor finishes the current event, retries cursor persistence, and stops before processing another event or user.

Budget exhaustion is tracked separately from forced timeouts and resets homeserver backoff. Other processors retain their existing hard-timeout behavior, with aborted tasks awaited to prevent overlapping runs.

### Pre-submission Checklist

- [x] I manually reviewed and tested the PR 

### DIFF
--- a/nexus-watcher/src/service/indexer/key_based.rs
+++ b/nexus-watcher/src/service/indexer/key_based.rs
@@ -17,13 +17,56 @@ use pubky_app_specs::PubkyId;
 use tokio::sync::watch::Receiver;
 use tracing::{debug, error, info, warn};
 
-use super::TEventProcessor;
+use super::{ProcessorResult, RunCompletion, RunContext, TEventProcessor, TimeoutPolicy};
 use crate::events::retry::RetryScheduler;
 use crate::events::EventHandler;
 use crate::service::runner::UserNotFoundBackoff;
 use crate::service::user_hs_resolver;
 
 const FETCH_EVENTS_429_BACKOFF_SECS: [u64; 3] = [1, 2, 3];
+const CURSOR_PERSIST_BACKOFF_MS: [u64; 2] = [100, 250];
+const KEY_BASED_TIMEOUT_GRACE: Duration = Duration::from_secs(10);
+
+/// Controls whether homeserver indexing continues after processing one user.
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum ProcessingControl {
+    /// Continue with the next user.
+    Continue,
+    /// End the run and report cooperative budget exhaustion.
+    StopOnBudgetExhaustion,
+}
+
+/// Progress from processing one user's events, before cursor persistence.
+struct EventProgress {
+    latest_cursor: Option<u64>,
+    result: Result<ProcessingControl, EventProcessorError>,
+}
+
+impl EventProgress {
+    /// Keep going; persist `latest_cursor` if any.
+    fn continuing(latest_cursor: Option<u64>) -> Self {
+        Self {
+            latest_cursor,
+            result: Ok(ProcessingControl::Continue),
+        }
+    }
+
+    /// Budget exhausted; persist `latest_cursor` if any, then stop.
+    fn stop_on_budget_exhaustion(latest_cursor: Option<u64>) -> Self {
+        Self {
+            latest_cursor,
+            result: Ok(ProcessingControl::StopOnBudgetExhaustion),
+        }
+    }
+
+    /// Handler or validation failed; persist `latest_cursor` if any, then return the error.
+    fn failed(latest_cursor: Option<u64>, error: EventProcessorError) -> Self {
+        Self {
+            latest_cursor,
+            result: Err(error),
+        }
+    }
+}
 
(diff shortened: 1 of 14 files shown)


nexus-watcher/src/service/indexer/key_based.rs
```diff
@@ -116,6 +116,11 @@ impl TEventProcessor for KeyBasedEventProcessor {
         Some(self.homeserver_id.as_ref())
     }
 
+    fn custom_timeout(&self) -> Option<Duration> {
+        // Some(Duration::from_mins(5))
+        Some(Duration::from_millis(1000))
```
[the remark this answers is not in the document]
tipogi: ae1a946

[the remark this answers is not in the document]
tipogi: out of context for that PR, opened an issue: https://github.com/pubky/pubky-nexus/issues/1013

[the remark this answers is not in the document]
tipogi: f774868

nexus-watcher/src/service/indexer/mod.rs
ok300, line 1: Good catch!

Would it make sense to add a test, since it touches such an important part of the scheduler?

tipogi: dd600bd

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -88,22 +88,25 @@ pub trait TEventProcessor: Send + Sync + 'static {
...
-                |_| error!(service = %instance_name, homeserver, "Event processor timed out"),
-            )
-            .map_err(|_| RunError::TimedOut)?;
+        let mut handle = tokio::spawn(self.run_internal().instrument(span));
+
+        let join_result = match tokio::time::timeout(timeout, &mut handle).await {
+            Ok(result) => result,
+            Err(_) => {
+                error!(service = %instance_name, homeserver, "Event processor timed out");
+                handle.abort();
```
ok300: > handle.abort();

Hard abort can leave permanently inconsistent data

This `abort()` can cancel during a multi-store event write. For example, a bookmark may commit to Neo4j at `events/handlers/bookmark.rs:39`, then abort before Redis/count updates at lines 51-57. The cursor write at `service/indexer/key_based.rs:260` is also skipped, but replay sees the existing graph edge and permanently skips the count increment.

Cancellation should occur between events, followed by persisting the safe cursor. This is also tracked by #1013

nexus-watcher/src/service/indexer/mod.rs
```diff
@@ -88,22 +88,25 @@ pub trait TEventProcessor: Send + Sync + 'static {
...
+                let _ = handle.await;
```
ok300: > let _ = handle.await;

Cancellation waiting is unbounded

After `handle.abort()`, this line awaits indefinitely. Tokio abort is cooperative, so non-yielding or blocking work can prevent `run()` from ever returning, blocking all later HSs in the sequential runner.

One solution would be to use a bounded cleanup grace period and retain explicit ownership of any task that has not terminated.

ok300 (commented):
I know this is WIP, but wanted to add my review for the current state. Maybe this helps with planning the next steps.

tipogi:
Closing this PR for now.

I still believe that dropping the `JoinHandle` on timeout and allowing the processor to continue in the background is undesirable. However, this PR introduces a broader cooperative-timeout model whose additional complexity does not currently have sufficient production evidence to justify it.

PR #1031, opened 2026-08-17
feat(api): search users by profile tags

### LINKED ISSUE #1030: feat(api): search users by profile tags
FE needs this for the search page redesign (pubky/pubky-app#1841). Tag search gets a "People" section: users whose profile is tagged with the searched tag(s).

We have this for posts already (`/v0/search/posts/by_tag/{tag}`) but for users it doesn't exist at any layer, no index, no query. The graph edge is already there, so it's a mirror of the posts search, not new architecture.

Proposal:

```
GET /v0/search/users/by_tags?tags=synonym,rust&skip=0&limit=20
-> [ { "user_id": "pxnu3...", "score": 12 }, ... ]
```

- `tags`: 1 to 5 labels, comma separated, same validation as post streams. Union semantics (any of the tags).
- `score`: how many taggers applied that label to the user, descending. With multiple tags, the sum.
- Only ids + scores, no hydration. FE hydrates with `POST /v0/stream/users/by_ids`.

Internals in short: new inverted Redis index label -> users (like the posts one), watcher keeps it fresh on tag put/del, Cypher fallback for multi tag, and a backfill migration for live deploys.

Acceptance:

- [ ] Endpoint returns tagged users ordered by tagger count, paginated
- [ ] Watcher updates the index on user tag put/del
- [ ] Backfill migration for existing data

I think this covers what FE needs for the People section, please confirm on this ticket if something is missing.

SHAcollision:
Closes #1030

BE support for the search redesign People section (pubky/pubky-app#1841). FE gets user ids + scores and hydrates with the existing `POST /v0/stream/users/by_ids`.

```
GET /v0/search/users/by_tags?tags=synonym,rust&skip=0&limit=20
-> [ { "user_id": "pxnu3...", "score": 12 }, ... ]
```

- `tags`: 1 to 5 labels, union semantics. `score`: how many taggers applied the label to the user, with multiple tags the sum. Descending, tie order unspecified.
- 1 label reads a new Redis index `Sorted:Tags:Global:User:Taggers:<label>`, 2 to 5 labels fall back to a Cypher query, same split the post streams do. The graph path stays in the default rate-limit tier like `/stream/posts?tags=`, we can move it to the expensive tier later if it shows up in metrics.
- The index score is derived from the `User:Taggers:{user}:{label}` set instead of counting increments, and the derive runs as one Lua script (SCARD + conditional ZADD/ZREM), so concurrent events can't commit a stale score and watcher retries converge on their own, no drift and no gating.
- `db mock` builds the index, and there is a `UsersByTagsIndexBackfill` migration for live deploys. The backfill only enumerates (user, label) pairs from the graph and derives every score from the live taggers sets, so it can't overwrite concurrent watcher writes with snapshot state.

Deploy note: after the new watcher is live, add `UsersByTagsIndexBackfill1786924800` to `backfill_ready` in the deployed config and run `nexusd db migration run`.

Deleted users are excluded: every profile write re-derives the user's per-label entries, so a tombstone evicts them and a recreated profile restores their retained tags. The derive re-checks the tombstone in the same Lua script so late tag events can't resurrect them, and the graph path filters the `[DELETED]` sentinel.

### DIFF
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -307,7 +307,9 @@ pub fn search_users_by_tags(labels: &[String], skip: Option<usize>, limit: Optio
         WHERE tag.label IN $labels AND u.name <> $deleted
         WITH u, COUNT(tag) AS score
         RETURN u.id AS user_id, score
-        ORDER BY score DESC, u.id ASC
+        // id DESC matches how Redis breaks equal scores (reverse-lex member
+        // order), keeping pagination windows identical across both paths
+        ORDER BY score DESC, u.id DESC
         ",
     );
 
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -3,6 +3,7 @@ use crate::db::graph::Query;
 use crate::db::kv::SortOrder;
 use crate::models::post::{KindFilter, StreamSource};
 use crate::models::resource::stream::ResourceSorting;
+use crate::models::user::USER_DELETED_SENTINEL;
 use crate::types::routes::HotTagsInputDTO;
 use crate::types::DomainTrust;
 use crate::types::Pagination;
@@ -273,11 +274,13 @@ pub fn global_tags_by_user() -> Query {
         // create_user_tag MERGEs one TAGGED edge per (tagger, tagged, label),
         // so COUNT(t) is the distinct tagger count.
         MATCH (tagger:User)-[t:TAGGED]->(u:User)
+        WHERE u.name <> $deleted
         WITH t.label AS label, u.id AS user_id, COUNT(t) AS score
         WITH label, COLLECT([toFloat(score), user_id]) AS sorted_set
         RETURN label, sorted_set
         ",
     )
+    .param("deleted", USER_DELETED_SENTINEL)
 }
 
 /// Enumerates the distinct (tagged user, label) pairs carried by user
@@ -288,9 +291,11 @@ pub fn get_user_tag_pairs() -> Query {
         "get_user_tag_pairs",
         "
         MATCH (:User)-[t:TAGGED]->(u:User)
+        WHERE u.name <> $deleted
         RETURN DISTINCT u.id AS user_id, t.label AS label
         ",
     )
+    .param("deleted", USER_DELETED_SENTINEL)
 }
 
 /// Users whose profile carries any of the given tag labels, scored by distinct
@@ -299,7 +304,7 @@ pub fn search_users_by_tags(labels: &[String], skip: Option<usize>, limit: Optio
     let mut cypher = String::from(
         "
         MATCH (tagger:User)-[tag:TAGGED]->(u:User)
-        WHERE tag.label IN $labels
+        WHERE tag.label IN $labels AND u.name <> $deleted
         WITH u, COUNT(tag) AS score
         RETURN u.id AS user_id, score
         ORDER BY score DESC, u.id ASC
@@ -313,7 +318,9 @@ pub fn search_users_by_tags(labels: &[String], skip: Option<usize>, limit: Optio
         cypher.push_str(&format!("LIMIT {}\n", limit.min(MAX_QUERY_LIMIT)));
     }
 
-    Query::new("search_users_by_tags", &cypher).param("labels", labels.to_vec())
+    Query::new("search_users_by_tags", &cypher)
+        .param("labels", labels.to_vec())
+        .param("deleted", USER_DELETED_SENTINEL)
 }
 
 // Retrieve all the tags of the post
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -264,6 +264,45 @@ pub fn global_tags_by_post_engagement() -> Query {
     )
 }
 
+/// Retrieves unique global tags on user profiles, returning per label the tagged
+/// user ids paired with their distinct tagger counts.
+pub fn global_tags_by_user() -> Query {
+    Query::new(
+        "global_tags_by_user",
+        "
+        // create_user_tag MERGEs one TAGGED edge per (tagger, tagged, label),
+        // so COUNT(t) is the distinct tagger count.
+        MATCH (tagger:User)-[t:TAGGED]->(u:User)
+        WITH t.label AS label, u.id AS user_id, COUNT(t) AS score
+        WITH label, COLLECT([toFloat(score), user_id]) AS sorted_set
+        RETURN label, sorted_set
+        ",
+    )
+}
+
+/// Users whose profile carries any of the given tag labels, scored by distinct
+/// tagger count summed across the searched labels.
+pub fn search_users_by_tags(labels: &[String], skip: Option<usize>, limit: Option<usize>) -> Query {
+    let mut cypher = String::from(
+        "
+        MATCH (tagger:User)-[tag:TAGGED]->(u:User)
+        WHERE tag.label IN $labels
+        WITH u, COUNT(tag) AS score
+        RETURN u.id AS user_id, score
+        ORDER BY score DESC, u.id ASC
+        ",
+    );
+
+    if let Some(skip) = skip {
+        cypher.push_str(&format!("SKIP {}\n", skip.min(MAX_QUERY_SKIP)));
+    }
+    if let Some(limit) = limit {
+        cypher.push_str(&format!("LIMIT {}\n", limit.min(MAX_QUERY_LIMIT)));
+    }
+
+    Query::new("search_users_by_tags", &cypher).param("labels", labels.to_vec())
+}
+
 // Retrieve all the tags of the post
 pub fn post_tags(user_id: &str, post_id: &str) -> Query {
     Query::new(
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -280,6 +280,19 @@ pub fn global_tags_by_user() -> Query {
     )
 }
 
+/// Enumerates the distinct (tagged user, label) pairs carried by user
+/// profiles, without counts: backfills derive each score from the live
+/// taggers set instead of trusting a snapshot value.
+pub fn get_user_tag_pairs() -> Query {
+    Query::new(
+        "get_user_tag_pairs",
+        "
+        MATCH (:User)-[t:TAGGED]->(u:User)
+        RETURN DISTINCT u.id AS user_id, t.label AS label
+        ",
+    )
+}
+
 /// Users whose profile carries any of the given tag labels, scored by distinct
 /// tagger count summed across the searched labels.
 pub fn search_users_by_tags(labels: &[String], skip: Option<usize>, limit: Option<usize>) -> Query {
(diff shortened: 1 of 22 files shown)


nexus-watcher/src/events/handlers/user.rs
```diff
@@ -46,6 +46,13 @@ pub async fn sync_put(user: PubkyAppUser, user_id: PubkyId) -> Result<(), EventP
     indexing_results.0?;
     indexing_results.1?;
     indexing_results.2?;
+
+    // Profile writes flip the tombstone state in both directions: deletion
+    // writes the sentinel, recreation clears it. Re-derive the user's
+    // per-label search entries afterwards, so tombstones are evicted and
+    // recreated profiles get their retained tags back.
+    UsersByTagSearch::sync_user_labels(&user_id).await?;
```
tipogi, line 54: this runs on every profile update, including normal name/bio edits. It scans all distinct labels and makes one Redis call per label sequentially. This helps retry recovery, but it could be expensive for highly tagged users.

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -264,6 +265,64 @@ pub fn global_tags_by_post_engagement() -> Query {
...
+    .param("deleted", USER_DELETED_SENTINEL)
+}
+
+/// Users whose profile carries any of the given tag labels, scored by distinct
+/// tagger count summed across the searched labels.
+pub fn search_users_by_tags(labels: &[String], skip: Option<usize>, limit: Option<usize>) -> Query {
+    let mut cypher = String::from(
+        "
+        MATCH (tagger:User)-[tag:TAGGED]->(u:User)
+        WHERE tag.label IN $labels AND u.name <> $deleted
```
tipogi, line 307: what about if we add an index for `TAGGED.label`? Multi-label searches filter directly on this relationship property, so without an index Neo4j may scan all TAGGED relationships, including for nonexistent labels

nexus-webapi/src/routes/v0/search/mod.rs
```diff
@@ -30,6 +30,10 @@ pub fn routes() -> Router<AppState> {
             SEARCH_USERS_BY_ID_ROUTE,
             get(users::search_users_by_id_handler),
         )
+        .route(
+            SEARCH_USERS_BY_TAGS_ROUTE,
+            get(users::search_users_by_tags_handler),
+        )
```
tipogi: multi-label requests perform graph aggregation, but this endpoint is registered in the _default_ rate-limit bucket. Should we add it in the _expensive_ bucket initially, until we have enough production metrics to confirm these queries are safe under the _default_ limits?

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -264,6 +265,64 @@ pub fn global_tags_by_post_engagement() -> Query {
...
+        WITH u, COUNT(tag) AS score
+        RETURN u.id AS user_id, score
+        ORDER BY score DESC, u.id ASC
```
tipogi: redis orders equal scores by user ID in _reverse lexicographical order_, while this graph query uses `u.id ASC`. This means equivalent searches can return different pagination windows depending on whether they use the Redis or graph path. Should we use `u.id DESC` here to keep both paths consistent?

nexus-common/src/models/user/search.rs
```diff
@@ -1,13 +1,219 @@
...
+    /// Backfills the index against live data: the graph only enumerates
+    /// candidate (user, label) pairs, and every score is derived atomically
+    /// from the current taggers set by [`Self::sync_index_score`]. Events
+    /// landing during the backfill are therefore never overwritten with
+    /// snapshot state; a pair whose tag was deleted mid-backfill derives an
+    /// empty set and stays out of the index.
+    ///
+    /// # Errors
+    /// Returns an error when the graph enumeration or a Redis sync fails.
+    pub async fn backfill_from_graph() -> ModelResult<()> {
```
tipogi: Since this full-dataset traversal is only used by *UsersByTagsIndexBackfill1786924800*, would it be better to keep the backfill orchestration in the migration? The model could expose `sync_index_score`, while the migration handles graph streaming, batching, concurrency, and progress reporting. This would also make the production backfill’s operational behavior more explicit.

nexus-common/src/models/user/search.rs
```diff
@@ -1,13 +1,219 @@
...
+        let rows = fetch_all_rows_from_graph(queries::get::get_user_tag_pairs()).await?;
+        for row in rows {
+            let user_id: String = row.get("user_id")?;
+            let label: String = row.get("label")?;
+            Self::sync_index_score(&user_id, &label).await?;
+        }
+        Ok(())
+    }
```
tipogi: This live backfill collects every distinct `(user, label)` pair into memory before starting Redis work, then performs one Redis round trip per pair sequentially. What about if we stream the graph rows and process them with bounded concurrency—for example, batches of 500 with 16–32 concurrent Redis operations? Since this may be a long-running production migration, it would also be useful to log periodic progress such as processed count, elapsed time, throughput, and batch failures.

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -264,6 +265,64 @@ pub fn global_tags_by_post_engagement() -> Query {
...
+pub fn global_tags_by_user() -> Query {
+    Query::new(
+        "global_tags_by_user",
+        "
+        // create_user_tag MERGEs one TAGGED edge per (tagger, tagged, label),
+        // so COUNT(t) is the distinct tagger count.
+        MATCH (tagger:User)-[t:TAGGED]->(u:User)
+        WHERE u.name <> $deleted
+        WITH t.label AS label, u.id AS user_id, COUNT(t) AS score
+        WITH label, COLLECT([toFloat(score), user_id]) AS sorted_set
```
aintnostressin, line 279: should tags from low trust users be counted here ?

nexus-common/src/models/user/search.rs
```diff
@@ -1,13 +1,219 @@
...
+    /// Indexes user profile tags into per-label global sorted sets from a
+    /// graph snapshot. Only for full rebuilds over quiescent or empty Redis
+    /// (`db mock`, full reindex); live backfills go through
+    /// [`Self::backfill_from_graph`] instead, which cannot overwrite
+    /// concurrent watcher writes with snapshot state.
+    ///
+    /// # Errors
+    /// Returns an error when the graph read or a Redis write fails.
+    pub async fn reindex() -> ModelResult<()> {
+        let rows = fetch_all_rows_from_graph(queries::get::global_tags_by_user()).await?;
```
aintnostressin: is it possible to paginate this rather than loading full set into memory ?

nexus-common/src/models/user/search.rs
```diff
@@ -1,13 +1,219 @@
...
+                    .map(|list| list.into_iter().map(Into::into).collect())
+                    .unwrap_or_default())
+            }
+            // Union across labels needs an aggregation over multiple sorted sets,
+            // so it goes to the graph instead
+            _ => Self::get_from_graph(labels, pagination.skip, pagination.limit).await,
+        }
+    }
+
+    async fn get_from_graph(
```
aintnostressin, line 110: `get_from_graph` functions usually return `GraphResult`. Can this code be consistent ?

nexus-common/src/models/user/search.rs
aintnostressin: I lean toward keeping all logic used by migration in migration files to make sure they are immutable.

nexus-common/src/models/user/search.rs
```diff
@@ -1,13 +1,219 @@
...
+    /// cardinality, and the member is removed when the set empties or when
+    /// the user's details carry the deleted sentinel. Everything runs as one
+    /// Lua script, so concurrent events for the same (user, label) cannot
+    /// commit a stale score and a concurrent tombstone cannot be raced; the
+    /// callers need no gating and retries converge on their own. Must run
+    /// after the taggers SADD/SREM settled.
+    ///
+    /// # Errors
+    /// Returns an error if the Redis script execution fails.
+    pub async fn sync_index_score(user_id: &str, label: &str) -> RedisResult<()> {
```
aintnostressin, line 151: I havent found a test case checking `removal_guard`.
This is going to be changed soon so it would be good to have a test so we can easily ensure it works after the change.

nexus-watcher/src/events/handlers/user.rs
SHAcollision: True, but gating on tombstone flips breaks retry recovery (retry sees old == new and skips). One cheap Lua call per label, off the request path, keeping it. Can pipeline if metrics say it is hot.

nexus-common/src/db/graph/queries/get.rs
SHAcollision: Already exists: `taggedLabelIndex` in `db/graph/setup.rs:35`.

nexus-webapi/src/routes/v0/search/mod.rs
SHAcollision: Agree, fixed in 1fba498a.

nexus-common/src/db/graph/queries/get.rs
SHAcollision: Agree, fixed in 1fba498a. `u.id DESC` matches redis reverse-lex on z32 ids, parity test now compares exact order.

nexus-common/src/models/user/search.rs
SHAcollision: @tipogi Agree, fixed in 1fba498a.

nexus-common/src/models/user/search.rs
SHAcollision: Agree on streaming and progress logs, fixed in 1fba498a. Skipped the concurrency, redis is single threaded and the backfill is idempotent, fail fast + re-run is simpler.

nexus-common/src/db/graph/queries/get.rs
SHAcollision: For now yes, global counts like posts by_tag, trust doesn't exist yet and we need this merged. Once trust lands we should use it everywhere, this endpoint included.

nexus-common/src/models/user/search.rs
SHAcollision: Agree, fixed in 136a1a67. Streams rows now, peak memory is one label. SKIP/LIMIT would re-run the aggregation per page so streaming instead of paginating.

nexus-common/src/models/user/search.rs
SHAcollision: Agree, fixed in 136a1a67.

nexus-common/src/models/user/search.rs
SHAcollision: @aintnostressin Orchestration moved into the migration in 1fba498a. I'd keep `sync_index_score` shared though: the backfill is race-safe exactly because it runs the same live derive the watcher runs, a frozen copy would drift from the live path.

nexus-common/src/models/user/search.rs
SHAcollision: They exist in `tags/users_by_tags_sync.rs`, the tombstone tests cover the guard firing and clearing through the real flow. Added a direct fixture-level case too in 136a1a67 so your change has a tight net.

aintnostressin (approved):

PR #1033, opened 2026-08-20
feat: improve metrics

tipogi:
## Summary
Essential request/error totals were missing or hard to find: HTTP had traces only, Neo4j had `errors` but no first-class request counter, and reach-scoped graph queries lumped WoT with following under one label.
- **HTTP:** `http.server.requests`, `http.server.errors` (5xx + 408), and `http.server.request.duration` (semconv seconds). Routes use the matched template (`unmatched` otherwise); unknown methods collapse to `_OTHER`. Incoming `traceparent` is ignored on this public API. Query strings are not recorded.
- **Neo4j:** `neo4j.query.requests` next to the existing `neo4j.query.errors`, same `query` label, so error rate is `errors / requests` without using histogram `_count`.
- **WoT:** post-stream `wot.post_stream.requests` / `errors` already existed. Trusted-network tag queries were already distinct Neo4j labels. `get_*_by_reach` labels are now split by reach and WoT depth (`…_following`, `…_wot_3`, …) because those queries have no companion meter — depth cannot live as an attribute the way it does on `post_stream_wot`.

## Pre-submission Checklist

### DIFF
--- a/nexus-common/src/db/graph/instrumented.rs
+++ b/nexus-common/src/db/graph/instrumented.rs
@@ -11,7 +11,7 @@ use std::time::{Duration, Instant};
 use tracing::warn;
 
 use super::ops::{Graph, GraphOps};
-use super::query::Query;
+use super::query::{Query, TelemetryValue};
 use crate::utils::ms;
 
 /// The OpenTelemetry meter name used by all Neo4j graph metrics.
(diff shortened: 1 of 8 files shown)


nexus-webapi/src/routes/middlewares/tracing.rs
```diff
@@ -1,41 +1,186 @@
...
+    let attrs = [
+        KeyValue::new("http.request.method", method.to_string()),
+        KeyValue::new("http.route", route.to_string()),
+        KeyValue::new("http.response.status_code", i64::from(status)),
+    ];
+    METRICS.requests.add(1, &attrs);
+    METRICS.duration.record(elapsed_secs, &attrs);
+    if is_failed_request(status) {
+        METRICS.errors.add(1, &attrs);
+    }
```
SpontaneousOverthrow: delete error counter, we have response code in labels that's enough

tipogi: dd58930

nexus-common/src/db/graph/instrumented.rs
```diff
@@ -169,18 +202,16 @@ impl Drop for InstrumentedStream {
...
-                    "Slow Neo4j query"
-                );
-            }
+            warn!(
+                total_ms = total.as_millis(),
+                execute_ms = self.execute_duration.as_millis(),
+                fetch_ms = fetch_duration.as_millis(),
+                rows = self.row_count,
+                query = %self.label,
+                attrs = %format_attrs(attrs),
```
aintnostressin: I'm not sure we should log telemetry attrs -  they are available on span

nexus-common/src/db/graph/instrumented.rs
```diff
@@ -301,11 +335,13 @@ impl<G: GraphOps> GraphOps for InstrumentedGraph<G> {
...
             }
             Err(e) => {
-                let attrs: &[KeyValue] = &query_attrs(label);
+                let attrs: &[KeyValue] = &attrs;
+                self.metrics.requests.add(1, attrs);
                 self.metrics.duration.record(ms(execute_duration), attrs);
                 self.metrics
                     .execute_duration
                     .record(ms(execute_duration), attrs);
+                self.metrics.rows.record(0, attrs);
```
aintnostressin: Can we drop this line ? This is a failure path so this value is more like a `undefined`.

aintnostressin (approved):

nexus-common/src/db/graph/instrumented.rs
tipogi: c4fa94c

nexus-common/src/db/graph/instrumented.rs
tipogi: c4fa94c

PR #1034, opened 2026-08-20
 fix: default `max_file_size` to the `pubky-app-specs` blob cap (existing `config.toml` keeps its value)

aintnostressin:
Previous default value was not matching the specs. 
By default nexus should use value from `pubky-app-specs` with an option for overwritting it.

### DIFF
--- a/nexus-common/default.config.toml
+++ b/nexus-common/default.config.toml
@@ -45,11 +45,12 @@ initial_backoff_secs = 60
 # Maximum backoff duration (in seconds) for a failing homeserver
 max_backoff_secs = 3600
 retry_processor_interval_ms = 10000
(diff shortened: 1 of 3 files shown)


SHAcollision (approved):
LGTM.

Two nits, none blocking:

- I think greptile's point is real. `read_or_create_config_file` only writes the template when the file is missing, it never merges. So any deploy that already has a `~/.pubky-nexus/config.toml` keeps `max_file_size = 52428800` and stays on 50 MiB. Not a bug in the PR, but the fix only reaches fresh installs. Maybe we add a line to the release notes so ops know to delete the key? After this merges nothing tells them 50 MiB is stale anymore.
- the commented `# max_file_size = 104857600` puts the magic number back into the file that drifts the most, which is kind of what this PR is removing.

PR #1043, opened 2026-08-29
fix(watcher): configure a global Pubky HTTP request timeout

ok300:
This PR adds a configurable total HTTP request timeout to the shared Pubky client.

Note this is not a connection timeout, but a request timeout. This means the entire request, including receiving the response body, has to complete within this time.

In more detail:

- Adds `[stack.net].pubky_http_request_timeout` to `config.toml`, defaulting to **30 seconds**.
  - The deadline includes response-body reads.
  - Value of `0` is rejected during config parsing.
- Uses the same configured singleton for both mainnet and testnet clients.

---

# Pre-submission Checklist

### DIFF
--- a/nexus-common/default.config.toml
+++ b/nexus-common/default.config.toml
@@ -87,6 +87,9 @@ testnet = false
 # testnet host, leave as "localhost" for local development. Change only if the
 # testnet homeserver/relay are on another machine (e.g. Docker setups).
 testnet_host = "localhost"
+# Total request timeout in seconds for all HTTP requests made by the shared Pubky client.
+# This includes response-body reads and should accommodate max_file_size blob downloads.
+pubky_http_request_timeout = 30
 # External HS PKs which are forbidden from being indexed
 external_hs_pk_blacklist = []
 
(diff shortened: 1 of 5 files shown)


nexus-common/default.config.toml
```diff
@@ -88,6 +88,9 @@ testnet = false
 # testnet host, leave as "localhost" for local development. Change only if the
 # testnet homeserver/relay are on another machine (e.g. Docker setups).
 testnet_host = "localhost"
+# Total request timeout in seconds for all HTTP requests made by the shared Pubky client.
+# This includes response-body reads and should accommodate max_file_size blob downloads.
+pubky_http_request_timeout = 30
```
aintnostressin: 30 seconds timeout with 100 mb max file size results in requirement of  ~3.5 MB/s (~28 Mbit/s) download speed sustained for the whole transfer. This becomes a real performance requirement for homeservers. Should we adjust this number?

aintnostressin: other config property names carry time unit - can it be suffixed with `_secs` ?

ok300: Good idea, done: 74a111da1014ecc6d468ee43e9c7e8490a7451e4

ok300: Done, bumped to 300s: ed2879375a7c43fc66de6acdeb726c77ccb97799

aintnostressin (approved):

PR #1049, opened 2026-09-01
feat(api): social graph status (trustrank) on the user view

SHAcollision:
Adds one field to `UserView` for the badge in Aldert's design:

```json
"social_graph_status": "established"
```

Can be `null`, `new`, `networked` or `established`.

It is optional. `null` means we have no ranking and the app should hide the badge.

Nexus decides the category, not the client. The raw score is a share that sums to 1, so from
outside you cannot tell if 0.00003 is good or bad without the whole distribution. So we send the
category and never the number.

The label is refreshed at the end of the trustrank job, so no new cron.

Not here: picking the seed accounts. Once that is set and the job runs, badges show up on their
own, no frontend change needed.

One thing to decide: the `established` cut is the top 5%, I made that number up. Maybe we look at
the real distribution after a couple of runs on staging and change it? It is one constant, cheap to
move.

### DIFF
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -298,6 +298,31 @@ pub fn get_user_tag_pairs() -> Query {
     .param("deleted", USER_DELETED_SENTINEL)
 }
 
+/// Users carrying positive trust, highest first: the ranked population behind
+/// the social graph badge.
+///
+/// `coalesce` rather than `IS NOT NULL` because GDS may write `0.0` or leave
+/// the property unset for a node unreachable from the seed set, and both mean
+/// the same thing. Only positive scores are ranked: everyone at zero shares one
+/// value, so their order would be arbitrary and would reshuffle every run.
+/// Absence from the ranking is what marks them new.
+///
+/// `user_id ASC` breaks score ties deterministically, so two runs over the same
+/// graph agree.
+pub fn get_trust_ranked_user_ids() -> Query {
+    Query::new(
+        "get_trust_ranked_user_ids",
+        "
+        MATCH (u:User)
+        WHERE coalesce(u.trust, 0.0) > 0
+          AND u.name <> $deleted AND NOT coalesce(u.deleted, false)
+        RETURN u.id AS user_id
+        ORDER BY u.trust DESC, user_id ASC
+        ",
+    )
+    .param("deleted", USER_DELETED_SENTINEL)
+}
+
 /// Users whose profile carries any of the given tag labels, scored by distinct
 /// tagger count summed across the searched labels.
 pub fn search_users_by_tags(labels: &[String], skip: Option<usize>, limit: Option<usize>) -> Query {
--- a/nexus-common/src/db/graph/queries/get.rs
+++ b/nexus-common/src/db/graph/queries/get.rs
@@ -301,11 +301,11 @@ pub fn get_user_tag_pairs() -> Query {
 /// Users carrying positive trust, highest first: the ranked population behind
 /// the social graph badge.
 ///
-/// `coalesce` rather than `IS NOT NULL` because GDS may write `0.0` or leave
-/// the property unset for a node unreachable from the seed set, and both mean
-/// the same thing. Only positive scores are ranked: everyone at zero shares one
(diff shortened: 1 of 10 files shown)


nexus-common/src/models/user/social_graph.rs
```diff
@@ -0,0 +1,208 @@
...
+            .iter()
+            .map(|member| (USER_SOCIAL_GRAPH_KEY_PARTS.as_slice(), member.as_slice()))
+            .collect();
+
+        // The population sizes the `established` cut, so it is read alongside
+        // the ranks rather than baked into them: tuning the cut needs no rebuild.
+        let (population, ranks) = tokio::try_join!(
+            Self::index_sorted_set_card(&USER_SOCIAL_GRAPH_KEY_PARTS, None),
+            Self::check_sorted_set_members(None, &pairs),
+        )?;
```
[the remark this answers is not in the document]
SHAcollision: Good catch, this was real. Fixed in 4248a13.

The count and the positions were two separate reads on two connections, so they could land on opposite sides of the rebuild. Now they go out as one MULTI/EXEC, ZCARD first then the ZSCOREs, so a response always comes from a single snapshot. Ends up being one round trip instead of two as well.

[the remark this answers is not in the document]
SHAcollision: Good catch, this was real. Fixed in 4248a13.

The count and the positions were two separate reads on two connections, so they could land on opposite sides of the rebuild. Now they go out as one MULTI/EXEC, ZCARD first then the ZSCOREs, so a response always comes from a single snapshot. Ends up being one round trip instead of two as well.

SHAcollision:
@greptile re-review pls

nexus-common/src/models/user/social_graph.rs
```diff
@@ -0,0 +1,202 @@
...
+        let mut ranked: Vec<String> = Vec::new();
+        while let Some(row) = rows.try_next().await.map_err(GraphError::from)? {
+            ranked.push(row.get("user_id")?);
+        }
+
+        // Empty drops the key, so readers fall back to "unavailable" rather
+        // than "everyone is new".
+        Self::replace_index_sorted_set(
+            &USER_SOCIAL_GRAPH_KEY_PARTS,
+            &Self::rank_elements(&ranked),
```
aintnostressin, line 118: We are already removing unbunded reads from other parts of the nexus so maybe it's also worth to consider batching it here.

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -298,6 +298,31 @@ pub fn get_user_tag_pairs() -> Query {
...
+/// Absence from the ranking is what marks them new.
+///
+/// `user_id ASC` breaks score ties deterministically, so two runs over the same
+/// graph agree.
+pub fn get_trust_ranked_user_ids() -> Query {
+    Query::new(
+        "get_trust_ranked_user_ids",
+        "
+        MATCH (u:User)
+        WHERE coalesce(u.trust, 0.0) > 0
```
aintnostressin: could `u.trust > 0` be enough here ?

aintnostressin (approved):
Nicely done 🙌 

I just added two small comments - nothing blocking at this point.

nexus-common/src/db/graph/queries/get.rs
SHAcollision: Yeah you are right, simpler. Checked it against the stack and it behaves the same: `null > 0` is null so an unset property drops out exactly like a zero does. And the coalesce would have killed any index on the property later. Changed in 26cd90d.

nexus-common/src/models/user/social_graph.rs
SHAcollision: Good point, and you are right that it is the odd one out. The graph read is already streamed, but I still collect the ids before writing, so it is not flat.

Reason it is not a quick fix: the swap is one MULTI so readers never catch a half written ranking, and that needs the full payload in hand. Streaming the writes means building into a staging key and renaming it at the end, which is roughly what I had earlier in this PR and took out for simplicity. Happy to bring it back, just did not want to grow this one more since the ranked set is empty until we set the seeds anyway.

Tracked in #1050 so it does not get lost. Leaving this open, close it if you are happy with that.

PR #1052, opened 2026-09-03
feat: influencers job

aintnostressin:
Adds three jobs to recompute influencers cache.

Currently cache is populated during request processing which may directly affects request time. 
After moving computation to job requests almost always skip this step.


aintnostressin:
@greptile review

aintnostressin:
@greptile review

SHAcollision (approved):
Looks good overall. Two things:

1. It conflicts with main now that #975 is in, and I think the conflict is not only textual. #975 makes the all_time cache miss fetch from graph and write into the live sorted set, and here that path becomes an early `return Ok(None)`. So when rebasing we need to keep the all_time recovery from #975 and only do the `CacheTimeframe` early return for the ranged ones. Can you double check that after the rebase?
2. Small one: the job keys are `influencers_cache_today` etc but the existing one is `trust-recompute`. Maybe use dashes for consistency? Renaming later would break configs, so better now. wdyt?

aintnostressin:
Recovery happens only on key absence. If watcher write lands before first request that populates the cache then it never reseeds for this logic.

PR #1061, opened 2026-09-09
feat(common): centralize RediSearch schema bootstrap in setup_cache

aintnostressin:
## Summary

`setup_graph()` is the single bootstrap point for Neo4j constraints and indexes, called from `Neo4jConnector::init`. Redis had no equivalent: the `postContentIdx` RediSearch index was created only by migrations, so `MockDb::sync_redis` and the bench setup hand-patched it in, while `db clear --yes` and the reindex bench loop flushed Redis without recreating it and silently lost full-text search.

This PR adds `nexus_common::db::setup_cache()` as the Redis counterpart and deletes the hand-patching.

- **`nexus-common/src/db/kv/setup.rs`** (new): `setup_cache()` is the single declaration of the search schema; future RediSearch indexes go here. It fetches a connection from the global connector and delegates to `pub(crate) setup_cache_on(conn)`, which issues the `FT.CREATE`. Deliberately **not** wrapped in a `OnceCell`: `MATCH (n) DETACH DELETE n` preserves Neo4j indexes so `setup_graph` can run once per process, but `FLUSHDB` destroys the FT index along with the keys, so `setup_cache` must be re-runnable after every flush. The doc comment states the boot requirement: Nexus needs a Redis with the query engine (docker-compose pins `redis:8.0.6-alpine`), and a failed `FT.CREATE` is fatal at connector init. A Redis without it stops every binary from starting rather than degrading search. This is a deliberate choice.
- **`RedisConnector::init`**: after the PING, applies the schema with `setup_cache_on` over a connection from the freshly built pool, **before** `REDIS_CONNECTOR.set(...)`. A failed `FT.CREATE` therefore leaves the global slot empty, so a later `StackManager::setup` with a different URI can still succeed in the same process. This mirrors `setup_graph()` in the Neo4j connector, minus the set-then-fail window.
- **`clear_redis()`**: re-applies the schema immediately after `FLUSHDB`, returning the flush connection to the pool first. This is the only flush path in the workspace, so it fixes `sync_redis`, `clear_database` and the reindex bench at once.
- **`nexus-webapi/src/mock.rs`, `benches/setup.rs`**: removed the hand-patched `create_post_content_index()` calls and the TODO. The bench setup now `.expect`s `StackManager::setup` instead of discarding its result, since the removed unwrap was what used to surface a broken Redis before any bench flushed the cache.
- **`post_content_index_author_setup_1780531200.rs`**: the v2 migration previously called the *shared* create/drop helpers, so it would silently start producing a v3 index the day a field is added to `setup_cache`. Its `FT.CREATE` / `FT.DROPINDEX` arg lists are now inlined and frozen, the same way the v1 migration already does. Both migrations stay registered; they are idempotent and still needed by deployed environments.
- **`kv/index/search.rs`**: `drop_post_content_index` is removed (-23 lines); its only caller was the v2 migration, which now carries a frozen copy. `ft_create_post_content_index` takes the connection explicitly so the connector can run it on its own pool. The `FT.CREATE` arg list is unchanged, but it now has a warning comment above it: adding or changing a field requires a matching index migration, because the frozen v2 migration would otherwise drop the new index and recreate v2 on a fresh environment during `nexusd db migration run`.
- **Public API break** for `nexus-common`: `pub nexus_common::models::post::drop_post_content_index` and `create_post_content_index` is removed.
- **`nexusd/tests/cache_setup/main.rs`** (new): `clear_redis_recreates_post_content_index` pins the no-`OnceCell` behavior. It asserts via `FT.INFO` that `postContentIdx` indexes `$.content`, `$.author`, `$.kind` over the `PostDetails::prefix()`-derived prefix both after stack setup and again after `clear_redis()`, then rebuilds the cache with `reindex::sync()` so a local run leaves the mock data in place. It lives in its own binary and is the only nexusd test that flushes Redis.
- **`.config/nextest.toml`**: the `cache_setup` binary gets `threads-required = 'num-test-threads'` so its `FLUSHDB` never overlaps with any other test in the run, including nexus-webapi's search suite under `--workspace`. A `[test-groups]` entry would not help: group limits only apply between members of the same group, so a one-test group protects nothing.

Untouched: `import_migrations`, the `Migration` trait, and the v1 migration.

Note: the reindex bench now runs with a live `postContentIdx` over `Post:Details:`, since `MockDb::drop_cache` recreates it on every iteration. Previously the first `FLUSHDB` destroyed the index and all later samples ran un-indexed. Its numbers are not comparable to previous runs.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo check --workspace --all-targets`
- [x] Dropped `postContentIdx` by hand, ran `nexusd db mock`: index recreated at connector init and again right after `FLUSHDB` before reindex
- [x] `FT.INFO postContentIdx` lists `content`, `author`, `kind` with prefix `Post:Details:`
- [x] `cargo nextest run -p nexus-webapi search --no-fail-fast`: 64 passed
- [x] `cargo nextest run -p nexusd -E 'binary(cache_setup)'`: 1 passed
- [x] `cargo nextest list --workspace -E 'package(nexusd) and binary(cache_setup)'` selects exactly the one test; nextest rejects an invalid `threads-required` value at config parse, so the override is honored
- [x] `nexusd db clear --yes` then `FT.INFO postContentIdx`: index still exists with all three fields (previously "Unknown index name")

### DIFF
--- a/nexus-common/src/db/kv/index/search.rs
+++ b/nexus-common/src/db/kv/index/search.rs
@@ -1,6 +1,7 @@
 use crate::db::config::FT_SEARCH_TIMEOUT_MS;
 use crate::db::get_redis_conn;
 use crate::db::kv::error::{RedisError, RedisResult};
+use deadpool_redis::Connection;
 use std::sync::OnceLock;
 use tracing::warn;
 
@@ -18,9 +19,13 @@ fn ft_search_timeout_ms() -> usize {
 /// Creates the post content index: $.content TEXT + $.author TAG CASESENSITIVE + $.kind TAG CASESENSITIVE.
 /// NOOFFSETS/NOHL kept; NOFIELDS dropped to allow field-targeted queries.
 /// Idempotent: short-circuits on "already exists".
-pub(crate) async fn ft_create_post_content_index(prefix: &str) -> RedisResult<()> {
-    let mut conn = get_redis_conn().await?;
-
+///
+/// Takes the connection explicitly so the connector can apply the schema on
+/// its own pool before registering itself globally.
+pub(crate) async fn ft_create_post_content_index(
+    conn: &mut Connection,
+    prefix: &str,
+) -> RedisResult<()> {
     // Adding or changing a field here requires a matching index migration.
     // PostContentIndexAuthorSetup1780531200 drops and recreates from a frozen
     // v2 copy, so a fresh environment would boot on this schema and then be
@@ -49,7 +54,7 @@ pub(crate) async fn ft_create_post_content_index(prefix: &str) -> RedisResult<()
         .arg("kind")
         .arg("TAG")
         .arg("CASESENSITIVE")
-        .query_async::<()>(&mut conn)
+        .query_async::<()>(conn)
         .await;
 
     match result {
--- a/nexus-common/src/db/kv/index/search.rs
+++ b/nexus-common/src/db/kv/index/search.rs
@@ -55,29 +55,6 @@ pub(crate) async fn ft_create_post_content_index(prefix: &str) -> RedisResult<()
     }
 }
 
-/// Drops the post content index without deleting the underlying documents.
-/// Idempotent: swallows "Unknown index name" so repeated calls are safe.
-pub(crate) async fn drop_post_content_index() -> RedisResult<()> {
-    let mut conn = get_redis_conn().await?;
-
-    let result = deadpool_redis::redis::cmd("FT.DROPINDEX")
-        .arg("postContentIdx")
-        .query_async::<()>(&mut conn)
-        .await;
-
-    match result {
-        Ok(()) => Ok(()),
-        Err(e) => {
-            let msg = e.to_string().to_lowercase();
-            if msg.contains("unknown index name") || msg.contains("no such index") {
-                Ok(())
-            } else {
-                Err(RedisError::CommandFailed(e.to_string().into()))
-            }
-        }
-    }
-}
-
 /// Assembles the RediSearch query string from a content fragment, optional author
 /// filter, and optional kind filter.
 ///
--- a/nexus-common/src/db/kv/index/search.rs
+++ b/nexus-common/src/db/kv/index/search.rs
@@ -29,7 +29,7 @@ pub(crate) async fn ft_create_post_content_index(
     // Adding or changing a field here requires a matching index migration.
     // PostContentIndexAuthorSetup1780531200 drops and recreates from a frozen
     // v2 copy, so a fresh environment would boot on this schema and then be
-    // silently downgraded to v2 by `nexusd db migration run`. See issue #N.
+    // silently downgraded to v2 by `nexusd db migration run`.
     let result = deadpool_redis::redis::cmd("FT.CREATE")
         .arg("postContentIdx")
         .arg("ON")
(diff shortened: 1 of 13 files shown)


aintnostressin:
@greptile review

aintnostressin:
@greptile review again

nexusd/tests/cache_setup/main.rs
```diff
@@ -0,0 +1,104 @@
...
+//! with the keys, so `setup_cache` must run again on every flush rather than
+//! once per process. Wrapping it in a `OnceCell` (the way `setup_graph` is)
+//! would pass every other test, because they boot the stack once and never
+//! flush, while silently breaking `nexusd db clear --yes` and the reindex
+//! bench. This test pins that behavior.
+//!
+//! It lives in its own binary and is the only nexusd test that flushes Redis.
+//! The other nexusd integration tests assert on graph state, not Redis, so no
+//! nextest serialization is needed. The cache is rebuilt with
+//! `reindex::sync()` before returning so a local run leaves the mock data
```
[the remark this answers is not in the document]
aintnostressin: Addressed in b20d3e0

nexus-common/src/db/kv/index/search.rs
```diff
@@ -18,9 +19,17 @@ fn ft_search_timeout_ms() -> usize {
...
+/// Takes the connection explicitly so the connector can apply the schema on
+/// its own pool before registering itself globally.
+pub(crate) async fn ft_create_post_content_index(
+    conn: &mut Connection,
+    prefix: &str,
+) -> RedisResult<()> {
+    // Adding or changing a field here requires a matching index migration.
+    // PostContentIndexAuthorSetup1780531200 drops and recreates from a frozen
+    // v2 copy, so a fresh environment would boot on this schema and then be
+    // silently downgraded to v2 by `nexusd db migration run`. See issue #N.
```
SHAcollision: this "See issue #N" looks like a placeholder that slipped in. maybe link the real issue or just drop that sentence?

nexus-common/src/models/post/search.rs
```diff
@@ -148,23 +147,13 @@ impl PostsByTagSearch {
     }
 }
 
-const POST_CONTENT_INDEX: &str = "postContentIdx";
-
 /// Creates the post content full-text index: $.content TEXT + $.author TAG CASESENSITIVE + $.kind TAG CASESENSITIVE.
 /// Includes NOOFFSETS/NOHL; NOFIELDS dropped to allow field-targeted queries.
 /// Idempotent: no-ops if the index already exists.
 pub async fn create_post_content_index() -> RedisResult<()> {
```
SHAcollision: I think this one has no callers anymore after this PR. since we already broke the public api by removing the drop helper, maybe we remove this too? otherwise we have two public ways to declare the same index and the "single declaration" idea gets a bit blurry. wdyt?

SHAcollision (approved):
LGTM, nice one.

Two small things inline, not blockers. Feel free to merge after that.

nexus-common/src/models/post/search.rs
aintnostressin: done 8f4a80f

nexus-common/src/db/kv/index/search.rs
aintnostressin: done 8f4a80f

PR #1065, opened 2026-09-12
feat(watcher): order homeserver polling by aggregate hosted trust

aintnostressin:
## What

Order homeserver polling by aggregate hosted trust instead of raw active-user count.

`get_all_homeservers_with_active_users` sorted purely by `count(u)`, and `pre_run` truncates that list to `monitored_homeservers_limit` (50 by default). Raw user count therefore decided which homeservers were polled **at all**, so stuffing a homeserver with cheaply-minted keys bought polling priority ahead of homeservers hosting well-connected users.

```cypher
WITH hs.id AS id,
     sum(coalesce(u.trust, 0.0)) AS hosted_trust,
     count(u) AS active_users
ORDER BY hosted_trust DESC, active_users DESC, id ASC
```

Trust is a property of a user's position in the follow graph rather than of their host, so freshly minted keys are unreachable from the seed set and sum to `0.0` however many of them there are.

## Why it is safe by default

`coalesce` keeps the default install unchanged: `[trust_rank] seed` ships empty, so every sum is `0.0` and the ordering falls through to the existing `active_users` sort — byte-for-byte today's behaviour. The trust term only starts doing work once an operator configures a seed set.

The `id` tiebreak makes truncation deterministic. Without it, two homeservers with equal trust and equal user counts could swap places between runs and change which ones fall outside the limit.

## Deliberately not changed

PKDNS resolution order. Trust-ordering there would destroy `bisection_order_user_pks`, whose comment cites up to 95% resolution failure without its anti-hotspot ordering. Trust-tier the TTL, never the order.

## Known gap (please read before merging)

The truncation is still a hard cut, so **ordering decides membership, not merely order**. Two consequences:

- Against the obvious attack this change helps: an attacker's zero-trust homeservers now sort *below* every homeserver hosting any trusted user, where under `count(u) DESC` the same stuffing puts them at the top and evicts real homeservers.
- But it introduces a new eviction path: a large, entirely legitimate homeserver whose community is not yet seed-reachable now sorts below a tiny one hosting a single trusted user, and at a hard cut of 50 it can drop off the list entirely.

Before this ordering is safe to run at the cap, `pre_run` needs a round-robin floor — reserve part of the 50 so the ranking sets order *within* a budget rather than membership *in* it. Once every homeserver is eventually polled, a gamed ranking only means the attacker is polled sooner, which is not an attack. That floor is deliberately not in this PR.

## Testing

- New: `test_event_processor_runner_orders_homeservers_by_hosted_trust` — two ranked users on one homeserver against eight unranked on another, so the crowded one wins on raw count. **Verified it fails without the query change** (negative control), so it is real coverage rather than an accidental pass.
- The test builds its own graph: no fixture user has a `HOSTED_BY` edge, so every homeserver in the shared fixture aggregates to `0.0` and cannot exercise this.
- It deletes the users it creates. `Sorted:Users:SocialGraph` is one global key built from `WHERE u.trust > 0`, and a stray scored test user ties the fixture's top user and wins the `id ASC` tiebreak, breaking `nexus-webapi`'s `test_social_graph_status`.
- `nexus-common`: 142 passed / 0 failed, stable across three runs.
- `cargo fmt --check` and clippy clean.

### CI caveat, pre-existing and unrelated to this change

The `nexus-webapi` suite is not green on this branch, and it is not green on `main` either — it just fails on a different side.

`tests/utils/server.rs` has two `OnceCell` test servers (`TEST_SERVER` and `TEST_SERVER_WITH_KEY_REPUBLISHER`), each building its **own fresh TempDir keypair**, so they produce different `StackConfig`s. `StackManager` is a process-global that rejects the second one, so whichever server initialises second panics at `server.rs:85` with `StackManager already initialized with a different StackConfig`, taking every test that needs it with it.

On `main` the republisher server loses and 12 tests fail. On this branch the plain server loses and 432 fail. The two sets are exactly complementary, which is the signature. A comment-only recompile does not flip it, and nothing in `nexus-webapi` calls the query changed here — this binary just perturbs a genuine race.

Worth fixing in its own PR (share one config/keypair between the two servers, or initialise `StackManager` once); it would likely take the 12 long-standing failures to zero.


aintnostressin:
@codex review

tipogi (approved):

PR #1068, opened 2026-09-15
feat(watcher): export metrics for the external HS monitoring limit

talosmachina:
Requested by @aintnostressin: the deployed instance should be alerted on before `monitored_homeservers_limit` starts dropping external homeservers, so both the cap in force and the number of homeservers a run indexes need to be exported.

## What this adds

`KeyBasedEventProcessorRunner::pre_run` records two gauges on every external-HS run:

- `watcher.external_hs.monitored_limit` - the configured cap in force.
- `watcher.external_hs.indexed` - external homeservers the run selected: the active ones, minus the primary HS and blacklisted ones, truncated to the limit.

`indexed / monitored_limit` reaches 1 exactly when the cap is what bounds coverage, so that ratio is the signal to alert on. With the default limit of 50, a deployment stays quiet until roughly 45 active external homeservers.

The second gauge counts the homeservers a run selected, not the ones that finished Ok. That is deliberate: only the selected count can reach the limit, while a HS in backoff or in error would otherwise look like headroom that does not exist. If the count of successful runs is wanted instead, it is `RunAllProcessorsStats::count_ok` in `post_run`.

Also included: a Monitoring section under `monitored_homeservers_limit` in `docs/decentralization.md`. No alert rule ships here: the gauges are enough for now.

## Verification

- `cargo test -p nexus-watcher --lib` - 46 passed, including a unit test that records a run and reads both gauges back from an `InMemoryMetricExporter`.
- `cargo fmt -p nexus-watcher`.


docker/otel/alerts.yaml
```diff
@@ -31,3 +31,14 @@ groups:
         annotations:
           summary: "Nexus watcher cursor is stalled"
           description: "The primary homeserver cursor did not advance during the last 15 minutes."
+
+      - alert: NexusExternalHomeserversNearMonitorLimit
```
aintnostressin: We don't need this now.

aintnostressin: @talosmachina

talosmachina: Removed in `eb50a7d`: the rule is gone from `docker/otel/alerts.yaml`, along with the docs paragraph that pointed at it. The two gauges stay, so the ratio is still readable and a rule can be added back later. If you meant the metric export itself is not needed now, say so and I will drop the rest of the PR.

aintnostressin:
@greptile review

docs/decentralization.md
```diff
@@ -79,6 +79,19 @@ endpoint (hence "key-based" — keyed on each user's pubky). Configured in
...
+
+| Metric | Type | Meaning |
+| --- | --- | --- |
+| `watcher.external_hs.monitored_limit` | gauge | The cap in force. |
+| `watcher.external_hs.indexed` | gauge | External HSs the last run selected for indexing: the active HSs, minus the primary HS and blacklisted ones, truncated to the limit. |
+
+`indexed / monitored_limit` is the saturation ratio: it reaches `1` exactly when
+the cap is what stops the least trusted external HSs from being indexed, which is
+the moment to raise the limit. Gauges keep their last value while the process is
+alive, so a `monitored_homeservers_limit` of `0` exports a zero denominator.
```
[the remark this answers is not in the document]
aintnostressin: Cut off homeservers are not `indexed` so it reflects true value. It doesn't matter if and how many homeservers were skipped.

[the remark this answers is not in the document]
talosmachina: Correct, and fixed in `7fe8e18`. The section now says `1` means the eligible set filled the cap, so the limit is what bounds coverage from there, and that an exactly-filled set and a truncated one both read `1`; the earlier sentence claimed `1` was proof that homeservers were dropped. The `EXTERNAL_HS_METRICS` doc comment in `key_based.rs` carried the same wording and is updated with it.

talosmachina:
@greptile-apps review

tipogi (approved):
LGTM

PR #1069, opened 2026-09-16
feat: hot tags cache jobs

tipogi:
## Summary
Global hot tags were filled from Neo4j on a cache miss, on the request path. This moves that scan onto per-timeframe daemon jobs, the same shape as the influencers cache jobs (#1052).
- Fill is always the top 100 post tags / 20 taggers. Request `skip` / `limit` / `taggers_limit` apply on re-read of the snapshot, not the graph result. Empty `Some([])` is a hit, not a miss.
- All four `Timeframe`s are TTL caches here, including AllTime, so AllTime has its own job (`hot-tags-cache-all-time`). Influencers AllTime stays a live index and does not.
- `[jobs.hot-tags-cache-*]` sections ship commented. Cron minutes are offset from influencers so the two sets do not collide. Uncommenting them on a build that predates these jobs will refuse to start.
- A global `skip` past the snapshot is an empty page, not an error: `skip` of 100, 101 and 10 000 all return `200 []`, so a client paging by `limit` terminates the same way at every `limit`. A cold-cache read at `skip >= 100` short-circuits instead of scanning, so the offset cannot be walked to force scans. Reach-scoped queries still page on the graph.
- `taggers_count` is the graph's `COUNT(DISTINCT user.id)`, which is unbounded, while `taggers_id` is a sample capped at 20. The cache stored only the sample, so a rebuilt response reported `min(count, 20)` for every tag. Taggers are now cached as `CachedTaggers { taggers, total }` and the read reports `total`.

## Pre-submission Checklist


talosmachina (commented):
**No findings.** Moves the global hot-tags graph scan off the request path into per-timeframe `nexusd` cache jobs, adds the four commented `[jobs.hot-tags-cache-*]` sections, caps the global `skip` at the 100-entry cache size, and switches the cache write to an atomic sorted-set replace. Reviewed `4898797`.

<details><summary>What I checked, and 7 candidates I ruled out</summary>

**Read in full**: `nexus-common/src/models/tag/stream.rs`, `nexus-common/src/models/tag/global.rs`, `nexus-webapi/src/routes/v0/tag/global.rs`, `nexusd/src/jobs/hot_tags.rs`, `nexusd/src/jobs/mod.rs` (`scheduled_jobs`), `nexus-common/src/db/kv/index/sorted_sets.rs` (`get_range`, `replace`), `nexus-common/src/config/daemon.rs`, `nexus-common/default.config.toml`, `nexus-webapi/tests/tags/hot.rs`
**Call sites traced**: `HotTags::get_hot_tags`, `fetch_and_cache`, `write_or_preserve_cache`, `Taggers::get_from_global_timeline`, `Bootstrap::add_global_hot_tags`, `db/reindex.rs`, `nexus-webapi/benches/tag.rs`
**CI**: green (format, lint, test, Greptile)

**Ruled out**
- *Empty graph result leaves no key, so every global request re-scans*: the old code had the same `if !hot_tags.is_empty()` guard, so the miss-on-every-request path is not introduced here.
- *The job and the request-path miss can write the cache concurrently and interleave taggers/scores*: the scores write is now an atomic replace, and the two-writer race predates this PR.
- *`skip > 100` returns 400 while the OpenAPI bound stays 10 000*: deliberate, and named in the `skip` param description; the shared route also serves the uncapped reach path, so the type bound cannot be narrowed.
- *AllTime is a 24h TTL cache here while influencers AllTime is a live index*: intentional, with its own job and a 12h cadence under the TTL.
- *`fetch_and_cache` hardcodes `Some(TaggedType::Post)`*: the API route already passed `Post`, and the cache read never consults `tagged_type`, so the bootstrap `None` to `Some(Post)` edit changes nothing the global path returns.
- *`taggers_count` capped at the cached map size*: pre-existing on this read path.
- *Equal-score tie order*: Redis breaks ties by member descending against the graph's `label ASC`; the read path is unchanged.
</details>

aintnostressin (approved):

aintnostressin (approved):

PR #1074, opened 2026-09-17
feat(search): filter posts by tag by the user's reach

aintnostressin:
## Reach filter for `GET /v0/search/posts/by_tag/{tag}` (1/3)

First of three stacked PRs.

Adds optional `user_id` + `reach` query params (`followers` | `following` | `friends` | `wot` | `wot_1`..`wot_3`, bare `wot` = depth 2). They must be given together, otherwise the request fails with 400. These are the same names and values used by `tags/hot`, `tags/taggers` and `stream/users`.

With a reach, results come from the graph via the existing `post_stream` query.

**Reach-mode semantics**
- The user's own posts are excluded.
- An unknown `user_id` returns `200 []`.
- Only parent posts are returned.
- `total_engagement` doesn't count mentions.
- `start`/`end` cursors don't carry over from unfiltered mode.

**Other changes**
- For streams with an observer, `post_stream` filters tags with an `EXISTS` semi-join instead of a per-tagger `MATCH`, with unchanged results. Streams without an observer (for example `stream/posts?tags=a,b` with `source=all`) keep the `MATCH`, so the planner can still start from the tag-label index.
- Search requests record WoT metrics under `source=search_wot`, apart from the post stream's.
- `StreamReach` is documented in OpenAPI as the string enum it accepts.
- New `search-reach.cypher` fixture, shared by all three PRs.

**Performance:** cost grows with the number of posts written by users in the reach, so `wot_2` and `wot_3` are the heaviest cases. `stream/posts?source=wot` already runs this same traversal.

**Tests**
- Every reach type, including the observer being excluded and a follow cycle.
- Engagement sort with `start`/`end`/`skip` pagination.
- Unknown user, and 400s for invalid params.
- Unit tests for `from_reach`, the tag filter form (`EXISTS` only with an observer), the metrics label and the `StreamReach` schema.


aintnostressin:
@greptile review

talosmachina (commented):
**No findings.** Adds optional `user_id` + `reach` (`followers` | `following` | `friends` | `wot` | `wot_1`..`wot_3`) to `GET /v0/search/posts/by_tag/{tag}`; with both set the results come from the graph through the existing `post_stream` query, and the plain path is unchanged. Reviewed `2dd5b611`, full tier.

<details><summary>What I checked, and 6 candidates I ruled out</summary>

**Read in full**: `models/post/stream.rs`, `models/post/search.rs`, `db/graph/queries/get.rs` (`post_stream`, `build_query_with_params`, the semi-join test), `models/post/metrics.rs`, `types/mod.rs`, `routes/v0/search/posts.rs`, `routes/v0/search/mod.rs`, `routes/v0/endpoint` wiring, `middlewares/rate_limit.rs`, `routes/mod.rs`, the new fixture and all four touched test files.
**Call sites traced**: `post_stream` has one production caller, `get_scored_from_graph` (`stream.rs:487`), reached by `get_from_graph` (every graph-backed stream) and by the new `get_scored_post_keys` (search). `PostsByTagSearch::get_by_label` (the index path) is untouched.
**CI**: green (format, lint, test 10m32s, Greptile). `db mock` imports the new fixture, so all four suites ran against it.

**Ruled out**
- *The `EXISTS` semi-join changes `stream/posts?tags=`*: same row set. Both forms require a `:User` TAGGED edge with `tag.label IN $labels`, and both collapse the old MATCH's per-tagger rows in `WITH DISTINCT p, author` before `SKIP`/`LIMIT`; `$labels` binding is unchanged (`get.rs:1354`). The WoT-domain tag tests exercise the same query.
- *`from_scored_entries` changes `last_post_score`*: both score sources (i64 `indexed_at`, small engagement counts) round-trip through f64 exactly, and `entries.last()` is the old `last_post_score`.
- *Reach mode leaks replies or the observer's own posts*: `WHERE author.id <> $observer_id` is in every reach arm and the parents-only filter covers all non-bookmark sources; the fixture's tagged friend reply and the follow cycle make both tests real.
- *`start`/`end` mean something else than in the index path*: `try_from_index_sorted_set` passes `(end, start)` to `get_range`, so `start` is the maximum in both. Only the score base differs, which the PR documents.
- *Plain tag search drops 300/min -> 20/min*: deliberate, stated in the PR body, and the new rate-limit test pins the route to the expensive bucket.
- *`user_id` reaching Cypher*: bound as `$observer_id` through `query.param`.

</details>

Merge confidence: 5/5, CI green, files read whole, and every new path has an integration test.

SHAcollision (commented):
nice work, the wiring is clean and reusing `post_stream` makes sense. I left some inline comments, 2 of them I think we should sort out before merging:

- own posts: #1071 goes the other way ("Me + reach"), so we need to pick one rule
- the `EXISTS` change might make `stream/posts?tags=` slower when there is no observer

the rest are small things, up to you. wdyt?

nexus-webapi/tests/post/search.rs
```diff
@@ -423,3 +430,180 @@ async fn test_post_search_skip_beyond_range() -> Result<()> {
...
+    assert_eq!(post_keys(&wot_2), expected);
+
+    // Bare `wot` is depth 2
+    let wot = get_request(&tag_reach_url(&format!("user_id={OBS}&reach=wot"))).await?;
+    assert_eq!(post_keys(&wot), expected);
+    Ok(())
+}
+
+#[tokio_shared_rt::test(shared)]
+async fn test_post_search_by_tag_reach_excludes_observer() -> Result<()> {
```
SHAcollision, line 524: I am not sure we want to exclude the observer here. #1071 changes the wot arm to `FOLLOWS*0..` so a reach reads as "Me + reach", which is what the app folks asked for. these 2 PRs merge without conflict but whichever lands second breaks the wot tests of the other one, and after that search would include own posts for `wot` but not for `following`/`followers`/`friends` 🤔 maybe we align on one rule for all reach types first?

nexus-common/src/db/graph/queries/get.rs
```diff
@@ -1136,12 +1136,12 @@ pub fn post_stream(
...
 
-    // Apply tags
+    // Apply tags as a semi-join: a MATCH would emit one row per matching
+    // tagger, only for WITH DISTINCT to collapse them again
     if tags.is_some() {
-        cypher.push_str("MATCH (:User)-[tag:TAGGED]->(p)\n");
         append_condition(
             &mut cypher,
-            "tag.label IN $labels",
+            "EXISTS { MATCH (:User)-[tag:TAGGED]->(p) WHERE tag.label IN $labels }",
```
SHAcollision: I think this can hurt `source=all`. with the old `MATCH ... WHERE tag.label IN $labels` the planner could start from `taggedLabelIndex`, but it can't start from inside an `EXISTS`, so it scans every post first and checks tags after. main sends multi-tag (`tags=a,b`) and tags + `kind` through this graph path, and the app uses both. for the reach sources `EXISTS` is fine because we start from the reach. could you run a PROFILE of `stream/posts?tags=a,b` on the bench graph? if it's slower maybe keep `EXISTS` only when there is an observer

nexus-webapi/src/routes/v0/search/mod.rs
```diff
@@ -25,6 +25,13 @@ pub fn expensive_routes() -> Router<AppState> {
             SEARCH_USERS_BY_TAGS_ROUTE,
             get(users::search_users_by_tags_handler),
         )
+        // With `user_id` + `reach` the search runs a graph traversal instead of
+        // a sorted-set read. Buckets apply per route, so the plain tag search
+        // shares the stricter limit
+        .route(
+            SEARCH_POSTS_BY_TAG_ROUTE,
```
SHAcollision: small one: the plain tag search is just a redis read and now gets 20/min, while `stream/posts?source=wot&tags=` runs this same traversal and stays in the default bucket. maybe keep this route in default for now so both match, and move them together later if the traversal turns out costly? not a blocker

nexus-common/src/models/post/metrics.rs
```diff
@@ -84,7 +87,18 @@ impl WotStreamMetrics {
 
 static METRICS: LazyLock<WotStreamMetrics> = LazyLock::new(WotStreamMetrics::new);
 
-/// Count one WoT post-stream request. `source` is `"wot"` or `"wot_domain"`.
+/// The `source` label for a search that runs the WoT stream query, so search
+/// traffic stays out of the post stream's series.
+pub(super) fn search_source(source: &'static str) -> &'static str {
+    match source {
+        "wot" => "search_wot",
+        "wot_domain" => "search_wot_domain",
```
SHAcollision: nit: `from_reach` never gives `WotDomain`, and non-wot sources never get here, so the `search_wot_domain` and `search` arms never run. we could drop them and just map `wot` to `search_wot`

nexus-webapi/tests/post/search.rs
aintnostressin: There is no decision yet and what I found even Talos PR is not correct - i think it doesn't align with any presented option. We either block it here, or proceed with main rules, then #1071 will need to update both places.

talosmachina: @aintnostressin to be exact about what #1071 changes, in case it matters for the rule you pick: the only edit is the `wot` arm of `post_stream`. `MATCH (observer)-[:FOLLOWS*1..{depth}]->(author:User) WHERE author.id <> $observer_id` becomes `MATCH (observer)-[:FOLLOWS*0..{depth}]->(author:User)`, so the observer binds as an author through the zero-length path and the reach is the observer plus their network at any depth; `WITH DISTINCT author` still collapses the duplicate the follow cycle would produce. `following`, `followers`, `friends` and `wot_domain` are untouched and keep their current rule, so this is "Me + reach" for `wot` only.

If your read of that diff turned up a case where it does not hold, name it and I will check it against the branch.

nexus-common/src/db/graph/queries/get.rs
aintnostressin: changed in [0a1068a](https://github.com/pubky/pubky-nexus/pull/1074/commits/0a1068a32b99a5783699e95d8bd3298bc550a7ee)

nexus-common/src/models/post/metrics.rs
aintnostressin: done [5b33512](https://github.com/pubky/pubky-nexus/pull/1074/commits/5b33512089d5863ebdb1315eec009b625cc89042)

nexus-webapi/src/routes/v0/search/mod.rs
aintnostressin: reverted [5634e4c](https://github.com/pubky/pubky-nexus/pull/1074/commits/5634e4c39f52262a7bf738f96bac1ef8ff5a33c1)

nexus-common/src/models/post/stream.rs
```diff
@@ -293,6 +292,62 @@ impl PostStream {
...
+    /// Returns [`ModelError::GraphOperationFailed`] on graph failures, including
+    /// `GraphError::QueryTimeout` when the query exceeds its budget.
+    pub(crate) async fn get_scored_post_keys(
+        source: StreamSource,
+        pagination: Pagination,
+        order: SortOrder,
+        sorting: StreamSorting,
+        tags: Option<Vec<String>>,
+    ) -> ModelResult<Vec<(String, f64)>> {
+        let wot = Self::wot_dimensions(&source)
```
tipogi, line 311: `source=search_wot` only applies to `wot.post_stream.*`. The query below is still built as `Query::new("post_stream")` with `telemetry_attr("source", "wot")` (`nexus-common/src/db/graph/queries/get.rs` ~1314), so `neo4j.query.duration{query=post_stream,source=wot}` (and errors/slow) will count tag-search `wot_*` as home-feed load.

Example: feed `source=wot` in 80ms + search `reach=wot_3` in 2s → Neo4j p95 for `source=wot` jumps even though the feed did not get slower. A search timeout also increments feed errors.

The unit test only checks `SEARCH_WOT_SOURCE != "wot"`. It does not look at the Cypher query attrs.

Could search pass `source=search_wot` (or a distinct query label) into `post_stream`?

tipogi:
@greptile can you focus in the cypher query in that review, I am not sure if it is the right direction. go!

nexus-common/src/models/post/stream.rs
aintnostressin: done 5b33512089d5863ebdb1315eec009b625cc89042

tipogi (approved):

PR #1075, opened 2026-09-18
ref: tidy `nexusd` into runtime, jobs, and migrations

tipogi:
## Summary
Tidy `nexusd` so the daemon host matches how it actually runs: **runtime** starts the process, **jobs** and **migrations** each have an engine plus a **catalog** of work, and **trust** stays the PageRank domain.
No scheduling, lock, or migration-phase behavior change. The production job list and migration set are the same.
- Split the jobs engine (`registry` / `runner`) out of `jobs/mod.rs`.
- Move cache jobs and `TrustRecomputeJob` into `jobs/catalog`. `JobRegistry::catalog` is the production list; `DaemonLauncher` loads config and starts API + watcher + that catalog.
- Rename `migrations/migrations_list` → `migrations/catalog` and register migrations there.
`nexusd` is the daemon, not an embedding library. Tests already build registries with `JobRegistry::new`. This PR does not add launcher injection or a `trust::TrustRecomputeJob` re-export.

## Pre-submission Checklist


talosmachina (commented):
**No findings.** Splits `nexusd`'s job engine (`registry`/`runner`) from the concrete jobs in `jobs/catalog`, moves the daemon start path into `runtime`, and renames `migrations_list` to `migrations/catalog`; behaviour, the production job list and the migration set are unchanged. Reviewed `c551e45`, full tier.

<details><summary>What I checked, and 5 candidates I ruled out</summary>

**Read in whole**: `jobs/mod.rs`, `jobs/registry.rs`, `jobs/runner.rs`, `jobs/catalog/mod.rs`, `runtime/launcher.rs`, `runtime/mod.rs`, `main.rs`, `lib.rs`, `migrations/mod.rs`, `migrations/catalog/mod.rs`, `migrations/manager.rs`, `trust/mod.rs`
**Ran**: `cargo check -p nexusd --all-targets` clean; `cargo test -p nexusd --lib` 78 passed.
**Parity checked**: `JobRegistry::catalog` registers the same eight jobs in the same order as the removed `job_registry` in `main.rs`, so `nexusd jobs list` output is unchanged; `import_migrations` registers the same seven migrations in the same order; all eight files under the old `migrations_list/` are 100% renames (`R100`); `MIGRATION_PATH` and the README point at `catalog/`, and `grep` finds no remaining reference to `migrations_list` anywhere in the repo.
**Test coverage**: 74 to 76 test functions, none dropped; the two additions are `catalog_names_are_the_eight_jobs` and `shipped_default_config_resolves_against_the_catalog`, the latter being the guard that the shipped `[jobs.*]` config still resolves against the catalog.

**Ruled out**
- *Narrowed visibility breaks a consumer*: `scheduled_jobs`, `ScheduledJob` and `jobs::run` go `pub` to `pub(crate)`, and the `DaemonLauncher` registry argument and the `trust::TrustRecomputeJob` re-export are removed, but no workspace crate depends on `nexusd` and there is no other call site in the repo. The PR body states this is the intent.
- *Startup fail-fast ordering changed*: the launcher still reads the config and resolves and validates the scheduled jobs before building the API context, same order as before the move.
- *`db migration new` writes to the old directory*: both the file path and the `mod.rs` append target derive from `MIGRATION_PATH`, now `nexusd/src/migrations/catalog/`.
- *Moved modules break their relative imports*: the two `super::` uses in the moved job files are rewritten to `crate::jobs::...`; the migration renames keep the same module depth.
- *Migration phase behaviour changed*: the moved migration files are byte-identical renames, so no phase code was touched.

**CI**: the `test` job is red on `nexus-webapi` `stream::user::influencers::test_global_influencers_skip_to_cache_end_is_empty_page` (`an empty window must not be treated as a cache miss and rewrite the key, ttl went to 3600`). `nexus-webapi` does not depend on `nexusd`, so this cannot originate in the diff; the same suite also failed a different test in a local full run, so it reads as a pre-existing flake in that shared-Redis suite.

</details>

Merge confidence: 3/5, CI red for a reason I could not tie to the change.

aintnostressin (approved):
good direction!

PR #1076, opened 2026-09-18
refactor(media): move variant generation out of nexus-common into nexus-webapi

aintnostressin:
## Why

Only the API derives file variants (a request for a variant that isn't on disk yet creates it). The watcher just publishes variant URLs. Yet all the conversion machinery lived in `nexus-common`: about 930 lines of non-test code plus 1,300 lines of tests, with a single consumer.

## What moved to `nexus-webapi::media`

- Concurrency gates, the subprocess runner, and the image/video processors
- `VariantController`, with `Blob::get_by_id` folded in as `ensure_variant`
- The content type a derived variant is served as. It now sits next to the format the processor writes, and is derived from it (`image/{IMAGE_FORMAT}`), so the two can't drift apart.

## What stays in `nexus_common::media`

Only what both services must agree on: `FileVariant` and the table of which variants a content type has. The watcher publishes URLs from that table, and the API validates requests against the same one.

## API surface

`nexus-webapi` is published, so the new module exposes as little as possible:

- `ensure_variant` is the only entry point. The existence check and the conversion are private, so they can't be called separately.
- `processors` is crate-private. Only `MediaProcessorError` is re-exported.

## Smaller cleanups

- `ModelError` loses its media variant. The watcher's matching error variant was unreachable and is removed.
- `FileVariant` derives `Copy`, and the files path is passed as `&Path`, so serving an existing variant doesn't allocate.
- `get_options_for_variant` drops an unused `file` argument.

## Testing

`cargo clippy --workspace --all-targets -- -D warnings` is clean. Lib tests pass: webapi 153, common 156, watcher 51. The DB-backed integration tests compile but weren't run.


aintnostressin:
@greptile review

talosmachina (commented):
**No findings.** Moves on-demand variant derivation (concurrency gates, subprocess runner, image/video processors) out of `nexus-common` into `nexus_webapi::media`, leaves only `FileVariant` and the content-type variant table shared, and folds `Blob::get_by_id` into `VariantController::ensure_variant`. Reviewed `6eb7698`, full tier.

<details><summary>What I checked, and 5 candidates I ruled out</summary>

**Read in full**: `nexus-webapi/src/media/mod.rs`, `media/processors/mod.rs`, `media/processors/image.rs`, `media/processors/video.rs`, `error.rs`, `routes/static/serve_dir.rs`, `routes/static/avatar.rs`, `routes/static/files.rs`, `nexus-common/src/media/mod.rs`, `models/file/blob.rs`, `nexus-watcher/src/events/handlers/file.rs`
**Call sites traced**: every `Blob::get_by_id`, `VariantController::*`, `MediaProcessorError` and `is_media_shed`/`is_load_shed` use at the merge base `88049d9` (`git grep`), against their replacements at the head.
**CI**: green (format, lint `--all-targets -D warnings`, test: common, webapi, watcher and nexusd suites against the Docker stack).

**Ruled out**
- *Behaviour change in `get_content_type_for_variant`*: the new dispatch matches the old processor dispatch arm by arm. `Main` still returns the file's own type, images still `image/webp` (now derived from `IMAGE_FORMAT`), and the `_` arm still passes the type through.
- *Media handling lost from `From<ModelError>`*: the only base callers of `Blob::get_by_id` were the avatar and static-file handlers, and both now reach `From<MediaProcessorError>`, which keeps the load-shed to 503 branch; the watcher never derived a variant, so its `MediaProcessorError` variant was unreachable.
- *Client-visible error text*: a genuine conversion failure was `Internal server error: MediaProcessorError` before (via `ModelError`'s terse `Display`) and is asserted to the same string now, with the converter stderr kept in the `{:?}` operator log.
- *Nothing runs the moved media paths*: `cargo nextest run -p nexus-webapi` runs the DB-backed suite in CI, including the tests that moved with the processors.
- *Variant table and processor options now live in different crates*: a variant added to the table without a matching `get_options_for_variant` arm would surface as a 404 rather than a 400. The same shape existed before, so it is not this diff's defect and is not posted.

</details>

Merge confidence: 5/5, CI green, files read whole, and the moved media paths run in the webapi integration suite.

tipogi (approved):
LGTM!

PR #1079, opened 2026-09-21
bench(streams): cover tag-filtered post streams for every source

aintnostressin:
The tag stream benches only cover `source=All` with a single tag. Tag-filtered streams on any other source, or with more than one tag, had no bench.

This adds one bench per stream request, in the style of the other stream benches:

- `stream_posts_multi_tag_*`: two tags (`bitcoin`, `fees`) on `All`. The existing `stream_posts_tag_*` benches cover a single tag, which is served from that tag's sorted set; two or more tags are not
- `stream_posts_{following,followers,friends}_tag_*`: a tag on each reach
- `stream_posts_wot_depth{1,2,3}_tag_*`: a tag on the web of trust
- `stream_posts_bookmarks_tag_*`: a tag on bookmarks
- `stream_posts_wot_domain_{me,depth2}_tag_*`: a tag on the domain web of trust, with domain tag `bot`
- `stream_posts_author_tag_*`: a tag on an author's posts

Each runs with both `timeline` and `total_engagement` sorting, for 22 benches in total. A shared `run_tag_stream` helper keeps them short, like `run_wot_depth` in `wot.rs`. The two existing single-tag benches now use it too; their bench names are unchanged, so existing Criterion baselines still compare.

Users and labels come from the mock graph, and each case was checked to return posts, so none of them times an empty stream. The reach sources reuse the observer from the reach and wot benches. Bookmarks and author reuse the users from their own benches. The domain streams use the domain tag `bot` rather than the wot bench's `bitcoin` and `dev`, because no author endorsed with those wrote a `bitcoin`-tagged post, so both streams would be empty. Authors endorsed as `bot` wrote 110. The `Me` domain stream uses the author as the observer, because the reach observer has endorsed no users.


aintnostressin:
@greptile review

talosmachina (commented):
**No findings.** Adds one benchmark per tag-filtered post stream (multi-tag, each reach, the three wot depths, both domain-wot trusts, bookmarks, author), each in both sortings, and routes the two existing single-tag benches through a shared `run_tag_stream` helper without changing their bench ids. Reviewed `b9b6002`, full tier.

<details><summary>What I checked, and the candidates I ruled out</summary>

**Read in full**: `nexus-webapi/benches/streams_benches/tag.rs`, `nexus-webapi/benches/streams.rs`, the sibling bench modules (`reach.rs`, `wot.rs`, `bookmarks.rs`, `author.rs`, `sorting.rs`, `utils.rs`, `mod.rs`), `PostStream::get_posts` / `collect_post_keys` / `can_use_index` (`nexus-common/src/models/post/stream.rs`), `db/graph/queries/get.rs::post_stream`, `types/mod.rs`
**CI**: green (format, lint, test); `lint` runs `cargo clippy --all-targets -D warnings`, so the bench targets compile
**Verified directly**: ran the eleven new stream cases as Cypher against the seeded mock graph in neo4j. Every one returns posts: multi-tag 113, following 3, followers 2, friends 2, wot d1 3, d2 113, d3 113, bookmarks 2, domain Me 110, domain d2 110, author 2. The comment counts in the diff (3 in following, 113 in wot_2, 110 for domain) match.

**Ruled out**
- *Bench ids changed, so Criterion baselines stop comparing*: the two existing ids are unchanged (`stream_posts_tag_timeline`, `stream_posts_tag_total_engagement`); only the body moved into the helper.
- *Duplicate benchmark id in the group*: every `bench_function` label in `nexus-webapi/benches/` is unique.
- *Wrong observer/source pairs*: the observer, bookmarks observer and author ids match `reach.rs`/`wot.rs`, `bookmarks.rs` and `author.rs`; `DomainTrust::Network(WotDepth::default())` is depth 2 (`types/mod.rs:30`), matching the `depth2` id.
- *`@labels` param never bound for the graph path*: `build_query_with_params` binds `labels` whenever tags is `Some`.
- *Bench measures an empty stream on the multi-tag and domain paths*: counts above; the `Me` observer (the author) does carry a `bot` endorsement, and the reach observer carries none.
- *The `run_setup()` per bench is repeated work*: it is a `Once`, so it runs once per process.

</details>

Merge confidence: 5/5, bench-only change, CI green (clippy `--all-targets` compiles the new benches) and every new case's stream verified non-empty against the seeded mock graph.

PR #1080, opened 2026-09-22
feat(search): filter users by tags by the user's reach

aintnostressin:
`GET /v0/search/users/by_tags` accepts `user_id` and `reach`, which must be given together. Results are the users in that reach (never `user_id` itself), with unfiltered scores and the same `score DESC, user_id DESC` tie-break. An unknown user returns `[]`.

Every reach and label count is served by the new
`search_users_by_tags_with_reach` graph query, which dedupes the reach, excludes the user and pages inside Neo4j, so only the requested window reaches Nexus. Friends are found as followed users who follow back, checked with EXISTS, which is much cheaper than matching both directions as one pattern.

# Pre-submission Checklist


aintnostressin:
@greptile review

talosmachina (commented):
**No findings.** `GET /v0/search/users/by_tags` gains optional `user_id` + `reach` params (both or neither) that narrow the results to the user's reach, excluding the user, with unfiltered scores and the same `score DESC, user_id DESC` tie-break. Reviewed `679eaec`, full tier.

<details><summary>What I checked, and 6 candidates I ruled out</summary>

**Read in full**: `nexus-common/src/db/graph/queries/get.rs` (the new `search_users_by_tags_with_reach`, the shared `users_by_tags_cypher`, `stream_reach_to_graph_subquery`, `reach_attrs`), `nexus-common/src/models/user/search.rs`, `nexus-webapi/src/routes/v0/search/users.rs`, `nexus-webapi/tests/user/search_by_tags.rs`, `nexus-common/src/types/mod.rs` (`StreamReach`, `WotDepth`)
**Call sites traced**: `get_by_labels_with_reach` to the route handler; `get_from_graph`, whose signature moved to a built `Query`; the other `stream_reach_to_graph_subquery` users (`get_tag_taggers_by_reach`, `get_hot_tags_by_reach`, `get_influencers_by_reach`)
**CI**: red. `stream::user::influencers::test_global_influencers_preview` failed on `first_influencer_ids != second_influencer_ids`, a clock-phase-dependent preview assertion this diff does not touch, with a separate fix in progress. Every `user::search_by_tags` test passed, including all seven new reach tests.

**Ruled out**
- *Reach dedupe wrong for WoT*: `WITH DISTINCT reach AS u` runs before the tag join, so a user reached through several paths scores once; `user_tag_search_dedupes_reach_before_tag_join` asserts the ordering for all six reaches.
- *Observer or stranger leaking into the reach*: `WHERE reach.id <> $user_id` drops the observer, and `test_user_search_by_tags_reach_excludes_observer` checks OBS and STRANGER across all six reaches and a second observer.
- *Unknown user erroring instead of returning `[]`*: the anchor `MATCH (user:User {id: $user_id})` yields no rows, so the whole query is empty; `test_user_search_by_tags_reach_unknown_user` covers it.
- *`reach` accepted without `user_id`*: the XOR check returns 400 before any query; `test_user_search_by_tags_reach_requires_both_params` covers both halves plus a malformed id.
- *Pagination drift between the graph window and the caller*: `SKIP`/`LIMIT` are applied inside Neo4j after `ORDER BY score DESC, u.id DESC`, and `test_user_search_by_tags_reach_pagination` compares every page against the full list for three reaches.
- *Shared-cypher refactor changing the unfiltered path*: `users_by_tags_cypher("")` reproduces the previous string with only whitespace differing; `test_user_search_by_tags_redis_graph_parity` still passes.

</details>

Merge confidence: 3/5, CI red on a pre-existing flake in an unrelated stream test.

tipogi (approved):

